PR #2, opened 2019-06-25
Replace the numpy structured type by a python extensiontype

caspervdw:
I replaced the numpy structured dtype by a plain python object, that has a static attribute containing the pointer to the actual GEOSGeometry.

I did not find any noticable performance decrease compared to the previous implementation, and similar performance as the geopandas-cython one.

If you have time for feedback or benchmarks, please let me know :)
@jorisvandenbossche @snowman2 @mrocklin


snowman2:
Looks like it still has the speed:

```python
In [1]: from pygeos import Point, Polygon                                       

In [2]: import numpy as np                                                      

In [3]: a = np.array([Point(i, i) for i in range(10000)], dtype=object)         

In [4]: poly = Polygon([(10,10), (10,100), (100,100), (100, 10)])               

In [5]: from pygeos import contains                                             

In [6]: %timeit contains(poly, a)                                               
251 µs ± 4.41 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

What are the benefits of this approach?

caspervdw:
Thanks for the benchmark

> What are the benefits of this approach?

This couples the geometries stored in the array to the python garbage collector. If an object is dereferenced through array slicing/indexing/masking, the underlying GEOSGeometry object will be deallocated.

In the previous implementation, this would only occur if you called `finalize_garr`. If you chain multiple GEOS operations together, this `finalize_garr` operation would add a lot of unnecessary overhead.

snowman2:
Interesting, thanks for the explanation. This is a very nice solution - it definitely simplifies things from a users perspective.

jorisvandenbossche:
Finally trying to get some thoughts written down:

First, thanks a lot for exploring this. And let me also say that I find the ufunc interface really nice. The additional functionalities you get from numpy (automatic broadcasting, ..) is nice.

And also, as @mrocklin said, having the GC handled is a *big* advantage (by using the custom PyObjects), compared to the cython GeometryArray approach in geopandas.

One thing is that the vectorized operations we need in geopandas are (I think) not all generalizable as ufuncs. So in addition to the automatic ufuncs, we might still need additional ("manual") functions in C/C++/Cython. For example, `unary_union` (where you first need to create a GeometryCollection, so the one in pygeos is not doing what the shapely's unary_union would do on an array of geometries). Other examles are the serialization/deserialization methods, the spatial join algo, .. 
Of course the approach for ufuncs does not exclude doing other things for those functionalities (in the same library)! 

Trying to summarize some other thoughts / concerns:

- **End-user usability**: it would be nice that the element you get out of an array, is the familiar shapely geometry object. 
  Ora (https://github.com/alexhsamuel/ora) which also does custom dtypes might be an interesting example (didn't look in detail yet myself, but it seems they are doing something like that, specifying what setitem and getitem for theird dtype should do)

- **Interfacing with cython/numba** (to easily extend with custom algorithms). I didn't try yet, but I suppose that if cython knows about the structure of the GeometryObject, this should work fine for cython (eg to write similar things as in geopandas-cython but using the object numpy array as holder of the pointers).

- **Parallelization**: I don't know if there is a way to parallelize the ufuncs (based on this https://github.com/numpy/numpy/issues/8208 it seems not). 
  This isn't done yet, but I was planning to try to parallelize the cython code in the geopandas branch (using cython's `prange`), as all those elementwise GEOS functions are in theory an "embarrassingly parrallel" problem. And it would be nice to give this speedup natively in geopandas/pygeos, instead of needing to use dask for this (also in that case, dask is still interesting for distributed workloads).

- **Spatial indexing**: I was also planning to explore the possible speed-up by using the STRTree provided by GEOS in operations like vectorized `contains`. For the current version of geopandas, that can provide a nice speedup based on some very short experiments, but when the actual operation is faster, the benefit might get smaller. So it might be less relevant for the vectorized operations, but it might still be interesting to keep the possibility open to look at.

- **Support for missing values**. Currently in the geopandas-cython version, we use a null pointers to indicate NAs. I think it is something we want for geopandas (you can easily end up with missing values due to alignment), but it's maybe not the most essential aspect to be directly supported in the c loops, we could also keep a separate mask for geopandas' use case.

**Creation and memory overhead**: one benefit of storing the raw pointers is that this has minimal additional memory footprint and overhead when creating a new array of geometries from an operation. But it seems with the very thin GeometryObject, this is also very OK.

Issue #26, opened 2019-08-18
get_geometry is not safe (segfaults on out of bound index)


jorisvandenbossche:
A small issue I noticed, doing a `pygeos.get_geometry(arr, i)` crashes when `i` is out of bounds.
Not sure on what level this can be checked, but ideally (IMO), a python function should never crash but raise an error.

caspervdw:
Agreed, this should be looked into.

caspervdw:
@jorisvandenbossche  I was looking at this issue. Currently, three functions suffer from it: `get_geometry`, `get_interior_ring`, and `get_point`. These are for collections, polygons and linestrings respectively.

Would it make sense to make a single function out of this (`get_element`) that identifies the geometry type, does the bounds checking using the correct function, and gets the element using the correct function?

PR #30, opened 2019-08-25
Document and test functions in geometry.py

caspervdw:
This solves #26

### DIFF
--- a/pygeos/test/test_geometry.py
+++ b/pygeos/test/test_geometry.py
@@ -12,12 +12,12 @@ def test_get_type_id():
     assert pygeos.get_type_id(all_types).tolist()[:-1] == list(range(8))
 
 
-def test_get_num_points():
-    assert pygeos.get_num_points(line_string) == 3
+def test_get_num_elements():
+    assert pygeos.get_num_elements(line_string) == 3
 
 
-def test_get_point():
-    actual = pygeos.get_point(line_string, 1)
+def test_get_element():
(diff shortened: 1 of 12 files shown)


pygeos/test/test_geometry.py
```diff
@@ -12,12 +12,12 @@ def test_get_type_id():
...
-    assert pygeos.get_num_points(line_string) == 3
+def test_get_num_elements():
+    assert pygeos.get_num_elements(line_string) == 3
 
 
-def test_get_point():
-    actual = pygeos.get_point(line_string, 1)
+def test_get_element():
+    actual = pygeos.get_element(line_string, 1)
     assert pygeos.equals(actual, pygeos.points(1, 0))
```
jorisvandenbossche: I would add a test that an out of bounds number actually raises now

caspervdw:
Thanks for the feedback. I looked a bit in how PostGIS does it, and decided to split the functions into three after all:
 - get_points / get_num_points
 - get_interior_ring / get_num_interior_rings
 - get_geometry / get_num_geometries

If the type is incorrect or the index out of bounds, it will return an Empty geometry (should become: Missing, but that one isn't there yet)

jorisvandenbossche:
> If the type is incorrect or the index out of bounds, it will return an Empty geometry (should become: Missing, but that one isn't there yet)

From a user point of view, I would rather expect an error?

PR #33, opened 2019-08-27
Split single ufuncs.c file in multiple files

jorisvandenbossche:
This is an attempt to split the single (big) ufuncs C file in multiple files.

I find this a bit easier to understand the code if it is not all in a huge file, but it was also just a good exercise to get familiar with the code, so @caspervdw you certainly don't need to like this ;)

### DIFF
--- a/src/pygeom.c
+++ b/src/pygeom.c
@@ -1,0 +1,316 @@
+#define PY_SSIZE_T_CLEAN
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+
+#include <Python.h>
+#include <structmember.h>
+
+#include "pygeom.h"
+#include "geos.h"
+
+
+/* This initializes a pointer to a NULL geometry */
+GeometryObject *Geom_Empty = NULL;
+
+/* Initializes a new geometry object, without Empty check */
+PyObject *GeometryObject_FROMGEOS(PyTypeObject *type, GEOSGeometry *ptr)
+{
+    if (ptr == NULL) {
+        Py_INCREF(Geom_Empty);
+        return (PyObject *) Geom_Empty;
+    }
+    GeometryObject *self = (GeometryObject *) type->tp_alloc(type, 0);
+    if (self == NULL) {
+        return NULL;
+    } else {
+        self->ptr = ptr;
+        return (PyObject *) self;
+    }
+}
+
+/* Initializes a new geometry object, with Empty check */
+PyObject *GeometryObject_FromGEOS(PyTypeObject *type, GEOSGeometry *ptr)
+{
+    void *context_handle = geos_context[0];
+    char empty;
+    if (ptr == NULL) {
+        return GeometryObject_FROMGEOS(type, ptr);
+    } else if (ptr == Geom_Empty->ptr) {
+        return GeometryObject_FROMGEOS(type, NULL);
+    } else {
+        empty = GEOSisEmpty_r(context_handle, ptr);
+        if (empty == 2) {
+            return NULL;
+        } else if (empty == 1) {
+            GEOSGeom_destroy_r(context_handle, ptr);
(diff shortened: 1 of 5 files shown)


caspervdw:
This is great, thanks for doing this work. I am surprised AppVeyor isn't picking up this PR, probably I switched PR building off for some reason.

src/pygeom.c
```diff
@@ -0,0 +1,316 @@
...
+    .tp_members = GeometryObject_members,
+    .tp_methods = GeometryObject_methods,
+    .tp_repr = (reprfunc) GeometryObject_repr,
+};
+
+int
+init_geom_type(PyObject *m)
+{
+    if (PyType_Ready(&GeometryType) < 0)
+        return NULL;
```
jorisvandenbossche: There is one remaining (newly introduced) build warning I didn't fully solve. It is complaining here that it is returning NULL (a pointer), while I declared it to return `int`. And also that the rest does not return anything.

But maybe I can return here 0 / -1, and do the `return NULL` in the actual `PyInit_ufuncs` function

caspervdw: I think you should indeed return 0 / -1 and depending on that, `return NULL` in the `PyInit_ufuncs` (in `pygeos/ufuncs.c`) if necessary

src/pygeom.c
```diff
@@ -0,0 +1,316 @@
...
+        return GeometryObject_FromWKB(type, value);
+    }
+    else if (PyUnicode_Check(value)) {
+        return GeometryObject_FromWKT(type, value);
+    }
+    else {
+        PyErr_Format(PyExc_TypeError, "Expected string or bytes, found %s", value->ob_type->tp_name);
+        return NULL;
+    }
+    ptr = GEOSGeom_clone_r(context_handle, arg);
```
caspervdw: This code is unreachable! Should be deleted. My bad, I think

src/pygeom.c
```diff
@@ -0,0 +1,316 @@
...
+    .tp_dealloc = (destructor) GeometryObject_dealloc,
+    .tp_members = GeometryObject_members,
+    .tp_methods = GeometryObject_methods,
+    .tp_repr = (reprfunc) GeometryObject_repr,
+};
+
+int
+init_geom_type(PyObject *m)
+{
+    if (PyType_Ready(&GeometryType) < 0)
```
caspervdw: I think the normal do this kind of check is `if (!PyType_Ready(&GeometryType))`

caspervdw:
I now realize that all feedback is related to stuff I wrote myself. Feel free to merge as is (except for the compiler warning, would be nice to solve that)

jorisvandenbossche:
No problem, I can address those at once when fixing the other issue.

One additional thought I had is if we want to keep the "ufuncs" name for the compiled module in Python? Since it already contains more as just the ufuncs (it also already contains the base Geometry type), and might contain more in the future (dtype, STRTree, other algos like a join). 
Eg, we could have a single compiled module `pygeos.lib`.

jorisvandenbossche:
Or, we could actually also expose them as multiple python modules, although I don't know how that works with interlinking between them on the C level if they are not compiled into a single linked module.

Anyway, in the end, users don't directly use this but rather the wrapped python functions, so the name is also not that important.

src/pygeom.c
jorisvandenbossche: Here this is actually correct, as `PyType_Ready` does not return NULL on failure but -1

caspervdw:
> One additional thought I had is if we want to keep the "ufuncs" name for the compiled module in Python? Since it already contains more as just the ufuncs (it also already contains the base Geometry type), and might contain more in the future (dtype, STRTree, other algos like a join).
Eg, we could have a single compiled module pygeos.lib.

I like the idea of making this more modular on the C level, and expose it with `pygeos.lib` as a single module. Something for a next round of refactor.

PR #36, opened 2019-09-02
[Done] Separating Missing (NaG) and Empty

### LINKED ISSUE #28: Handling of missing geometries
Related to the discussion in https://github.com/geopandas/geopandas/pull/1062

Currently, missing geometries (np.nan, None), are handled similarly to empty geometries:

```
In [1]: import pygeos

In [4]: a = np.array([None, np.nan, pygeos.Geometry("POLYGON EMPTY"), pygeos.box(0, 0, 1, 1)])

In [5]: a
Out[5]: 
array([None, nan, <pygeos.Empty>,
       <pygeos.Geometry POLYGON ((1 0, 1 1, 0 1, 0 0, 1 0))>],
      dtype=object)

In [6]: pygeos.is_empty(a)
Out[6]: array([ True,  True,  True, False])
```

But I think, based on your comment in https://github.com/geopandas/geopandas/pull/1062, that it is the idea to separate both concepts?

But also then, there are still some aspects that need to be decided. Quoting @caspervdw from the linked geopandas issue:

> I am curious about how to deal with missing values in the various GEOS functions, especially for:
>
> - predicates (unary and binary)
> - integer-returning functions (like “get_num_points”
> - aggregation functions (should “create_collection” skip or propagate NaN?)
>
> Is it an idea to implement something like NaG (not-a-geometry)?

### LINKED ISSUE #29: Handling of empty geometries
Related to https://github.com/caspervdw/pygeos/issues/28 about missing values and the discussion in https://github.com/geopandas/geopandas/pull/1062. 
But even when we split missing values as a separate concept, there are still some potential pitfalls and corner cases regarding empty geometries itself (as I experienced on that geopandas PR).

Currently, there is a `pygeos.Empty` "singleton" defined as an empty GeometryCollection:

https://github.com/caspervdw/pygeos/blob/de5ed5c5f0a55bf471531e328d1a1b0476bcafd2/src/ufuncs.c#L1213-L1216

However, GEOS has also the notion of eg empty polygons. 
See https://github.com/Toblerity/Shapely/issues/742 for some discussion in Shapely about this.

Having a more consistent empty handling as Shapely would be nice, but we might need to consider the potential consequences. Eg I remember that some file formats are strict regarding only allowing a single geometry type in a single column (eg only (multi)polygons). It might be that then an empty geometrycollection instead of empty polygon can give problems (purely hypothetical, didn't check it).

caspervdw:
Closes #28, closes #29

This is some work separating the concepts of missing geometries (not-a-geometry or NaG) values and empty geometries. The idea is to keep a single NaG object (like None) that has its GEOSGeometry pointer set to NULL.

All geometry-returning functions propagate the NaG; predicates return False for any import NaG; integer returning functions return -1 for NaG.

I added two additional functions to identify NaG values: `is_geometry` and `is_null`.

Further there should be functions that create arrays of empty geometries.


jorisvandenbossche:
Cool! I will try to take a look at it one of the coming days. Currently at EuroScipy conference, so a bit busy, but, I will sit together here with one of the numpy developers who is interested in seeing the dtype stuff :-)

src/pygeom.c
```diff
@@ -9,14 +9,20 @@
...
-/* Initializes a new geometry object, without Empty check */
-PyObject *GeometryObject_FROMGEOS(PyTypeObject *type, GEOSGeometry *ptr)
+/* Initializes a new geometry object */
+PyObject *GeometryObject_FromGEOS(PyTypeObject *type, GEOSGeometry *ptr)
 {
     if (ptr == NULL) {
-        Py_INCREF(Geom_Empty);
-        return (PyObject *) Geom_Empty;
+        if (NaG == NULL) {
+            /* First time: initialize the NaG object */
```
jorisvandenbossche: Shouldn't we do this first time initialization in the ufuncs module initialization function?

src/ufuncs.c
```diff
@@ -502,24 +485,36 @@ static void Y_i_func(char **args, npy_intp *dimensions,
...
-        if ((ret < 0) | (ret > NPY_MAX_INT)) {
-            RAISE_ILLEGAL_GEOS;
-            return;
+        /* get the geometry: return on error */
+        if (!get_geom(*(GeometryObject **)ip1, &in1)) { return; }
+        if (in1 == NULL) {
+            /* NaG results in -1 */
+            *(npy_int *)op1 = -1;
+        } else {
+            /* trust that GEOS calls the error handler if necessary */
```
jorisvandenbossche: This is mentioned many times, but wondering, do we actually test this? (in which occasion would GEOS raise an error? sometimes one invalid geometries?)

src/ufuncs.c
```diff
@@ -169,26 +157,34 @@ static void YY_b_func(char **args, npy_intp *dimensions,
...
+        if (!get_geom(*(GeometryObject **)ip1, &in1)) { return; }
+        if (!get_geom(*(GeometryObject **)ip2, &in2)) { return; }
+        if ((in1 == NULL) | (in2 == NULL)) {
+            /* in case of a missing value: return 0 (False) */
+            ret = 0;
+        } else {
+            /* call the GEOS function */
+            ret = func(context_handle, in1, in2);
+            /* return for illegal values (trust HandleGEOSError for SetErr) */
+            if ((ret < 0) | (ret > 1)) { return; }
```
jorisvandenbossche: `ret != 1` instead ?

pygeos/predicates.py
```diff
@@ -73,18 +75,77 @@ def is_empty(geometry, **kwargs):
...
+    Examples
+    --------
+    >>> is_null(Geometry("POINT (0 0)"))
+    False
+    >>> is_null(Geometry("GEOMETRYCOLLECTION EMPTY"))
+    False
+    >>> is_null(None)
+    True
+    >>> is_null(float("nan"))
+    True
```
jorisvandenbossche: So this is somewhat related to the comment I made below about limiting the allowed input (checking each element from being float gives an overhead, and for me it is not necessary a needed feature to be able to pass all of three potential missing-like objects)

pygeos/set_operations.py
```diff
@@ -78,8 +76,8 @@ def intersection_all(geometries, axis=0, **kwargs):
     >>> line_2 = Geometry("LINESTRING(1 1, 3 3)")
     >>> intersection_all([line_1, line_2])
     <pygeos.Geometry LINESTRING (1 1, 2 2)>
-    >>> intersection_all([[line_1, line_2, Empty]], axis=1).tolist()
-    [<pygeos.Empty>]
+    >>> intersection_all([[line_1, line_2, NaG]], axis=1).tolist()
+    [<pygeos.NaG>]
```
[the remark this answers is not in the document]
caspervdw: Yes, I'd like do continue that discussion in #37

pygeos/measurement.py
```diff
@@ -39,7 +39,7 @@ def distance(a, b):
     1.0
     >>> distance(Geometry("POLYGON ((3 0, 5 0, 5 5, 3 5, 3 0))"), point)
     3.0
-    >>> distance(Empty, point)
```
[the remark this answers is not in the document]
caspervdw: GEOS actuall returns `0.0` if you do this. Which is strange. We could change the GEOS API here.

pygeos/predicates.py
```diff
@@ -73,18 +75,77 @@ def is_empty(geometry, **kwargs):
...
+
+    Examples
+    --------
+    >>> is_geometry(Geometry("POINT (0 0)"))
+    True
+    >>> is_geometry(Geometry("GEOMETRYCOLLECTION EMPTY"))
+    True
+    >>> is_geometry(None)
+    False
+    >>> is_geometry(float("nan"))
```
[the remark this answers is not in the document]
caspervdw: Agreed, all these None types are really not necessary. I dropped the NaG and now only None is supported.

src/ufuncs.c
caspervdw: There are some cases tested:
https://github.com/caspervdw/pygeos/search?q=pygeos.GEOSException&unscoped_q=pygeos.GEOSException

GEOS should do a callback to the error message handler (`HandleGEOSError`), which I defined such that a GEOSException is set. I have never encountered the situation where GEOS didn't do this.

caspervdw:
Thanks for the remarks, I made some changes:
 - Dropped the NaG and NaN support in favour of only None
 - is_null is now replaced by is_missing (which is just np.equals(arr, None) under the hood)

caspervdw:
> Additionally to those, it would alos be useful to have a fast ufunc exposed to just check that the input only contains "valid objects" (in pygeos array's context, so a geometry or a NaG) as well. So basically a is_geometry(a) | is_null(a), but as single ufunc this can be more efficient.

Maybe good to wait to see how this numpy-dtype extension thing plays out?

jorisvandenbossche:
> Maybe good to wait to see how this numpy-dtype extension thing plays out?

On the short term, it unfortunately won't play out. See the discussion in #34. It will need changes in numpy to be feasible (but changes they would welcome / want to plan).

pygeos/constructive.py
```diff
@@ -51,8 +51,8 @@ def boundary(geometry, **kwargs):
     <pygeos.Geometry MULTIPOINT EMPTY>
     >>> boundary(Geometry("POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))"))
     <pygeos.Geometry LINESTRING (0 0, 1 0, 1 1, 0 1, 0 0)>
-    >>> boundary(Geometry("MULTIPOINT (0 0, 1 2)"))
-    <pygeos.NaG>
+    >>> boundary(Geometry("MULTIPOINT (0 0, 1 2)")) is None
```
jorisvandenbossche: Alternative could be to do `print(..)` which then actually shows `None` (but both are of course workarounds for None having no repr)

pygeos/measurement.py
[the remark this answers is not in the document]
jorisvandenbossche: It's indeed strange.In GeoPandas we return NaN (and I kept that in the recent refactoring, as we had been doing that for a long time). See the bullet point in https://github.com/geopandas/geopandas/pull/1062#issuecomment-520127793. But, it's also annoying to need to deviate from GEOS ..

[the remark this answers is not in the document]
caspervdw: I did pushed the change, I think it makes no sense at all to return 0.0 when computing the distance to an empty set.

caspervdw:
> > Maybe good to wait to see how this numpy-dtype extension thing plays out?
> 
> On the short term, it unfortunately won't play out. See the discussion in #34. It will need changes in numpy to be feasible (but changes they would welcome / want to plan).

Oh I wasn 't following the discussion. So I implemented 3 ufuncs `is_missing`, `is_geometry`, `is_valid_input`

PR #38, opened 2019-09-04
Change doc for Conda Linux install

ThomasG77:


README.rst
```diff
@@ -109,12 +109,18 @@ these using Anaconda::
...
 
     $ export GEOS_INCLUDE_PATH=$CONDA_PREFIX/Library/include
     $ export GEOS_LIBRARY_PATH=$CONDA_PREFIX/Library/lib
     $ pip install pygeos
 
+On Linux::
+
+    $ export GEOS_INCLUDE_PATH=$CONDA_PREFIX/include
+    $ export GEOS_LIBRARY_PATH=$CONDA_PREFIX/lib
+    $ pip install pygeos
```
jorisvandenbossche: Not purely related to your edit since it is already in there right now (and your fix is certainly correct), but: is exporting those env variables actually needed? 
When installing with conda, it should provide the `geos-config` command in the environment, and then the `get_geos_paths` in `setup.py` should automatically find geos in the conda env (at least I am not setting any evn variables, and it is working for me).

ThomasG77: I'm not sure it's required. I have adjusted the paths as I've seen they were wrong. Due to your feedback, trying in a virtual environment, like you, I was able to make successfully the install e.g:

```bash
    conda create -n pygeodemo python=3.6
    conda activate pygeodemo
    pip install pygeos
```

caspervdw: I remember when I tried installing with conda, `geos-config` was not available through conda. But now I try again I see that it is actually there.

So I am in favor of removing the setting of the two variables. I have no idea how this would work on a Mac. Anybody got experience with that?

caspervdw:
And thanks for your contribution.

As a general note, we should really move to conda-forge on the next release.

README.rst
jorisvandenbossche: A difference might be that you need to activate your environment first or restart the console before it is available.

No experience with Mac.

I would remove setting in the default instructions, but maybe we should still mention that those can be set in case GEOS is not found automatically.

jorisvandenbossche:
> As a general note, we should really move to conda-forge on the next release.

I started a conda-forge recipe: https://github.com/conda-forge/staged-recipes/pull/9492

caspervdw:
Great! For this PR, we should drop the exports on Linux and OSX (Travis tests without)

Maybe state in an “advanced” section that you can manually specify paths.

Shall we tag 0.4 after this?

ThomasG77:
I think I can close the PR without a need for a merge as it helped think about a better improvement than the PR itself. Ok for you?

jorisvandenbossche:
OK, I will do a PR updating the description now there are conda packages. But @ThomasG77 thanks anyway!

PR #45, opened 2019-09-27
ENH: add wkt / wkb ufuncs

jorisvandenbossche:
Start for adding from_wkt / to_wkt / from_wkb / to_wkt ufuncs.

They are right now already defined on the Geometry object, but those are useful as ufuncs as well. 
We can probably remove the implementation in pygeom.c to avoid duplication. But the question is whether we want to keep those as methods in the public API?

I should maybe also move this into a separate .c file, as the ufuncs.c is starting to become quite long? (although it are ufuncs)

Still need to make the nice Python functions, only the barebone ufuncs for now for the readers. 
Also still need to do proper error handling (ensure that I destroy the reader when there is an error)

### DIFF
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -171,4 +171,8 @@ def from_wkb(geometry, **kwargs):
     <pygeos.Geometry POINT (1 1)>
 
     """
+    # ensure the input has object dtype, to avoid numpy inferring it as a
+    # fixed-length string dtype (which removes trailing null bytes upon access
+    # of array elements)
+    geometry = np.asarray(geometry, dtype=object)
     return lib.from_wkb(geometry, **kwargs)
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -1,9 +1,10 @@
 import numpy as np
+
 from . import Geometry  # noqa
 from . import lib
 
 
-__all__ = ["from_wkb", "from_wkt", "to_wkb", "to_wkb"]
+__all__ = ["from_wkb", "from_wkt", "to_wkb", "to_wkt"]
 
 
 def to_wkt(
@@ -15,10 +16,10 @@ def to_wkt(
     **kwargs
 ):
     """
-    Converts to the Well-Known Text (WKT) representation of a  Geometry.
+    Converts to the Well-Known Text (WKT) representation of a Geometry.
 
     The Well-known Text format is defined in the `OGC Simple Features
-    Specification for SQL <http://www.opengis.org/techno/specs.htm>`__.
+    Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
 
     Parameters
     ----------
@@ -70,3 +71,65 @@ def to_wkt(
         np.bool(old_3d),
         **kwargs,
     )
+
+
+def to_wkb(geometry, **kwargs):
+    """
+    Converts to the Well-Known Binary (WKB) representation of a Geometry.
+
+    The Well-Known Binary format is defined in the `OGC Simple Features
+    Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+
+    Examples
+    --------
+    >>> to_wkb(Geometry("POINT (1 1)"))
+    b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?'
+
+    """
+    return lib.to_wkb(geometry, **kwargs)
+
+
+def from_wkt(geometry, **kwargs):
+    """
+    Creates geometries from the Well-Known Text (WKT) representation.
+
+    The Well-known Text format is defined in the `OGC Simple Features
+    Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
+
+    Parameters
+    ----------
+    geometry : str or array_like
+        The WKT string(s) to convert.
+
+    Examples
+    --------
+    >>> from_wkt('POINT (0 0)')
+    <pygeos.Geometry POINT (0 0)>
+
+    """
+    return lib.from_wkt(geometry, **kwargs)
+
+
+def from_wkb(geometry, **kwargs):
+    """
+    Creates geometries from the Well-Known Binary (WKB) representation.
+
+    The Well-Known Binary format is defined in the `OGC Simple Features
+    Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
+
+    Parameters
+    ----------
+    geometry : str or array_like
+        The WKB byte object(s) to convert.
+
+    Examples
+    --------
+    >>> from_wkb(b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?')
+    <pygeos.Geometry POINT (1 1)>
+
+    """
+    return lib.from_wkb(geometry, **kwargs)
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -57,9 +57,9 @@ def to_wkt(
     Notes
     -----
     The defaults differ from the default of the GEOS library. To mimic this,
-    use:
+    use::
 
-    >>> to_wkt(geometry, rounding_precision=-1, trim=False, output_dimension=2)
+        to_wkt(geometry, rounding_precision=-1, trim=False, output_dimension=2)
 
     """
     if not np.isscalar(rounding_precision):
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -28,7 +28,7 @@ def to_wkt(
         The rounding precision when writing the WKT string. Set to a value of
         -1 to indicate the full precision.
     trim : bool, default True
-        Whether to trim unnecessary decimals (trailing zero's).
+        Whether to trim unnecessary decimals (trailing zeros).
     output_dimension : int, default 3
         The output dimension for the WKT string. Supported values are 2 and 3.
         Specifying 3 means that up to 3 dimensions will be written but 2D
@@ -46,7 +46,6 @@ def to_wkt(
     'POINT (0.000 0.000)'
     >>> to_wkt(Geometry("POINT (0 0)"), rounding_precision=-1, trim=False)
     'POINT (0.0000000000000000 0.0000000000000000)'
-
     >>> to_wkt(Geometry("POINT (1 2 3)"), trim=True)
     'POINT Z (1 2 3)'
     >>> to_wkt(Geometry("POINT (1 2 3)"), trim=True, output_dimension=2)
@@ -111,7 +110,6 @@ def to_wkb(geometry, hex=False, output_dimension=3, byte_order=-1, include_srid=
     b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?'
     >>> to_wkb(Geometry("POINT (1 1)"), hex=True)
     '0101000000000000000000F03F000000000000F03F'
-
     """
     if not np.isscalar(hex):
         raise TypeError("hex only accepts scalar values")
@@ -148,7 +146,6 @@ def from_wkt(geometry, **kwargs):
     --------
     >>> from_wkt('POINT (0 0)')
     <pygeos.Geometry POINT (0 0)>
-
     """
     return lib.from_wkt(geometry, **kwargs)
 
@@ -169,7 +166,6 @@ def from_wkb(geometry, **kwargs):
     --------
     >>> from_wkb(b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?')
     <pygeos.Geometry POINT (1 1)>
-
     """
     # ensure the input has object dtype, to avoid numpy inferring it as a
     # fixed-length string dtype (which removes trailing null bytes upon access
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -73,7 +73,7 @@ def to_wkt(
     )
 
 
-def to_wkb(geometry, **kwargs):
+def to_wkb(geometry, hex=False, output_dimension=3, byte_order=-1, include_srid=False, **kwargs):
     """
     Converts to the Well-Known Binary (WKB) representation of a Geometry.
 
@@ -83,14 +83,45 @@ def to_wkb(geometry, **kwargs):
     Parameters
     ----------
     geometry : Geometry or array_like
+    hex : bool, default False
+        If true, export the WKB as a hexidecimal string. The default is to
+        return a binary bytes object.
+    output_dimension : int, default 3
+        The output dimension for the WKB. Supported values are 2 and 3.
+        Specifying 3 means that up to 3 dimensions will be written but 2D
+        geometries will still be represented as 2D in the WKB represenation.
+    byte_order : int
+        Defaults to native machine byte order (-1). Use 0 to force big endian
+        and 1 for little endian.
+    include_srid : bool, default False
+        Whether the SRID should be included in WKB (this is an extension
+        to the OGC WKB specification).
 
     Examples
     --------
     >>> to_wkb(Geometry("POINT (1 1)"))
     b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?'
+    >>> to_wkb(Geometry("POINT (1 1)"), hex=True)
+    '0101000000000000000000F03F000000000000F03F'
 
     """
-    return lib.to_wkb(geometry, **kwargs)
+    if not np.isscalar(hex):
+        raise TypeError("hex only accepts scalar values")
+    if not np.isscalar(output_dimension):
+        raise TypeError("output_dimension only accepts scalar values")
+    if not np.isscalar(byte_order):
+        raise TypeError("byte_order only accepts scalar values")
+    if not np.isscalar(include_srid):
+        raise TypeError("include_srid only accepts scalar values")
+
+    return lib.to_wkb(
+        geometry,
+        np.bool(hex),
+        np.intc(output_dimension),
+        np.intc(byte_order),
+        np.bool(include_srid),
+        **kwargs,
+    )
 
 
 def from_wkt(geometry, **kwargs):
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -1,6 +1,6 @@
 import numpy as np
 from . import Geometry  # noqa
-from . import ufuncs
+from . import lib
 
 
 __all__ = ["from_wkb", "from_wkt", "to_wkb", "to_wkb"]
@@ -62,7 +62,7 @@ def to_wkt(
     if not np.isscalar(old_3d):
         raise TypeError("old_3d only accepts scalar values")
 
-    return ufuncs.to_wkt(
+    return lib.to_wkt(
         geometry,
         np.intc(rounding_precision),
         np.bool(trim),
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -1,0 +1,72 @@
+import numpy as np
+from . import Geometry  # noqa
+from . import ufuncs
+
+
+__all__ = ["from_wkb", "from_wkt", "to_wkb", "to_wkb"]
+
+
+def to_wkt(
+    geometry,
+    rounding_precision=-1,
+    trim=False,
+    output_dimension=2,
+    old_3d=False,
+    **kwargs,
+):
+    """
+    Converts to the Well-Known Text (WKT) representation of a  Geometry.
+
+    The Well-known Text format is defined in the `OGC Simple Features
+    Specification for SQL <http://www.opengis.org/techno/specs.htm>`__.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    rounding_precision : int
+        The rounding precision when writing the WKT string. A value of -1
+        indicates the full precision.
+    trim : bool
+        Whether to trim unnecessary decimals (trailing zero's).
+    output_dimension : int
+        The output dimension for the WKT string. Supported values are 2 and 3.
+        Specifying 3 means that up to 3 dimensions will be written but 2D
+        geometries will still be represented as 2D in the WKT string.
+    old_3d : bool
+        Enable old style 3D/4D WKT generation. By default, new style 3D/4D WKT
+        (ie. "POINT Z (10 20 30)") is returned, but with ``old_3d=True``
+        the WKT will be formatted in the style "POINT (10 20 30)".
+
+    Examples
+    --------
+    >>> to_wkt(Geometry("POINT (0 0)"))
+    'POINT (0.0000000000000000 0.0000000000000000)'
+    >>> to_wkt(Geometry("POINT (0 0)"), trim=True)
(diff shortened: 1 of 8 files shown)


jorisvandenbossche:
Some other questions:

- Do we want to keep the hex / non-hex in a single ufunc? I suppose checking for each element individually which of the two it is will give some overhead. Also, you will typically either have all bytes or all hex, not a mixture?
- The hex input is also bytes, while you might typically have this as a string? (not sure myself) For example the `shapely.wkb.dumps(   hex=True)` gives a string.

src/ufuncs.c
```diff
@@ -928,6 +929,211 @@ static void create_collection_func(char **args, npy_intp *dimensions,
...
+    /* Create the WKT writer */
+    writer = GEOSWKTWriter_create_r(context_handle);
+    if (writer == NULL) {
+        return;
+    }
+
+    GEOSWKTWriter_setRoundingPrecision_r(context_handle, writer, precision);
+    GEOSWKTWriter_setTrim_r(context_handle, writer, trim);
+    GEOSWKTWriter_setOutputDimension_r(context_handle, writer, dimension);
+    GEOSWKTWriter_setOld3D_r(context_handle, writer, use_old_3d);
```
caspervdw: I think these should be available as arguments to the functions. The only way I know to do that, is to actually add them to the ufunc inputs. So you'll get 5 input arguments. 

I do think they should e scalars, we don't want to call these 4 methods in the inner loop. I encountered the same problem in the buffer function, which I solved by asserting that it is a scalar.

src/ufuncs.c
```diff
@@ -928,6 +929,211 @@ static void create_collection_func(char **args, npy_intp *dimensions,
...
+        PyObject **out = (PyObject **)op1;
+
+        if (in1 == NULL) {  
+            Py_XDECREF(*out);
+            Py_INCREF(Py_None);
+            *out = Py_None;
+        } else {
+            wkt = GEOSWKTWriter_write_r(context_handle, writer, in1);
+            Py_XDECREF(*out);
+            *out = PyUnicode_FromFormat(format, wkt);
```
caspervdw: In this case `PyUnicode_FromString` is better I think

src/ufuncs.c
```diff
@@ -928,6 +929,211 @@ static void create_collection_func(char **args, npy_intp *dimensions,
...
+    reader = GEOSWKBReader_create_r(context_handle);
+    if (reader == NULL) {
+        return;
+    }
+
+    UNARY_LOOP {
+        in1 = *(PyObject **)ip1;
+
+        /* Cast the PyObject (only bytes) to char* */
+        if (!PyBytes_Check(in1)) {
```
caspervdw: I guess we also need to accept `Py_None`

src/ufuncs.c
```diff
@@ -928,6 +929,211 @@ static void create_collection_func(char **args, npy_intp *dimensions,
...
+        if (wkb == NULL) {
+            return;
+        }
+
+        /* Check if this is a HEX WKB */
+        if (size != 0) {
+            is_hex = ((wkb[0] == 48) | (wkb[0] == 49));
+        } else {
+            is_hex = 0;
+        }
```
caspervdw: I think the performance hit here is negligible, so keep it imho

src/ufuncs.c
```diff
@@ -928,6 +929,211 @@ static void create_collection_func(char **args, npy_intp *dimensions,
...
+
+        /* Cast the PyObject (bytes or str) to char* */
+        if (PyBytes_Check(in1)) {
+            wkt = PyBytes_AsString(in1);
+            if (wkt == NULL) { return; }
+        }
+        else if (PyUnicode_Check(in1)) {
+            wkt = PyUnicode_AsUTF8(in1);
+            if (wkt == NULL) { return; }
+        } else {
```
caspervdw: Same comment as above, we should also accept `None`

src/ufuncs.c
```diff
@@ -928,6 +929,211 @@ static void create_collection_func(char **args, npy_intp *dimensions,
...
+                        npy_intp *steps, void *data)
+{
+    void *context_handle = geos_context[0];
+    GEOSGeometry *in1;
+
+    GEOSWKBWriter *writer;
+    unsigned char *wkb;
+    size_t size;
+    
+    int dimension = 3;
```
caspervdw: See comment in the to_wkt function, I think these should be arguments

pygeos/test/test_readers_writers.py
```diff
@@ -0,0 +1,52 @@
...
+
+from pygeos.ufuncs import from_wkt, from_wkb, to_wkt, to_wkb
+
+
+POINT11_WKB = (
+    b"\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?"
+)
+
+
+def test_from_wkt():
```
caspervdw: There is some of overlap with the wkb/wkt tests in `test_geometry`, ideally these should be merged

caspervdw:
Thanks for this addition! I think the wkt/wkb functions on the Geometry object should either be dropped, or should make use of the ufuncs. I don't like the duplicate code.

I guess the hex input might also be a string indeed. Maybe best to try to parse both bytes and strings in the ufunc.

src/ufuncs.c
```diff
@@ -928,6 +929,211 @@ static void create_collection_func(char **args, npy_intp *dimensions,
...
+    char is_hex;
+
+    /* Create the WKB reader */
+    reader = GEOSWKBReader_create_r(context_handle);
+    if (reader == NULL) {
+        return;
+    }
+
+    UNARY_LOOP {
+        in1 = *(PyObject **)ip1;
```
jorisvandenbossche: @caspervdw can you explain this line? Why the double `**` ?

Getting this right was what I first didn't get to work, but in the end found this in one of the other functions you already wrote with object dtype input.

src/ufuncs.c
jorisvandenbossche: That means that you can't give them default values? (on the ufunc level, can still do that in the python wrapper)

jorisvandenbossche:
Thanks for the comments! 
I already addressed them partly. For `to_wkt`, I already added the parameters. But, that brings up a new question: what should be the defaults? 
For now, I used the defaults as they are in GEOS. But, those are not necessarily the most user friendly defaults (shapely also uses `trim=True` and `output_dimension=3` as default, which deviates from GEOS).

src/ufuncs.c
caspervdw: Yes, I struggled with that too. The `(Pyobject **)` tells the interpreter that `ip1` is a pointer to a `PyObject *`. 

Then we take the value of that (with the extra star at the start), so that `in1` becomes a `PyObject *`.

caspervdw:
I agree that it is better here to deviate from the GEOS defaults. In the `ToWKB` I wrote earlier I used:

```
char trim = 1;
int precision = 6;
int dimension = 3;
int use_old_3d = 0;
```

caspervdw:
The ToWKT part is good to go. I like the docstring!

jorisvandenbossche:
So I will:

- remove the methods on the Geometry object (and see how to use the to_wkt ufunc in the repr of that object, this might be a hard one)
- convert the existing tests for those to use the ufuncs / consolidate it with my tests
- also add python wrappers for the from_wkt/from_wkb/to_wkb functions

That sounds good?

jorisvandenbossche:
For the hex/bytes question, I will add accepting hex-strings instead of only bytes. However, for the actual bytes representation, I would personally only accept bytes, and not strings to be interpreted as bytes (that feels cleaner).

> I think the performance hit here is negligible, so keep it imho

Yes, it's indeed only a minor overhead (2-3% when I checked, but that is also in the uncertainty range). 
A reason to still have it as a keyword instead of automatic detection, is that I would find it very strange to have a mixture of bytes / hex in a single array. I am not sure we should support that.

---

BTW, I did some performance timings in general, and this PR gives exactly the same performance (as expected) as the cython version I wrote before, and it gives around a 20x speed-up for `from_wkb`/`to_wkb` and a 10x speed-up for `from_wkt`/`to_wkt` (for an array of Points, so which is the "easiest" (where the looping is relatively more overhead compared to the actual GEOS call), with bigger geoms the speed-up will be a bit less).

caspervdw:
> However, for the actual bytes representation, I would personally only accept bytes, and not strings to be interpreted as bytes (that feels cleaner).

I agree, WKB should always be `bytes` and never `string`. `WKT` could be both.

> Yes, it's indeed only a minor overhead (2-3% when I checked, but that is also in the uncertainty range). I would find it very strange to have a mixture of bytes / hex in a single array. I am not sure we should support that.

The mixture of both would be odd indeed, but that's no reason to not implement it in this way, right? Maybe adding it as a keyword argument (`hex=True/False`) like in the `ToWKB` is the way to go then. I would be happy with both options.

jorisvandenbossche:
OK, made a bunch of updates:

- added python wrappers for the other methods as well
- added parameters to `to_wkb`
- consolidated the tests (and added a `str(Geometry)` functionality to the GeometryObject to return default WKT)
- return hex as string in `to_wkb`, and accept strings in `from_wkb`

I already updated the tests to no longer use the `Geometry.to/from_wkb/wkt` methods, but did no yet finally remove the methods, because I am running in to the following issue where `Geometry.from_wkb` works, but the ufunc version `pygeos.from_wkb` does not (so the tests that I converted to use the ufuncs are failing):

```
In [1]: import pygeos

In [2]: p = pygeos.points(1, 1)

In [3]: p.to_wkb() == pygeos.to_wkb(p) 
Out[3]: True

In [4]: p.to_wkb(byte_order=0) == pygeos.to_wkb(p, byte_order=0)
Out[4]: True

In [5]: wkb = pygeos.to_wkb(p, byte_order=0)

In [6]: pygeos.Geometry.from_wkb(wkb) 
Out[6]: <pygeos.Geometry POINT (1 1)>

In [7]: pygeos.from_wkb(wkb) 
<ipython-input-7-7d2b49c9458f> in <module>
----> 1 pygeos.from_wkb(wkb)

~/scipy/repos/pygeos/pygeos/io.py in from_wkb(geometry, **kwargs)
    172 
    173     """
--> 174     return lib.from_wkb(geometry, **kwargs)

GEOSException: ParseException: Unexpected EOF parsing WKB
```

This example is with a different byte order, but it also occurs with the default byte order for a different geometry type, eg:

```
In [8]: p = pygeos.Geometry("POLYGON ((0 0, 2 0, 2 2, 0 2, 0 0))") 

In [9]: p.to_wkb() == pygeos.to_wkb(p) 
Out[9]: True

In [10]: wkb = pygeos.to_wkb(p)

In [11]: pygeos.Geometry.from_wkb(wkb) 
Out[11]: <pygeos.Geometry POLYGON ((0 0, 2 0, 2 2, 0 2, 0 0))>

In [12]: pygeos.from_wkb(wkb) 
<ipython-input-12-7d2b49c9458f> in <module>
----> 1 pygeos.from_wkb(wkb)

~/scipy/repos/pygeos/pygeos/io.py in from_wkb(geometry, **kwargs)
    172 
    173     """
--> 174     return lib.from_wkb(geometry, **kwargs)

GEOSException: ParseException: Unexpected EOF parsing WKB
```

but I don't directly see the difference between the code of the ufunc or the object's method ..

jorisvandenbossche:
One additional question regarding naming: currently I am using the `to_wkt` / `from_wkt` pattern. But, it could also be `as_wkt` / `from_wkt` (more similar as the postgis terminology) or `read_wkt` / `write_wkt` (more close to GEOS terminology of WKTReader/Writer)

jorisvandenbossche:
OK, looked into the failure I described above (difference between `ufuncs::from_wkb` and `Geometry.from_wkb`), by printing the wkb chars passed to the GEOS function in c:

```
In [1]: import pygeos 
   ...: p = pygeos.Geometry("POLYGON ((0 0, 2 0, 2 2, 0 2, 0 0))")  
   ...: wkb = pygeos.to_wkb(p)

In [2]: pygeos.from_wkb(wkb) 
Size: 77
01 03 00 00 00 01 00 00 00 05 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
00 00 00 00 00 40 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 40 00 00 00 00 00 00 00 40 00 
00 00 00 00 00 00 00 00 00 00 00 00 00 00 40 
...
GEOSException: ParseException: Unexpected EOF parsing WKB

In [3]: pygeos.Geometry.from_wkb(wkb)
Size: 93
01 03 00 00 00 01 00 00 00 05 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
00 00 00 00 00 40 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 40 00 00 00 00 00 00 00 40 00 
00 00 00 00 00 00 00 00 00 00 00 00 00 00 40 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
Out[3]: <pygeos.Geometry POLYGON ((0 0, 2 0, 2 2, 0 2, 0 0))>
```

So you can see that although the same python object is passed, in the working case there are additional null bytes, while those have been removed in the failing ufunc case. 
And the original python object has indeed those null bytes:

```
In [7]: wkb
Out[7]: b'\x01\x03\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
\x00\x00'

In [8]: len(wkb)
Out[8]: 93
```

And what is more, this only happens for the ufunc if you pass a scalar, not if you actually pass an array:

```
In [6]: pygeos.from_wkb(np.array([wkb], dtype=object))
Size: 93
01 03 00 00 00 01 00 00 00 05 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 40 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 40 00 00 00 00 00 00 00 40 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 40 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
Out[6]: 
array([<pygeos.Geometry POLYGON ((0 0, 2 0, 2 2, 0 2, 0 0))>],
      dtype=object)
```

So there seems to be going something wrong how numpy treats a non-numpy array as input to the ufunc in this case. I don't know if it is something that I am doing wrong, or if it is a bug in numpy. 
But at least it provides a way to workaround it in this case.

jorisvandenbossche:
> I don't know if it is something that I am doing wrong, or if it is a bug in numpy.
But at least it provides a way to workaround it in this case.

Ah, this is because of the default dtype inference of numpy. If you pass a bytes like object to an array (without specifying a dtype), you get a `S` dtyped array, and accessing a single element of that removes trailing null bytes:

```
In [24]: np.array(wkb) 
Out[24]: 
array(b'\x01\x03\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00@',
      dtype='|S93')

In [25]: np.array(wkb).item() 
Out[25]: b'\x01\x03\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@'

In [26]: wkb 
Out[26]: b'\x01\x03\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x00\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

So a roundtrip to numpy `S93` and back to object removes the trailing null bytes.  
Some explanation in https://stackoverflow.com/questions/37435309/numpy-trimming-trailing-zeros-in-byte-strings and https://github.com/numpy/numpy/issues/3878. The reason they do that because the trailing null bytes as end-of-string is the only way to have shorter strings in a fixed-length string dtype, but for our use case this is of course totally unacceptable. So we probably should disallow those string dtypes as input.

jorisvandenbossche:
@caspervdw the ufuncs are now fully working, but a remaining item: can we deduplicate the code between ufuncs.c and pygeom.c ? Because pygeom.c does still need to have basic read/write functionality for the repr and the constructor (but without options). 
But I don't directly see an easy way to reuse the function we defined for the ufunc. You would need to define the arguments in the way the the ufunc function expects them, I don't directly succeed in that.

pygeos/io.py
```diff
@@ -0,0 +1,178 @@
...
+    Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    rounding_precision : int, default 6
+        The rounding precision when writing the WKT string. Set to a value of
+        -1 to indicate the full precision.
+    trim : bool, default True
+        Whether to trim unnecessary decimals (trailing zero's).
```
caspervdw:
```suggestion
        Whether to trim unnecessary decimals (trailing zeros).
```

pygeos/io.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Examples
+    --------
+    >>> to_wkt(Geometry("POINT (0 0)"))
+    'POINT (0 0)'
+    >>> to_wkt(Geometry("POINT (0 0)"), rounding_precision=3, trim=False)
+    'POINT (0.000 0.000)'
+    >>> to_wkt(Geometry("POINT (0 0)"), rounding_precision=-1, trim=False)
+    'POINT (0.0000000000000000 0.0000000000000000)'
+
```
caspervdw: Can you remove the empty line?

pygeos/io.py
```diff
@@ -0,0 +1,178 @@
...
+        Whether the SRID should be included in WKB (this is an extension
+        to the OGC WKB specification).
+
+    Examples
+    --------
+    >>> to_wkb(Geometry("POINT (1 1)"))
+    b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?'
+    >>> to_wkb(Geometry("POINT (1 1)"), hex=True)
+    '0101000000000000000000F03F000000000000F03F'
+
```
caspervdw: Can you remove the empty line?

pygeos/io.py
```diff
@@ -0,0 +1,178 @@
...
+    Parameters
+    ----------
+    geometry : str or array_like
+        The WKB byte object(s) to convert.
+
+    Examples
+    --------
+    >>> from_wkb(b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?')
+    <pygeos.Geometry POINT (1 1)>
+
```
caspervdw: Can you remove the empty line?

pygeos/test/test_io.py
```diff
@@ -0,0 +1,165 @@
...
+def test_from_wkt_all_types(geom):
+    wkt = pygeos.to_wkt(geom)
+    actual = pygeos.from_wkt(wkt)
+    assert pygeos.equals(actual, geom)
+
+
+@pytest.mark.parametrize(
+    "wkt", ("POINT EMPTY", "LINESTRING EMPTY", "GEOMETRYCOLLECTION EMPTY")
+)
+def test_from_wkt_empty(wkt):
```
caspervdw: Can you also make this test for the WKB?  I vaguely remember segfaults when trying to serialize and deserialize empty geometries from an to WKB (especially the POINT EMPTY)

jorisvandenbossche:
> I think we should investigate the overhead if we define part of the Geometry object in pure python. Maybe create the base class in C (with the static _ptr attribute), and subclass it in Python adding extra features such as the to_wkb.

But the ufuncs would still return the base extension type. So do you mean that in the python functions, we would (for those methods that return geometries) loop through it and wrap each element in the python class?

caspervdw:
Maybe we could make a kind of accessor like:

Geometry.obj

That creates a Python object with all the bells and whistles

jorisvandenbossche:
Will update for your comments when having some time this week.

One more question about the naming: I now used `to_wkt` / `from_wkt` pattern. But, it could also be `write_wkt` / `read_wkt` (in line of GEOS's Writer/Reader classes) or `as_wkt` / `to_wkt` (more like postgis).

caspervdw:
I don't like the "read" and "write" namings, as we also do not speak of "readers" and "writers" (that is a bit C-ish). We just convert representations of geometries. 

I like the idea of aligning with PostGIS `as_wkt` / `from_wkt`, but Python packages often use the `to_` and `from_` prefixes (e.g. `pandas.DataFrame.to_csv` or `numpy.ndarray.to_list`)

So my preference is keeping it as is. Also the least amount of work.

jorisvandenbossche:
One more question: are you OK with dropping WKB input in the `Geometry(..)` constructor? 
For WKT it is convenient to be able to do `Geometry("POINT (1 1)"`), but for WKB I think people can just use `from_wkb` ? (then only the WKT io has some duplication in pygeom.c)

jorisvandenbossche:
- cleaned up empty lines
- splitted tests (not to a single assert per test, but at least more splitted)
- added test for empty geometries (empty point does not work for WKB, as expected, but does still give a wrong error (MemoryError), so that is still a TODO)
- removed the methods on the Geometry object (and simplified the WKT code, since that doesn't need to handle with options any more -> that is covered by pygeos.to/from_wkt)

jorisvandenbossche:
And the "POINT EMPTY" error in to_wkb is now also fixed: it still raises an error, but now the proper one with a informative error message:

```
In [2]: geom = pygeos.Geometry("POINT EMPTY") 

In [3]: pygeos.to_wkb(geom)
<ipython-input-3-2272ebf49184> in <module>
----> 1 pygeos.to_wkb(geom)

~/scipy/repos/pygeos/pygeos/io.py in to_wkb(geometry, hex, output_dimension, byte_order, include_srid, **kwargs)
    127         np.intc(byte_order),
    128         np.bool(include_srid),
--> 129         **kwargs,
    130     )
    131 

GEOSException: IllegalArgumentException: Empty Points cannot be represented in WKB

```

PR #47, opened 2019-09-30
[WIP] Convert shapely geometries

jorisvandenbossche:
This PR adds a `from_shapely` functionality, to convert an array of shapely objects into pygeos objects. 
This is something I will often need to use pygeos in GeoPandas (since all (testing) code their is written to handle shapely input).

Work in progress for now, the None handling is not yet working.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ class build_ext(_build_ext):
 
 
 module_ufuncs = Extension(
-    "pygeos.ufuncs", sources=["src/ufuncs.c", "src/pygeom.c"], **get_geos_paths())
+    "pygeos.ufuncs", sources=["src/ufuncs.c", "src/pygeom.c", "src/io.c"], **get_geos_paths())
 
 
 try:
--- a/src/io.c
+++ b/src/io.c
@@ -1,0 +1,139 @@
+#define PY_SSIZE_T_CLEAN
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+
+#include <Python.h>
+#include <math.h>
+
+#define NO_IMPORT_ARRAY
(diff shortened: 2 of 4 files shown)


caspervdw:
Do we really need to go through C here? We will have the interpreter overhead anyway as shapely stores its GEOSGeometry pointer in a (dynamic) attribute.

jorisvandenbossche:
How would you convert a shapely object into a pygeos object on the python level? (without going through WKB)

caspervdw:
> How would you convert a shapely object into a pygeos object on the python level? (without going through WKB)

Calling the GEOSClone function on a pointer originating from a Python int is asking for problems. It will be easy to get segfaults. Also I am thinking about GEOS versioning issues.

I was thinking about WKB indeed. In general I like using stable API protocols like that. Does it add that much overhead?

jorisvandenbossche:
> Calling the GEOSClone function on a pointer originating from a Python int is asking for problems.

That's what Shapely is doing all the time .. 
But of course, they know that the GEOS function they are calling comes from the same GEOS installation as with which they created the geometry. And indeed, in our case you could end up with a different GEOS .. (one included in the shapely wheel, and pygeos using another one for some reason). I didn't think of that. As long as it are two separate projects, we should probably indeed not do this.

Will do a check of performance difference between both options.

jorisvandenbossche:
I did a quick check, and to give an idea of the speed-up, for an array of 100000 shapely points, going through WKB takes ~960ms, while the method in this PR (cloning) takes ~40ms.

But ok, for now this is not a blocker, and agree that GEOS versioning issues makes this problematic. But when we think again about better integration with shapely, this might something to reconsider.

PR #48, opened 2019-09-30
Refactor ufuncs module into lib module

jorisvandenbossche:
@caspervdw I started this afternoon to try to split the files a bit more and have a separate ufuncs file and a general `lib` module instead of `ufuncs` (since not everything are ufuncs, I think mentioned it before on some PR, not sure if you are OK with that?).

Then I saw that you also did some changes to move out geos things into geos.c / pygeom.c in the coordinates PR, and included that here as well to see if all is working.

One thing that I ran into is that the ufunc registry functions don't copy the passed arguments, and you need to ensure that all those objects are kept alive as long as the ufunc objects exist. That caused segfaults when calling `init_ufuncs` defined in `ufuncs.c` from `lib.c` (as all those arguments where not declared in ufuncs.h, only the init function, I suppose). For now I "fixed" this by just having a ufuncs.h and including that fully instead of a ufuncs.c + ufuncs.h.

### DIFF
--- a/pygeos/__init__.py
+++ b/pygeos/__init__.py
@@ -1,6 +1,6 @@
-from .ufuncs import GEOSException  # NOQA
-from .ufuncs import Geometry  # NOQA
-from .ufuncs import geos_version  # NOQA
+from .lib import GEOSException  # NOQA
+from .lib import Geometry  # NOQA
+from .lib import geos_version  # NOQA
 from .geometry import *
 from .creation import *
 from .constructive import *
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -1,7 +1,7 @@
 from enum import IntEnum
 import numpy as np
 from . import Geometry  # NOQA
-from . import ufuncs
+from . import lib
 
 
 __all__ = [
@@ -54,7 +54,7 @@ def boundary(geometry, **kwargs):
     >>> boundary(Geometry("MULTIPOINT (0 0, 1 2)")) is None
     True
     """
-    return ufuncs.boundary(geometry, **kwargs)
+    return lib.boundary(geometry, **kwargs)
 
 
 def buffer(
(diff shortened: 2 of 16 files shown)


jorisvandenbossche:
@mattip or @seberg do you have any experience with this? What I am trying to do is to register ufuncs defined in one .c file (`ufuncs.c`) to add them to the C extension module and register them with numpy in another .c file (`lib.c`). Should that be possible? (getting segfaults right now) 
Main reason I am trying to do that is for better code organisation / smaller files.

seberg:
Hmmm, I do not see why that should not work, but probably I am missing some linking issue. How does it segfault? The way the ufunc API works is by populating a struct with function pointers.

jorisvandenbossche:
It segfaults on import.

jorisvandenbossche:
(small note: what is in the diff here works. It's only when I have `ufuncs.c` with all data and functions and `ufuncs.h` with just the `init_ufuncs` declaration to include in `lib.c` that it segfaults)

jorisvandenbossche:
@seberg Ah, found it. I also need to guard for the ufunc API (`#define NO_IMPORT_UFUNC`, in addition to `NO_IMPORT_ARRAY`)

jorisvandenbossche:
@caspervdw since you said elsewhere you were OK with the rename of `ufuncs` to `lib`, I am going to merge this, so I can fix the merge conflicts with your coords PR

The main difference is the return of 0 or 1. For now, I left it as is (so not changing to return 1 as you did it in your PR), as this seems to be how all numpy examples do it. But we can of course still discuss / change in a separate PR.

caspervdw:
Thanks @jorisvandenbossche  for pushing this. 

If you are talking about the change in the return value of `init_geos` or `init_geom_type`; I think I changed that to 0 (error) or 1 (OK) indeed. That lead to slightly nicer syntax. But if numpy does it like -1 (error), 0 (OK), then apparently that is customary so :+1: on keeping it. It does matches exit codes now I think about it.

jorisvandenbossche:
>  It does matches exit codes now I think about it.

I indeed matches that, but then is inconsistent with all other functions that return 0 / NULL on exception ..
But so the rule is then for the `init_` functions we follow the exit-code style, for all other things we follow the more natural 0 on exception style.

PR #53, opened 2019-10-08
Finish API docs

caspervdw:
I think we might be heading to a next release. It is good to have the API docs finished then.

edit: I renamed some functions to stick with PostGIS namings, those are the namings most GIS experts know.

edit2: the shared_paths functions needs some additional attention (I think it should output 2 arrays or an Nx2 array of geometries), I removed it for now

### DIFF
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -28,8 +28,8 @@ def line_interpolate_point(line, distance, normalize=False):
     <pygeos.Geometry POINT (0 8)>
     >>> line_interpolate_point(line, [0.25, -0.25], normalize=True).tolist()
     [<pygeos.Geometry POINT (0 4)>, <pygeos.Geometry POINT (0 8)>]
-    >>> line_interpolate_point(Geometry("LINESTRING EMPTY"), 1) is None
-    True
+    >>> line_interpolate_point(Geometry("LINESTRING EMPTY"), 1)
+    <pygeos.Geometry POINT EMPTY>
     """
     if normalize:
         return lib.line_interpolate_point_normalized(line, distance)
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -14,7 +14,7 @@ def line_interpolate_point(line, distance, normalize=False):
         Negative values measure distance from the end of the line. Out-of-range
         values will be clipped to the line endings.
     normalize : bool
-        If normalizei is set to True, the distance is a fraction of the total
+        If normalize is set to True, the distance is a fraction of the total
         line length instead of the absolute distance.
 
     Examples
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -1,28 +1,90 @@
 from . import lib
+from . import Geometry  # NOQA
 
-__all__ = ["interpolate", "line_merge", "project", "shared_paths"]
+__all__ = ["line_interpolate_point", "line_locate_point", "line_merge"]
 
 
-def interpolate(line, normalize=False):
+def line_interpolate_point(line, distance, normalize=False):
+    """Returns a point interpolated at given distance on a line.
+
+    Parameters
+    ----------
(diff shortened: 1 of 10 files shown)


pygeos/linear.py
```diff
@@ -1,28 +1,90 @@
...
+    """Returns a point interpolated at given distance on a line.
+
+    Parameters
+    ----------
+    line : Geometry or array_like
+    distance : float or array_like
+        Negative values measure distance from the end of the line. Out-of-range
+        values will be clipped to the line endings.
+    normalize : bool
+        If normalizei is set to True, the distance is a fraction of the total
```
jorisvandenbossche:
```suggestion
        If normalize is set to True, the distance is a fraction of the total
```

pygeos/linear.py
```diff
@@ -1,28 +1,90 @@
...
+    """Returns (multi)linestrings formed by combining the lines in a
+    multilinestrings.
+
+    Parameters
+    ----------
+    line : Geometry or array_like
+
+    Examples
+    --------
+    >>> line = Geometry("LINESTRING(0 2, 0 10)")
```
jorisvandenbossche: this line is not used in the rest of the examples

src/ufuncs.c
```diff
@@ -225,8 +225,26 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
 static PyUFuncGenericFunction Y_Y_funcs[1] = {&Y_Y_func};
 
 /* Define the geom, double -> geom functions (Yd_Y) */
-static void *interpolate_data[1] = {GEOSInterpolate_r};
-static void *interpolate_normalized_data[1] = {GEOSInterpolateNormalized_r};
+static void *GEOSInterpolateProtectEmpty_r(void *context, void *geom, double d) {
+    int n = GEOSGeomGetNumPoints_r(context, geom);
+    if (n < 2) {
+        /* return NULL for empty linestrings to protect against a segfault */
```
jorisvandenbossche: Should this be reported to GEOS, or is it documented that the user needs to check this guarantee?

src/ufuncs.c
```diff
@@ -225,8 +225,26 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
...
+        return NULL;
```
jorisvandenbossche: Other option could also be to return an empty point? (although I think empty points are not very well supported?) 
In general it's always a bit difficult to reason about which of the two is best. But maybe a good rationale is that if GEOS does not return an empty geometry, we shouldn't create one.

jorisvandenbossche:
A general remark on the API docs (but which is not specific to this PR, can also open an issue about it): I personally prefer to document the functions matching the recommended usage.

I *think* we recommend everybody to use `pygeos.union(..)` (or `from pygeos import union`), and not to use `pygeos.set_operations.union(..)` (or `from pygeos.set_operations import union`) ? 
In that case, I think it is clearer for the user to also document that function as `pygeos.union` (the signature at https://pygeos.readthedocs.io/en/latest/set_operations.html#pygeos.set_operations.union gives the long name)

Of course, documenting it as it is done now by module gives a nice structure to the docs. And it makes it easier to do this in sphinx (we can use the `.. automodule::` feature, instead of needing to list the functions per module manually)

src/ufuncs.c
caspervdw: I don't know what I would prefer. But I think if GEOS doesn't create an empty geometry, we should do it either.

caspervdw: Check out https://github.com/libgeos/geos/pull/126 , it is already solved upstream. I guess we have to keep the check for backwards compat, but do return an empty point.

jorisvandenbossche:
Updates look good! Maybe add a test case for the empty linestring one? (I know there is a doctest, but I think it is good practice to not rely on those for coverage)

PR #57, opened 2019-10-13
ENH: add equals_exact predicate

jorisvandenbossche:
This adds a `equals_exact` predicate. 
I thought the python wrapper for the ufunc was missing for this one, so added it to `predicates.py`, but only then saw that it is actually exposed through `equals` with a tolerance argument. 
But since I already did the work, directly opening the PR instead of an issue.

I would argue that it is better to have a separate `equals_exact` instead of combining it in `equals`:

- There are doing something different: spatial equality (based on the DE-9IM model) vs exact coordinate equality. Therefore, there are some differences between both functions to be aware of as a user (eg different ordering of interiors or different direction of the ring). It's not just adding `tolerance` to the `equals` (eg `equals_exact(.., tolerance=0)` is not the same as `equals(..)`)
- Other libraries also seem to make this distinction. Starting with GEOS of course, but also shapely and postgis do

### DIFF
--- a/pygeos/predicates.py
+++ b/pygeos/predicates.py
@@ -24,6 +24,7 @@ __all__ = [
     "overlaps",
     "touches",
     "within",
+    "equals_exact",
 ]
 
 
@@ -490,7 +491,6 @@ def equals(a, b, tolerance=0.0, **kwargs):
     a, b : Geometry or array_like
     tolerance : float or array_like
 
-
     Examples
     --------
     >>> line = Geometry("LINESTRING(0 0, 5 5, 10 10)")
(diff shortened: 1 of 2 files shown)


pygeos/predicates.py
```diff
@@ -635,3 +633,43 @@ def within(a, b, **kwargs):
...
+    Examples
+    --------
+    >>> point1 = Geometry("POINT(50 50)")
+    >>> point2 = Geometry("POINT(50.1 50.1)")
+    >>> equals_exact(point1, point2)
+    False
+    >>> equals_exact(point1, point2, tolerance=0.2)
+    True
+    >>> equals_exact(point1, None, tolerance=0.2)
+    True
```
caspervdw:
```suggestion
    False
```

jorisvandenbossche:
This is passing now, thanks for the catch

PR #58, opened 2019-10-16
ENH: Add STRtree

### LINKED ISSUE #9: Wrap the STRTree

caspervdw:
Closes #9

### DIFF
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -15,4 +15,10 @@ class STRtree:
         self._tree = lib.STRtree(np.asarray(geometries, dtype=np.object), leafsize)
 
     def query(self, envelope):
+        """Return all items whose extent intersect the given envelope.
+
+        Parameters
+        ----------
+        envelope : Geometry
+        """
         return self._tree.query(envelope)
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -29,6 +29,9 @@ class STRtree:
     def __init__(self, geometries, leafsize=5):
         self._tree = lib.STRtree(np.asarray(geometries, dtype=np.object), leafsize)
 
+    def __len__(self):
+        return self._tree.count
+
     def query(self, envelope):
         """Return all items whose extent intersect the given envelope.
 
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -22,3 +22,7 @@ class STRtree:
         envelope : Geometry
         """
         return self._tree.query(envelope)
+
+    @property
+    def geometries(self):
+        return self._tree.geometries
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -8,7 +8,22 @@ class STRtree:
     """A query-only R-tree created using the Sort-Tile-Recursive (STR)
     algorithm.
 
-    For two-dimensional spatial data.
+    For two-dimensional spatial data. The actual tree will be constructed at the first
+    query.
+
+    Parameters
+    ----------
+    geometries : array_like
+    leafsize : int
+        the maximum number of child nodes that a node can have
+
+    Examples
+    --------
+    >>> import pygeos
+    >>> geoms = pygeos.points(np.arange(10), np.arange(10))
+    >>> tree = pygeos.STRtree(geoms)
+    >>> tree.query(pygeos.box(2, 2, 4, 4)).tolist()
+    [2, 3, 4]
     """
 
     def __init__(self, geometries, leafsize=5):
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -1,0 +1,18 @@
+from pygeos import lib
+import numpy as np
+
+__all__ = ["STRtree"]
+
+
+class STRtree:
+    """A query-only R-tree created using the Sort-Tile-Recursive (STR)
+    algorithm.
+
+    For two-dimensional spatial data.
+    """
+
+    def __init__(self, geometries, leafsize=5):
+        self._tree = lib.STRtree(np.asarray(geometries, dtype=np.object), leafsize)
+
+    def query(self, envelope):
+        return self._tree.query(envelope)
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -35,6 +35,7 @@ class STRtree:
         Parameters
         ----------
         envelope : Geometry
+            The envelope of the geometry is taken automatically.
         """
         return self._tree.query(envelope)
 
--- a/pygeos/test/test_strtree.py
+++ b/pygeos/test/test_strtree.py
@@ -13,7 +13,8 @@ def tree():
 
 
 def test_init_with_none():
-    pygeos.STRtree(np.array([None]))
+    tree = pygeos.STRtree(np.array([None]))
+    assert tree.query([point]).size == 0
 
 
 def test_init_with_no_geometry():
--- a/pygeos/test/test_strtree.py
+++ b/pygeos/test/test_strtree.py
@@ -3,23 +3,7 @@ from pygeos import box
 import pytest
 import numpy as np
 from numpy.testing import assert_array_equal
-import sys
-from contextlib import contextmanager
-from .common import point, empty
-
-
-@contextmanager
-def assert_increases_refcount(obj):
-    before = sys.getrefcount(obj)
-    yield
-    assert sys.getrefcount(obj) == before + 1
-
-
-@contextmanager
-def assert_decreases_refcount(obj):
-    before = sys.getrefcount(obj)
-    yield
-    assert sys.getrefcount(obj) == before - 1
+from .common import point, empty, assert_increases_refcount, assert_decreases_refcount
 
 
 @pytest.fixture
@@ -34,13 +18,29 @@ def test_init_with_none():
 
 def test_init_increases_refcount():
     arr = np.array([point])
-    with assert_increases_refcount(point):
+    with assert_increases_refcount(point), assert_increases_refcount(arr):
         _ = pygeos.STRtree(arr)
 
 
 def test_del_decreases_refcount():
     arr = np.array([point])
     tree = pygeos.STRtree(arr)
+    with assert_decreases_refcount(point), assert_decreases_refcount(arr):
+        del tree
+
+
+def test_geometries_property():
+    arr = np.array([point])
+    tree = pygeos.STRtree(arr)
+    assert arr is tree.geometries
+
+
+def test_flush_geometries(tree):
+    arr = np.array([point])
+    tree = pygeos.STRtree(arr)
+    # You shouldn't change this array inplace
+    arr[:] = None
+    # But still it does not lead to a memory leak
     with assert_decreases_refcount(point):
         del tree
 
--- a/pygeos/test/test_strtree.py
+++ b/pygeos/test/test_strtree.py
@@ -1,0 +1,5 @@
+import pygeos
+
+
+def test_init():
+    pygeos.lib.STRtree(5)
--- a/pygeos/test/test_strtree.py
+++ b/pygeos/test/test_strtree.py
@@ -1,4 +1,5 @@
 import pygeos
+import pytest
 import numpy as np
 import sys
 from contextlib import contextmanager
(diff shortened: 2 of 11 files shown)


caspervdw:
There was some design decision to make here. I chose to let the GEOS strtree store a pointer to the actual GeometryObject, so that it can DECREF it before deallocation.

The other options would be copying the GEOSGeometry object at the strtree init, and destroying it on deallocation. I think this copy would give unwanted overhead.

caspervdw:
@jorisvandenbossche  Do you know if there is any meaning to the order that is returned by tree.query? It might get inversed in my implementation.

pygeos/strtree.py
```diff
@@ -0,0 +1,24 @@
...
+        self._tree = lib.STRtree(np.asarray(geometries, dtype=np.object), leafsize)
+
+    def query(self, envelope):
+        """Return all items whose extent intersect the given envelope.
+
+        Parameters
+        ----------
+        envelope : Geometry
+        """
+        return self._tree.query(envelope)
```
jorisvandenbossche: In shapely's version of STRTree, this is also the design to return the actual geometries. 
But for me, that has almost always been a reason to *not* use it (compared to rtree), as in most practical applications I needed indexes (of course because I was specifically working with dataframes, and not just geometries, so in those cases you often want to get the row of the returned geometry eventually, and then you need an index)

src/strtree.c
```diff
@@ -0,0 +1,175 @@
...
+        return NULL;
+    }
+
+    n = PyArray_SIZE((PyArrayObject *) arr);
+    for(i = 0; i < n; i++) {
+        /* get the geometry */
+        ptr = PyArray_GETPTR1((PyArrayObject *) arr, i);
+        obj = *(GeometryObject **) ptr;
+        /* skip incase obj was no geometry or None */
+        if (!get_geom(obj, &geom)) { continue; }
```
jorisvandenbossche: Shouldn't we rather raise an error in case of not geometry/None? (that's what we do for all ufuncs I think?)

jorisvandenbossche:
> There was some design decision to make here. I chose to let the GEOS strtree store a pointer to the actual GeometryObject, so that it can DECREF it before deallocation.
The other options would be copying the GEOSGeometry object at the strtree init, and destroying it on deallocation. I think this copy would give unwanted overhead.

Yeah, I don't know how costly cloning the GEOSGeometry would be. The increasing/decreasing of the ref count in the current code seems relatively straightforward (I am trying to think of cases where this could give problems, but not directly thinking of something).


> Since there are probably good use cases of directly iterating over the geometries as well, it might make sense to return both? (eg with an option to toggle between both)

Looking at the code, that might not be that easy, as the object that is returned is actually inserted in the tree. So either you need 2 implementations the tree creation (creating code duplication), or either you need to convert the integer to the geometry in the query method (giving additional indexing overhead in case you only want the geometry)

src/strtree.h
```diff
@@ -0,0 +1,25 @@
...
+#define _RTREE_H
+
+#include <Python.h>
+#include "geos.h"
+
+
+typedef struct {
+    PyObject_HEAD
+    void *ptr;
+} STRtree;
```
jorisvandenbossche: STRTreeObject to be consistent with GeometryObject ?

pygeos/test/test_strtree.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def test_query_no_geom(tree):
+    with pytest.raises(TypeError):
+        tree.query("I am not a geometry")
+
+
+def test_query_none(tree):
+    with pytest.raises(TypeError):
+        tree.query(None)
```
jorisvandenbossche: Should we return an empty array instead? Or have an option for it? 
Just thinking aloud (it might be useful, as you typically need to handle the empty case anyway, as valid geometries can also return that)

pygeos/strtree.py
caspervdw: Returning indices instead of the geometries itself seems to be a better approach indeed. That seems to be how scipy's KDTree does it:
https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.spatial.KDTree.html

Then we would have to store the original 1D ndarray on the STRtree object, dereference that on deallocation (which would in turn dereference the contained geometries).

A big downside would be that we would have memory leaks / segfaults if one changes the array after constructing the STRtree. If we do this, we should put an explicit warning that this array should not be changed in the docs. By the way, also how scipy does it: "This array is not copied, and so modifying this data will result in bogus results."

src/strtree.c
caspervdw: Agreed. Not geometry -> error ; None -> empty array

pygeos/test/test_strtree.py
caspervdw: Agreed, see above. As `pygeos.intersects(geometry, None) == False`, I think returning an empty list here would indeed make more sense.

caspervdw:
@jorisvandenbossche  I took another shot at this. Now `query` returns the indices, which can be used directly on `tree.geometries` to get the geometries.

To have the garbage collecting write, I still needed to insert the actual geometry objects into the tree. So I defined a struct `STRtreeElem` that contains the index (for returning) and the geometry object (for garbage collecting).

jorisvandenbossche:
Did a quick experiment with combining the STRTree with a contains predicate (1 million points /1 polygon), and here I get a speed-up of 2-3x by using a spatial index (assuming we have a "warm" tree): https://nbviewer.jupyter.org/gist/jorisvandenbossche/66d1d81935cf6084d77ae6d1158f6fae 
In practice, it will of course depend a lot on the specifics (how many elements are be returned by the query (in this experiment it is 1%), how costly is the actual operation, ...). But this modest speed-up at least indicates it's not totally useless ;) (and there will be cases where it can be much more interesting)

One thing I noticed is that the first usage of the tree is actually very slow (more than 100x slower). It might be that GEOS only constructs the actual tree on first use? (not sure how I would otherwise interpret that) 

If that is correct, it seems that creation of the tree is actually quite slow .. (outside of our control of course, but can limit practical usage)

Anyway, the current interface certainly seems to be useful / practical!

jorisvandenbossche:
> To have the garbage collecting right, I still needed to insert the actual geometry objects into the tree. So I defined a struct STRtreeElem that contains the index (for returning) and the geometry object (for garbage collecting).

Why is that? I understand from the C API that the caller remains ownership on the GEOSGeometry objects? Do we need to keep the geometries alive as long as the STRTree is alive?

caspervdw:
> One thing I noticed is that the first usage of the tree is actually very slow (more than 100x slower). It might be that GEOS only constructs the actual tree on first use? (not sure how I would otherwise interpret that)

You're correct, it gets constructed on the first query. There's one sentence in the docs from which you can derive that `However, once the tree has been built (explicitly or on the first call to query), items may not be added or removed.` (see https://geos.osgeo.org/doxygen/classgeos_1_1index_1_1strtree_1_1STRtree.html)

> Why is that? I understand from the C API that the caller remains ownership on the GEOSGeometry objects? Do we need to keep the geometries alive as long as the STRTree is alive?

We can't ensure that the caller doesn't dereference the geometry array with which the STRtree is constructed. If that happens, we we get a crash. You need to keep references to the geometries on the tree object to solve this issue.

jorisvandenbossche:
Hmm, it's a pity that a trigger to construct the tree is not included in the C-API (as I think it can be useful to control as a user when the tree gets build. So eg that you can do it in advance, and the first query will already be fast).

> We can't ensure that the caller doesn't dereference the geometry array with which the STRtree is constructed. If that happens, we we get a crash. You need to keep references to the geometries on the tree object to solve this issue.

It seems that you both keep a reference to the array, and then also store each geometry of that array inside the tree. Is the first (keeping a reference to the array) not enough?

But to understand: so the geometries need to be kept alive as long as the tree exists? 
I don't fully understand why that is. From the code, when inserting an item in the tree, it seems it only uses the envelope of the geometry, and for the rest does not store or reference it (https://github.com/libgeos/geos/blob/master/capi/geos_ts_c.cpp#L5939)

caspervdw:
> It seems that you both keep a reference to the array, and then also store each geometry of that array inside the tree. Is the first (keeping a reference to the array) not enough?
> But to understand: so the geometries need to be kept alive as long as the tree exists?
> I don't fully understand why that is. From the code, when inserting an item in the tree, it seems it only uses the envelope of the geometry, and for the rest does not store or reference it (https://github.com/libgeos/geos/blob/master/capi/geos_ts_c.cpp#L5939)

I did some tests, and it appears that you need to keep a reference to the geometry object until the first query has been executed. In no case I got a hard crash, worst case there are geometries missing from the result of the `query`. See below:

```
geometries = pygeos.points(np.arange(10), np.arange(10))
tree = pygeos.STRtree(geometries)
geometries[1] = None
tree.query(pygeos.box(0, 0, 2, 2))
>>> [0, 2]  # unexpected
```

```
geometries = pygeos.points(np.arange(10), np.arange(10))
tree = pygeos.STRtree(geometries)
tree.query(pygeos.box(0, 0, 2, 2))
geometries[1] = None
tree.query(pygeos.box(0, 0, 2, 2))
>>> [0, 1, 2]  # correct
```

Keeping the objects in the `tree.geometries` array is not 100% foolproof, as it could be changed inplace. But I think it is good enough, as long as we document that changing the array after constructing the tree will lead to bogus results (as is done in scipy.KDTree)>

jorisvandenbossche:
> But I think it is good enough, as long as we document that changing the array after constructing the tree will lead to bogus results

That sounds good to me as well

pygeos/strtree.py
```diff
@@ -0,0 +1,28 @@
...
+
+__all__ = ["STRtree"]
+
+
+class STRtree:
+    """A query-only R-tree created using the Sort-Tile-Recursive (STR)
+    algorithm.
+
+    For two-dimensional spatial data.
+    """
```
jorisvandenbossche: Maybe add a `Parameters` section here (for geometries and leafsize of the init)

pygeos/strtree.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, geometries, leafsize=5):
+        self._tree = lib.STRtree(np.asarray(geometries, dtype=np.object), leafsize)
+
+    def query(self, envelope):
+        """Return all items whose extent intersect the given envelope.
+
+        Parameters
+        ----------
+        envelope : Geometry
```
jorisvandenbossche: You can pass whatever geometry, and the envelope is taken under the hood? (it's not that you first need to take the envelope yourself, but the current doc is not fully clear about it)

pygeos/strtree.py
jorisvandenbossche: And maybe also a small example?

```

>>> geoms = pygeos.points(np.arange(10), np.arange(10)) 
>>> tree = pygeos.STRtree(geoms)  
>>> tree.query(pygeos.box(2, 2, 4, 4))
array([2, 3, 4])
```

pygeos/test/test_strtree.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@pytest.fixture
+def tree():
+    geoms = pygeos.points(np.arange(10), np.arange(10))
+    yield pygeos.STRtree(geoms)
+
+
+def test_init_with_none():
+    pygeos.STRtree(np.array([None]))
```
jorisvandenbossche: Maybe also do an actual query here? (since the tree is actually only created on first query, that seems relevant for the all-None case that also this query doesn't segfault or so)

jorisvandenbossche:
So the problem is that it is failing with creating/querying the tree if it was empty (not geometries or all None), and for older GEOS? 

It might be this is a bug in GEOS that was fixed? 
We could maybe return try to detect this case and return an error in that case when creating the tree? (eg keep a counter while inserting into the tree, and raise an error if the counter is 0 ?)

caspervdw:
I did the work already to keep track of the count, only forgot to put the logic in place at the query()

Will have a look next time around (might take a while- feel free to wrap this up)

jorisvandenbossche:
Ah, yes, I see there is already a count. Yep, I can wrap this up. 

But do we always want to raise an error for an empty tree? Or only for older GEOS versions? Always is the easiest, but it could be useful to have it also work for empty trees (with recent GEOS it does the right thing).

caspervdw:
I think it is best to return an empty array (of the correct type) when the tree count is zero. Checking the count is very cheap so I would not make a version conditional there.

jorisvandenbossche:
Ah, I missed that as well ;) You are already using the count to return this empty array. OK, yes, that's a much better solution!

jorisvandenbossche:
It turned out you already had everything in place, but it was not used because the count was not initialized ..

jorisvandenbossche:
This is all green now. So going to merge! (we can always further refine with follow-up PRs if needed)

PR #59, opened 2019-10-20
Try different Travis script

caspervdw:


.travis.yml
```diff
@@ -13,24 +13,15 @@ matrix:
...
-  - conda update --yes conda;
-  - conda config --set changeps1 no --set restore_free_channel true;
-  - conda config --append channels conda-forge
-  - conda create -n testenv --yes python=$PYTHON_VERSION $DEPS pytest;
-  - conda activate testenv;
-  - conda --version ; python --version ; pip --version;
+  - wget https://repo.anaconda.com/pkgs/misc/conda-execs/conda-latest-$TRAVIS_OS_NAME-64.exe -O conda.exe
+  - chmod +x conda.exe
+  - export CONDA_ALWAYS_YES=1
+  - ./conda.exe config --set changeps1 no --set restore_free_channel true;
```
jorisvandenbossche: Any idea if the `restore_free_channel` is (still) needed? 
Is it for the older geos versions?

.travis.yml
```diff
@@ -13,24 +13,15 @@ matrix:
...
+  - ./conda.exe config --append channels conda-forge
```
jorisvandenbossche: Conda-forge recommends to "add" instead of "append" it (put it at the top of the channel list). Is there a reason to append it here?

jorisvandenbossche:
Just some minor questions, a green CI is more important ;)

.travis.yml
caspervdw: Yes, it's for the older GEOS versions

.travis.yml
caspervdw: I always like to be the default channel the highest priority one. But apart from that, no good reason

jorisvandenbossche: For just numpy and geos, that is probably fine since there cannot be any incompatibility (geos is not built against numpy). But in general it is recommended to have conda-forge first once you have conda-forge packages that are not pure python (because the conda-forge ones are not always binary compatible with defaults, so if you have a binary packages, all its dependent binary packages should also come from conda-forge)

.travis.yml
caspervdw: Not so sure actually, the default channel seems to list GEOS 3.5, which is the earliest version. Testing without !

.travis.yml
caspervdw: Right, I did run into this issues from time to time. Good to now there is actually a recommendation that would help ;)

I switched it to optional for only the newest build, let's see if it works.

caspervdw: When changing the --append to --add on the appveyor build, it actually breaks the build. It installs pytest 3.2.2 instead of 5.2

jorisvandenbossche:
> When changing the --append to --add on the appveyor build, it actually breaks the build. It installs pytest 3.2.2 instead of 5.2

Ah, that was specifically for Python 3.5, and that's because conda-forge does not build packages anymore for 3.5 (so therefore only getting old packages).

Sidenote: many packages are starting to drop 3.5 support (eg pandas will do so in the next release), so we can also think about that

Issue #54, opened 2019-10-09
Conversion to Shapely objects


jorisvandenbossche:
For GeoPandas (and for interoperability with Shapely), I need a way to convert geometries from pygeos to Shapely and vice versa.

As discussed in https://github.com/pygeos/pygeos/pull/47, one way is to go through WKB. In that PR I had a way to directly clone the GEOSGeometry pointer, but if pygeos and shapely are not using the same GEOS library, this can cause problems.

However, going through WKB is also not fully faithful. For example, empty points don't have a WKB representation.

caspervdw:
After all I think the performance difference is so large that we can take the risk of incompatible versions. We can just put a warning in the docstring?

I was just working on a `from_shapely` ufunc, for performance comparison, and I find 3.7 ms for `from_shapely` and 123 ms going through WKB.

jorisvandenbossche:
Maybe there would be a way to detect that both are using the same GEOS library, and only do it then.

brendan-ward:
The conversion via pointers to GEOS geometries underneath `shapely` objects was implemented in #61 and works great other than incompatible GEOS versions as mentioned in #122.

Changed the title to reflect that this is about conversion.

The conversion to `shapely` objects is already handled out in `geopandas` where the `shapely` library is available as a core dependency.  In order to provide same here, we'd have to include `shapely` as an optional dependency.

@jorisvandenbossche should this be handled in `pygeos` or downstream libraries as you've done for `geopandas`?

jorisvandenbossche:
As reference, the code in GeoPandas for this conversion is here:

https://github.com/geopandas/geopandas/blob/1b0f006ba1a07f7896d7e5854fa5954ca1d83213/geopandas/_vectorized.py#L61-L74

Given that this uses some semi-private APIs from shapely, it would be nice to not have each downstream library do this, and rather include a `to_shapely` method in pygeos (but given the work to integrate it in shapely itself, which renders this method useless, this is also not the biggest priority for myself)

brendan-ward:
@jorisvandenbossche presumably pygeos merging with shapely to make shapely 2.0 also renders this unnecessary here, but would be handled as part of that process?  I.e., there would just be one set of binary objects involved at that point, instead of 2 as now.  Short term conversion only seems necessary to use pygeos in geopandas in the meantime?

jorisvandenbossche:
[...]

PR #61, opened 2019-10-21
ENH: Implement from_shapely

caspervdw:
Following discussions in #47 and #54 

@jorisvandenbossche  I took the ufunc approach, and implemented some checks. I think 'segfault and memoryleak protection' is now as good as shapely. 

Conversion of a list of 10000 shapely geometries takes 4 ms with this function. against 120 ms going through WKB.

### DIFF
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -213,7 +213,9 @@ def from_shapely(geometry, **kwargs):
     if isinstance(geometry, ShapelyGeometry):
         # this so that the __array_interface__ of the shapely geometry is not
         # used, converting the Geometry to its coordinates
-        geometry = np.array([geometry], dtype=np.object).reshape(())
+        arr = np.empty(1, dtype=object)
+        arr[0] = geometry
+        arr.shape = ()
     else:
-        geometry = np.asarray(geometry, dtype=object)
-    return lib.from_shapely(geometry, **kwargs)
+        arr = np.asarray(geometry, dtype=object)
+    return lib.from_shapely(arr, **kwargs)
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -2,9 +2,10 @@ import numpy as np
 
 from . import Geometry  # noqa
 from . import lib
+from . import geos_version
 
 
-__all__ = ["from_wkb", "from_wkt", "to_wkb", "to_wkt"]
+__all__ = ["from_shapely", "from_wkb", "from_wkt", "to_wkb", "to_wkt"]
 
 
 def to_wkt(
@@ -13,7 +14,7 @@ def to_wkt(
     trim=True,
     output_dimension=3,
     old_3d=False,
-    **kwargs
+    **kwargs,
 ):
     """
     Converts to the Well-Known Text (WKT) representation of a Geometry.
@@ -80,7 +81,9 @@ def to_wkt(
     )
 
 
-def to_wkb(geometry, hex=False, output_dimension=3, byte_order=-1, include_srid=False, **kwargs):
+def to_wkb(
+    geometry, hex=False, output_dimension=3, byte_order=-1, include_srid=False, **kwargs
+):
     r"""
     Converts to the Well-Known Binary (WKB) representation of a Geometry.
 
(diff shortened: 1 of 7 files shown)


pygeos/io.py
```diff
@@ -13,7 +14,7 @@ def to_wkt(
     trim=True,
     output_dimension=3,
     old_3d=False,
-    **kwargs
+    **kwargs,
```
jorisvandenbossche: This might give syntax errors on py 3.5? (another reason to drop it ;-))

caspervdw: I just ran the black code formatter, which makes some guesses about what you want. This will break 3.5 indeed.

pygeos/io.py
```diff
@@ -172,3 +175,38 @@ def from_wkb(geometry, **kwargs):
...
+    ----------
+    geometry : shapely Geometry object or array_like
+
+    Examples
+    --------
+    >>> from shapely.geometry import Point   # doctest: +SKIP
+    >>> from_wkt(Point(1, 2))   # doctest: +SKIP
+    <pygeos.Geometry POINT (1 2)>
+    """
+    from shapely.geos import geos_version_string as shapely_geos_version
```
jorisvandenbossche: Shapely also has the `shapely._buildcfg.geos_library_path` which might make it possible to check that both are using the exact same library. But the same version should be enough? (I am no expert in this, I don't know if it possible that if the same version is built with a different C toolchain, that this can still give binary incompatibilities)

pygeos/io.py
```diff
@@ -172,3 +175,38 @@ def from_wkb(geometry, **kwargs):
...
+    if not shapely_geos_version.startswith(geos_version):
+        raise ImportError(
+            "The shapely GEOS version ({}) is "
+            "incompatible".format(shapely_geos_version)
+        )
+
+    if isinstance(geometry, BaseGeometry):
+        # this so that the __array_interface__ of the shapely geometry is not
+        # used, converting the Geometry to its coordinates
+        geometry = np.array([geometry], dtype=np.object).reshape(())
```
jorisvandenbossche: To avoid issue with multi-geometries, you need something like

```
arr = np.empty(1, dtype=object)
arr[:] = [geometry]
```

pygeos/io.py
jorisvandenbossche: You have a command line argument to ensure the output is py3.5 compatible: https://github.com/psf/black#command-line-options

pygeos/io.py
caspervdw: Maybe, I am also no expert on this. Accessing private shapely attributes does have its own problems.

caspervdw: One question to approach this issue: how to get the actual library path (at runtime) from our C extension?

jorisvandenbossche: From shapely, it seems this:

```
In [15]: import shapely.geos                                                                                                                                                                                       

In [16]: shapely.geos._lgeos                                                                                                                                                                                       
Out[16]: <CDLL '/home/joris/miniconda3/envs/test-pygeos-shapely-wheels/lib/python3.7/site-packages/shapely/.libs/libgeos_c-bd8d3f16.so.1.10.2', handle 55af2a25e300 at 0x7f4eaa8e8dd8>
```

shows where it is coming from. Above, this is from a wheel that includes geos. 
And below is from a conda env:

```
In [25]: shapely.geos._lgeos                                                                                                                                                                                       
Out[25]: <CDLL '/home/joris/miniconda3/envs/dev/lib/libgeos_c.so.1', handle 5603028c7f00 at 0x7f714ea6cf28>
```

(but not sure how to get this path out of the repr)

But that's only for shapely of course, we don't have that for pygos

jorisvandenbossche:
I did a quick experiment today: built a pygeos wheel in a conda environment (so with geos compiled from conda-forge), and then install that in a new environment, and also install a shapely wheel in that. So both have their own copy of GEOS. And for a few geometries that I tested, this worked OK. 

I am not sure that is a comprehensive test, though (also no idea how it works on other platforms). Would still be good to have feedback from somebody who knows more about it.

caspervdw:
@jorisvandenbossche  I think I'd like to merge this PR. Maybe there are some combinations of binaries that will lead to issues, but I don't see how we could protect from that better currently.

jorisvandenbossche:
Yes, I am fine with merging this. We can revisit later if we get more experience with it.

PR #69, opened 2019-10-28
[ENH] Add bounds function

caspervdw:
I did not use the MinX etc. functions from the GEOS C-API, because:
 - each of these 4 functions construct the envelope
 - these functions are not available in GEOS 3.5

Instead, I construct the envelope and extract the coordinates from it.

### DIFF
--- a/pygeos/measurement.py
+++ b/pygeos/measurement.py
@@ -1,7 +1,7 @@
 from . import lib
 from . import Geometry  # NOQA
 
-__all__ = ["area", "distance", "extent", "length", "hausdorff_distance"]
+__all__ = ["area", "distance", "bounds", "length", "hausdorff_distance"]
 
 
 def area(geometry, **kwargs):
@@ -49,8 +49,8 @@ def distance(a, b, **kwargs):
     return lib.distance(a, b, **kwargs)
 
 
-def extent(geometry, **kwargs):
-    """Computes the extent (bounds) of a geometry.
+def bounds(geometry, **kwargs):
+    """Computes the bounds (extent) of a geometry.
 
     For each geometry these 4 numbers are returned: min x, min y, max x, max y.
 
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
One other note is about naming: although this is using the same name as Postgis has ("extent"), note that this is not the same function: in postgis it is 1) returning a box object (not something we want to mimic I think) and 2) a reducing function (so rather the "total extent") mentioned above.

jorisvandenbossche:
Opinion on the naming? (see last comment) In shapely it is "bounds".

caspervdw:
I would also like to go for “bounds“. We do not want confusion with the ST_Extent from postgis.

Also a “total_bounds” would be nice, probably just an addition on the python level. Something for another PR.

jorisvandenbossche:
Yes, agree the total bounds can be done separately.

Just pushed a extent -> bounds rename

jorisvandenbossche:
The failure is due to older numpy versions. I can reproduce locally with numpy 1.15. It already segfaults on import (I suppose on registering the ufunc).

It seems older numpy has a problem with the `()->(4)` signature. 
If I replace it with `()->()`, it actually "works" (it doesn't segfault, and the function runs, but of course returns the wrong result)

jorisvandenbossche:
OK, thanks to the response I got on the numpy issue I opened (https://github.com/numpy/numpy/issues/14949), I added a workaround for the older numpy versions: we only then need to construct and specify the `out` argument, but that's no problem (we can easily do that in the python wrapper function).

jorisvandenbossche:
All green now! ;)

jorisvandenbossche:
@caspervdw OK to merge for you?

jorisvandenbossche:
I am going to merge this, it's handy to have for my geopandas integration testing. And with the extra patch, it's working nicely on older numpy versions as well.

PR #70, opened 2019-10-28
[WIP] Empty points in WKB

### LINKED ISSUE #60: Fix the ToWKB for empty points
The WKB spec does not allow empty points. Some reads:

https://github.com/Toblerity/Shapely/issues/742
https://trac.osgeo.org/postgis/ticket/3181

If I'm correct (needs checking), we should serialize it to POINT (NaN, NaN), and deserialize back to POINT EMPTY.

caspervdw:
Closes #60

### DIFF
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -5,9 +5,9 @@ import pytest
 from .common import all_types, point
 
 
-POINT11_WKB = (
-    b"\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?"
-)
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Question: if we add this special case handling in `to_wkb`, should we do something similar for `from_wkb` ?  
If we want "full" roundtrip to/from WKB, it should convert it back to an empty point. But, of course, that gets more ambiguous, as a `Point(nan, nan)` can actually also exist...

caspervdw:
The discussion on PostGIS trac seemed to have touched this point also. I think they decided to support the roundtrip, so import Point(nan, nan) as Point empty.

caspervdw:
Closing as this work is outdated

PR #72, opened 2019-11-04
Add dynamic Geometry __getattr__ and __dir__

caspervdw:
This adds all the pygeos functions as Geometry methods:

```
pygeos.Geometry("POINT (1 1)").is_empty()
>>> False
```

By extending the Geometry __dir__ and __getattr__ methods with a dynamic python function.

Code completion and documentation work also

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -2,6 +2,7 @@ from enum import IntEnum
 import numpy as np
 from . import Geometry  # NOQA
 from . import lib
+from .geometry import register_geometry_method
 
 
 __all__ = [
@@ -33,6 +34,7 @@ class BufferJoinStyles(IntEnum):
     BEVEL = 3
 
 
+@register_geometry_method
 def boundary(geometry, **kwargs):
     """Returns the topological boundary of a geometry.
 
@@ -57,6 +59,7 @@ def boundary(geometry, **kwargs):
     return lib.boundary(geometry, **kwargs)
 
 
+@register_geometry_method
 def buffer(
     geometry,
     radius,
@@ -159,6 +162,7 @@ def buffer(
     )
 
 
+@register_geometry_method
 def centroid(geometry, **kwargs):
     """Computes the geometric center (center-of-mass) of a geometry.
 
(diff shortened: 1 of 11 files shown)


caspervdw:
@jorisvandenbossche What do you think of this approach for extending our C extension type with pure Python code?

jorisvandenbossche:
@caspervdw sorry for the slow reply 

This is certainly a cool and actually easy approach to extend the methods of the Geometry class. 

Another idea that has been floating around in my head, given shapely's current subclasses for the different geometry types (Polygon, Linestring, Point, etc): it is easy to make the Geometry class subclassable. But would it be possible, if we define such subclasses in python (and then we can add methods there), to have a function in C that, given a GeometryObject, creates one of those python subclasses (depending on the geometry type id). Then we could call that (maybe optionally) in those ufuncs where new objects are created.

caspervdw:
@jorisvandenbossche We could also wrap the Geometry objects in a shapely object through this trick.

jorisvandenbossche:
How do you mean exactly?

caspervdw:
We could make the Geometry objects appear as shapely geometry object by pointing the `__getattr__` from the C code to a function that does

```
def shapely_getattr(obj, name):
    obj_shapely = as_shapely(obj)  # construct shapely geometry
    return getattr(obj_shapely, name)
```

jorisvandenbossche:
But your small code snippet still assumes there are separate Shapely geometry objects? IMO the goal would be to replace those with new objects based on Geometry, while trying to provide a similar API.

caspervdw:
@jorisvandenbossche  I think providing a similar API is the way to go indeed.

I reworked this a bit, every function now can be decorated with `register_geometry_method`, so that it it accessible from the Geometry object. There is also a geometry type specific registry so that for instance only linestrings expose the methods that are meant for line strings.
```
@register_geometry_method
def is_empty(geometry):
    ...

@register_geometry_method([GeometryType.LINESTRING])
def point_on_line(line):
    ...
```

caspervdw:
Closing, it is superseded by #104

Issue #73, opened 2019-11-04
include shared_paths function


mattijn:
Firstly, congrats on this package! It looks much promising!

I noticed that the `pygeos.shared_paths()` function was written out of the package in https://github.com/pygeos/pygeos/pull/53. It would be great to have it back!

I tried to include the removed code again and got it somehow working (see code [change](https://github.com/pygeos/pygeos/compare/master...mattijn:MH/add-shared-paths)), see below for some tests. Not sure how to continue or which output/dimension is looked after to make the function reappear. 

The tests I did:

```python
from shapely import geometry
import pygeos
import numpy as np
from IPython.display import SVG, display
```

```python
def svg_style(lines):
    svg_lines = lines._repr_svg_()
    svg_lines = svg_lines.replace('stroke="#66cc99"', 'stroke="orange"', 1)
    svg_lines = svg_lines.replace('stroke-width="0.0432"', 'stroke-width="0.1"', 1)
    svg_lines = svg_lines.replace('opacity="0.8"', 'opacity="0.4"', 1)

    display(SVG(svg_lines))    
```

```python
# prepare some linestrings
g1 = geometry.LineString([(0, 0), (1, 0), (1, 1), (0, 1), (0, 0)])
g2 = geometry.LineString([(1, 0), (2, 0), (2, 1), (1, 1), (1, 0)])
g3 = geometry.LineString([(0, 0), (2, 0), (2, 1), (0, 1), (0, 0)])
```

```python
# shared paths in opposite direction of both inputs
lines_opposite_direction = geometry.MultiLineString([g1, g2])

# orange is g1
# green is g2
svg_style(lines_opposite_direction)
```

<img width="500" alt="image" src="https://user-images.githubusercontent.com/5186265/68161460-37cd9100-ff56-11e9-8e5a-b3e6c829274c.png">

```python
# shared paths in same direction of both inputs
lines_same_direction = geometry.MultiLineString([g1,g3])

# orange is g1
# green is g3
svg_style(lines_same_direction)
```

<img width="500" alt="image" src="https://user-images.githubusercontent.com/5186265/68161511-50d64200-ff56-11e9-8d26-6bc7e072d364.png">

```python
# convert to pygeos
pyg1 = pygeos.from_wkt(g1.wkt)
pyg2 = pygeos.from_wkt(g2.wkt)
pyg3 = pygeos.from_wkt(g3.wkt)
```

```python
# only contain shared paths in backward direction
[...]

PR #77, opened 2019-11-11
Reintroduce shared_paths

mattijn:
This PR reintroduce the function `pygeos.shared_paths()` as was discussed in https://github.com/pygeos/pygeos/issues/73.

The `shared_paths()` function returns a `pygeos.Geometry GEOMETRYCOLLECTION`. 

To access the shared paths that have the same direction in both inputs use index 0 in `pygeos.get_geometry()` using the result of `shared_paths()` and use index 1 to access the shared paths that have the opposite direction.

### DIFF
--- a/pygeos/test/test_linear.py
+++ b/pygeos/test/test_linear.py
@@ -70,5 +70,4 @@ def test_shared_paths_linestring():
     g2 = pygeos.linestrings([(0, 0), (1, 0)])
     actual1 = pygeos.shared_paths(g1, g2)
     assert pygeos.equals(pygeos.get_geometry(actual1, 0), g2)
-    assert pygeos.shared_paths(None, None) == None
-    assert len(pygeos.shared_paths([[g1, g2], [g1, g2]], axis=1)) == 2
+    assert pygeos.shared_paths(None, None) is None
--- a/pygeos/test/test_linear.py
+++ b/pygeos/test/test_linear.py
@@ -65,6 +65,7 @@ def test_line_merge_geom_array():
     assert pygeos.equals(actual[0], line_string)
     assert pygeos.equals(actual[1], multi_line_string)
 
+
 def test_shared_paths_linestring():
     g1 = pygeos.linestrings([(0, 0), (1, 0), (1, 1)])
     g2 = pygeos.linestrings([(0, 0), (1, 0)])
--- a/pygeos/test/test_linear.py
+++ b/pygeos/test/test_linear.py
@@ -64,3 +64,9 @@ def test_line_merge_geom_array():
     actual = pygeos.line_merge([line_string, multi_line_string])
     assert pygeos.equals(actual[0], line_string)
     assert pygeos.equals(actual[1], multi_line_string)
+
+def test_shared_paths_linestring():
+    g1 = pygeos.linestrings([(0, 0), (1, 0), (1, 1)])
+    g2 = pygeos.linestrings([(0, 0), (1, 0)])
+    actual = pygeos.shared_paths(g1, g2)
+    assert pygeos.equals(pygeos.get_geometry(actual, 0), g2)
--- a/pygeos/test/test_linear.py
+++ b/pygeos/test/test_linear.py
@@ -68,5 +68,7 @@ def test_line_merge_geom_array():
 def test_shared_paths_linestring():
     g1 = pygeos.linestrings([(0, 0), (1, 0), (1, 1)])
     g2 = pygeos.linestrings([(0, 0), (1, 0)])
-    actual = pygeos.shared_paths(g1, g2)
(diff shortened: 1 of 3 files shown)


caspervdw:
Thanks @mattijn for your contribution. I don't remember why I removed this function. It might have had some edge cases that lead to segfaults and I didn't have time to cover them.

Before this can be merged, this needs some work on unittests and documentation. Do you have time to have a look?

mattijn:
I added a test and docstring including an example.

jorisvandenbossche:
Thanks for the update!

Some other things:

- Could you add a test with None as input?
- The `axis` is not tested right now. But wondering: is this needed? Or how is it actually working? (the shared_path is working with 2 geoms, so in a reduction, is it doing that each time with the next geom and the result of the previous? Does that work with the 2-item GeometryCollection ?

mattijn:
Added a test using `None`s and one using the `axis` parameter.

I was also surprised on the `axis` part, but it works with a `np.vstack` array-like input:
```python
arr1 = [g1, g2]
arr2 = [g1, g2]
geom = np.vstack((arr1, arr2)) # [[g1, g2], [g1, g2]]
pygeos.shared_paths(geom, axis=1)
```
```
array([<pygeos.Geometry GEOMETRYCOLLECTION (MULTILINESTRING ((0 0, 1 0)), MULTILINESTRING EMPTY)>,
       <pygeos.Geometry GEOMETRYCOLLECTION (MULTILINESTRING ((0 0, 1 0)), MULTILINESTRING EMPTY)>],
      dtype=object)
```
So no fancy stuff

jorisvandenbossche:
Regarding None, it seems that if only the right operand is None, it doesn't give the expected behaviour (at least to me):

```
In [6]: g1 = pygeos.linestrings([(0, 0), (1, 0), (1, 1)]) 
   ...: g2 = pygeos.linestrings([(0, 0), (1, 0)])   

In [7]: pygeos.shared_paths(None, None) 

In [8]: pygeos.shared_paths(g1, None) 
Out[8]: <pygeos.Geometry LINESTRING (0 0, 1 0, 1 1)>

In [9]: pygeos.shared_paths(None, g1)
```

pygeos/test/test_linear.py
```diff
@@ -64,3 +64,11 @@ def test_line_merge_geom_array():
     actual = pygeos.line_merge([line_string, multi_line_string])
     assert pygeos.equals(actual[0], line_string)
     assert pygeos.equals(actual[1], multi_line_string)
+
+def test_shared_paths_linestring():
+    g1 = pygeos.linestrings([(0, 0), (1, 0), (1, 1)])
+    g2 = pygeos.linestrings([(0, 0), (1, 0)])
+    actual1 = pygeos.shared_paths(g1, g2)
+    assert pygeos.equals(pygeos.get_geometry(actual1, 0), g2)
+    assert pygeos.shared_paths(None, None) == None
```
jorisvandenbossche:
```suggestion
    assert pygeos.shared_paths(None, None) is None
```

(in practice doesn't matter much, but for checking to None, using `is` is seen as better practice)

jorisvandenbossche:
And for the axis behaviour: your example works as long as the axis you specify has only two elements. But once it has more, you get an error:

```
In [12]: pygeos.shared_paths([[g1, g2, g1], [g1, g2, g1]], axis=1) 
<ipython-input-12-b4f3ff6754a6> in <module>
----> 1 pygeos.shared_paths([[g1, g2, g1], [g1, g2, g1]], axis=1)

~/scipy/repos/pygeos/pygeos/linear.py in shared_paths(geom1, geom2, axis)
    113     """    
    114     if geom2 is None:
--> 115         return lib.shared_paths.reduce(geom1, axis=axis)
    116     else:
    117         return lib.shared_paths(geom1, geom2)

GEOSException: IllegalArgumentException: Geometry is not lineal
```

This is because it first calculates the shared path of the first two items, resulting in a GeometryCollection, and then it calculates the shared path of this collection with the third item. But you can't pass GeometryCollection to this algo, it needs linear geometries.

So therefore (unless you have a good use case for it?), I would maybe suggest to leave out the `axis` behaviour for now. We can always easily add it back later if there is demand for it.

mattijn:
Removed the `axis` parameter together with the `reduce` functionality.

jorisvandenbossche:
@mattijn thanks for the update! 

I pushed some small stylistic edits (hope you are fine with that; that's a bit easier than giving a bunch of small nitpicky comments ;))

Looks good to me now. @caspervdw ?

caspervdw:
Looks good to me too! Two questions:
- could you propagate `**kwargs` in the function?
- could you add a test that feeds a non-linestring to the function? It should raise.

caspervdw:
@mattijn merging this, thanks again

Issue #31, opened 2019-08-25
Don't use a global geos context handle


jorisvandenbossche:
Currently a global handle is defined:

https://github.com/caspervdw/pygeos/blob/de5ed5c5f0a55bf471531e328d1a1b0476bcafd2/src/ufuncs.c#L1206-L1211

However, if you want to use pygeos in multiple threads (when releasing the GIL, https://github.com/caspervdw/pygeos/issues/7), I *think* we need a context per thread.

In the cython branch of geopandas, we had a context manager to create and destroy a thread:

https://github.com/geopandas/geopandas/blob/e92536340fd7f1ed92925b960226ab10adb37717/geopandas/vectorized.pyx#L1270-L1278

which was then used in each vectorized function (here the ufuncs) around the code that called into GEOS.

I suppose we will need something similar here, although not sure what the best way would be (since the ufuncs are templated, just creating in the beginning / destroying at the end might be not too much duplication). 

The version in pygeos also sets the error message handler, which we didn't do in the cython version. Is that for converting geos errors into python errors? (and does that work how it is currently coded? Shapely wraps some logging functionality with it)

caspervdw:
I actually had this in the ufuncs before. It's not so much extra boilerplate, but I was worried about the additional overhead it would create. Are you sure you need a context per thread? Or is the context already a thread-safe way to storing global variables?

> The version in pygeos also sets the error message handler, which we didn't do in the cython version. Is that for converting geos errors into python errors? (and does that work how it is currently coded? Shapely wraps some logging functionality with it)

Yes I pack the GEOS error message into a python GEOSException. For example, if you try to get the `pygeos.boundary` of a geometrycollection. That should probably also be changed when we want to release the GIL..

jorisvandenbossche:
Hmm, I don't directly find any discussion about this (and also not directly any GEOS docs on this), but for some reason I thought this was the motivation for changing that in the geopandas-cython branch (we initially had a global context as well).

jorisvandenbossche:
I found an example in the GEOS source (https://github.com/libgeos/geos/blob/master/tests/thread/threadtest.c), which initializes a handle in each thread. And from https://trac.osgeo.org/geos/wiki/RFC3 I seem to understand the same (I understand it as the handle being a way to store global variables in a thread-safe way, as through the per-context handle, each thread has its copy of the global variables; "allowing each thread to have it's own copy of the data" from the document).


We should probably do some timings to check the overhead of creating a context on each function call.

Issue #7, opened 2019-07-08
Release the GIL where possible


caspervdw:
To take advantage of multithreading possibilities, we should release the GIL whenever we do operations that do not involve PyObjects (most of them)

jorisvandenbossche:
Currently, the ufunc inner loop functions still have python interaction. Eg the ones that return a geometry:

https://github.com/caspervdw/pygeos/blob/e885283d23b26775512cd4b8810786bcc4ac15f4/src/ufuncs.c#L43-L55

But also in the input handling

https://github.com/caspervdw/pygeos/blob/e885283d23b26775512cd4b8810786bcc4ac15f4/src/ufuncs.c#L32-L34

the `get_geom` function deals with Python objects.

So that seems a problem for releasing the GIL here with the current approach?

jorisvandenbossche:
In the `get_geom` function

https://github.com/caspervdw/pygeos/blob/e885283d23b26775512cd4b8810786bcc4ac15f4/src/ufuncs.c#L397-L418

it seems this is mainly for the cases where the input is not a GeometryObject struct but None or np.nan ? 
So here, a solution might be to have a missing value indicator that is also a GeometryObject ? (eg with a null pointer)

caspervdw:
The functions that create new geometries definitely cannot release the GIL. I could collect the C pointers in an initial loop and then replace them with Python objects in a second, which would leave most of the time to release the GIL.

Concerning the inputs, I have to read up on this, but my it is my impression that only reading from a static attribute from an object will be possible without the GIL. Do you know if this is possible? Otherwise the same approach as above could be taken: first collect the pointers, then execute the GEOS functions without GIL

jorisvandenbossche:
I sent an email to the numpy mailing list to ask about this, whether it is possible to let numpy release the GIL in the case where we only need access to the pointer attribute (I think by default it does not release the GIL in ufuncs if the arrays are of object dtype).

For collecting the pointers first: that would mean that all ufuncs are written for arrays of only pointers, and then the level above that (the python wrappers?) ensures to first create this array of pointers and calls the ufunc on those? (and potentially wrap again in the end)

I was doing a quick profiling of both steps separately with the contains in cython. And for this case, collecting the pointers takes around 10% of the time of the actual contains operation. So that might be acceptable

caspervdw:
> I sent an email to the numpy mailing list to ask about this,

Thanks!!

[...]

PR #80, opened 2019-11-22
Release the GIL within the ufuncs where possible

jorisvandenbossche:
xref https://github.com/pygeos/pygeos/issues/7

@caspervdw I cleaned up my previous experiment with releasing the GIL (and added handling of None), and pushed it here. This is only doing it for the "YY_d" (eg distance) for testing the approach.

Tested with:

```python
import numpy as np
import pygeos

N = 1_000_000
arr = pygeos.points(np.random.rand(N, 2))
arr[np.random.randint(0, N, size=100)] = None
poly = pygeos.box(0, 0, 0.2, 0.2)

res1 = pygeos.distance(poly, arr)
```
and it doesn't give a noticeable slowdown in a single thread. Testing it with dask with multi-threading, then it gives a speed-up of around 3x on my 4-core laptop:

```python
import dask.array
arr_dask = dask.array.from_array(arr, chunks=100_000)

res2 = arr_dask.map_blocks(pygeos.distance, poly, dtype=float)
assert np.allclose(res1, res2.compute(), equal_nan=True)
%timeit res2.compute()
```

Now, some things / questions about the approach:

- I do a first pass to check the data, and then in a second nogil pass I get the pointers out without checking. 
  A problem might be that in the meantime, those objects could have changes, if the array gets modified while the ufunc is running.
- An alternative is probably what I did in https://github.com/pygeos/pygeos/issues/7#issuecomment-522708942, which is copying the pointers into a new array in a first pass and looping over that in a second nogil pass. However, with the other approach I wanted to avoid the copying of the pointers. Also, in this approach, we will probably need to increase the reference count on all objects to avoid they get released while the nogil loop is running in case someone modifies the array.
- I still need to check how this works with the geos context and geos exceptions being raised (does each function needs its own context? xref https://github.com/pygeos/pygeos/issues/31). To test this, I need to find a code example where GEOS actually raises an error. In principle the GEOSDistance function "Return 0 on exception", but not sure how to trigger this with an example (we also don't have tests for this, only in the IO tests, but for those it is more difficult to release the gil).

Since for both approaches, the fact that the array might get modified while the ufunc is running is problematic, I am wondering if we can do something about that? Like making the array non writeable when passing to the ufunc (but, how does that work with slices and views? How does that work if the same array is passed to multiple ufuncs? ..)

### DIFF
--- a/src/pygeom.c
+++ b/src/pygeom.c
@@ -150,7 +150,6 @@ PyTypeObject GeometryType = {
     .tp_str = (reprfunc) GeometryObject_str,
 };
 
-
 /* Get a GEOSGeometry pointer from a GeometryObject, or NULL if the input is
 Py_None. Returns 0 on error, 1 on success. */
 char get_geom(GeometryObject *obj, GEOSGeometry **out) {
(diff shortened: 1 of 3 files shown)


src/pygeom.c
```diff
@@ -168,6 +167,32 @@ char get_geom(GeometryObject *obj, GEOSGeometry **out) {
     }
 }
 
+/* Check if the input is a GeometryObject or Py_None. Returns 0 on error, 1 on success. */
+char check_geom(GeometryObject *obj) {
+    if (!PyObject_IsInstance((PyObject *) obj, (PyObject *) &GeometryType)) {
```
caspervdw: If we would simplify this check to an exact check

```
if (Py_TYPE(obj) != (PyTypeObject *) GeometryType)
```

We might be able to do this with the GIL released?

src/ufuncs.c
```diff
@@ -543,11 +543,32 @@ static void YY_d_func(char **args, npy_intp *dimensions,
...
+    NPY_BEGIN_THREADS_DEF;
+    NPY_BEGIN_THREADS;
+
+    for(i = 0; i < n; i++, ip1 += is1, ip2 += is2, op1 += os1) {
+        /* get the geometries in a second pass in "unsafe" way -> we already
+        checked that the input is either a GeometryObject or either None */
+        get_geom_nogil(*(GeometryObject **)ip1, &in1);
+        get_geom_nogil(*(GeometryObject **)ip2, &in2);
+
+        if ((in1 == 0) | (in2 == 0)) {
```
caspervdw: I would prefer `== NULL` here

PR #79, opened 2019-11-12
Add extern in header files

caspervdw:
@thomersch this might solve your issue #78

PR #83, opened 2019-12-14
Remove header include that makes clang complain

sgillies:
With this include in place, the build fails and clang reports "duplicate symbol" _STRTreeType. The project builds when I remove the header include and all the tests pass on my macbook.

Here are my gcc version details:

```
$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk/usr/include/c++/4.2.1
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

### DIFF
--- a/src/lib.c
+++ b/src/lib.c
@@ -13,7 +13,6 @@
 #include "pygeom.h"
 #include "coords.h"
 #include "ufuncs.h"
-#include "strtree.h"
 
 
 /* This tells Python what methods this module has. */


caspervdw:
Thanks @sgillies . I really don’t understanding why this is working. If we don’t include “strtree.h”, how is “init_strtree_type” defined?

I am clearly missing something as CI passes... could you help me out?

jorisvandenbossche:
A similar problem on Mac was previously solved with adding "extern" -> https://github.com/pygeos/pygeos/pull/79

But indeed, I also don't understand why it works without including strtree. 

We should probably try to add Mac testing, which should be possible on Travis (or on Azure / GitHub actions)

sgillies:
It works because C compilers use a ton of magic 😄 

Looking more closely, the compiler does warn about the implicit declaration.

```
    src/lib.c:55:9: warning: implicit declaration of function 'init_strtree_type' is invalid in C99 [-Wimplicit-function-declaration]
        if (init_strtree_type(m) < 0) {
            ^
    src/lib.c:55:9: warning: this function declaration is not a prototype [-Wstrict-prototypes]
```

There are also a lot of other warnings.

Perhaps a single .c file would help. There is only one extension module, lib.so, so we really only require a single source file.

vincentsarago:
I'm stuck here 😢 too

```
$ clang --version
Apple clang version 11.0.0 (clang-1100.0.33.17)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

$ pip install pygeos
... 
... 
... 
duplicate symbol '_STRtreeType' in:
    build/temp.macosx-10.14-x86_64-3.7/src/strtree.o
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: command 'clang' failed with exit status 1
```

So @sgillies if you closed your PR how did you get further ?

sgillies:
@vincentsarago I stopped there and haven't gone further. I really do believe the way to go is one source file per extension module. If you want to try the project out as it is, removing the include statement will unblock you.

jorisvandenbossche:
@vincentsarago could you try if an approach like in https://github.com/pygeos/pygeos/pull/79 would help?

This is clearly not a solved problem, so I opened https://github.com/pygeos/pygeos/issues/89 to track this / continue the discussion.

PR #86, opened 2019-12-20
BUG: Check types in create_collections

### LINKED ISSUE #74: BUG: passing geometries of wrong type to constructive functions gives segfault
At least noticed this with `pygeos.multipolygons` when passing points or polygons:

```
In [2]: a = pygeos.points(np.arange(10), np.arange(10)) 

In [3]: a
Out[3]: 
array([<pygeos.Geometry POINT (0 0)>, <pygeos.Geometry POINT (1 1)>,
       <pygeos.Geometry POINT (2 2)>, <pygeos.Geometry POINT (3 3)>,
       <pygeos.Geometry POINT (4 4)>, <pygeos.Geometry POINT (5 5)>,
       <pygeos.Geometry POINT (6 6)>, <pygeos.Geometry POINT (7 7)>,
       <pygeos.Geometry POINT (8 8)>, <pygeos.Geometry POINT (9 9)>],
      dtype=object)

In [4]: pygeos.points(a) 
...
TypeError: ufunc 'points' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

In [5]: pygeos.multipolygons(a) 
Out[5]: Segmentation fault (core dumped)
```

Passing those to `pygeos.multipoints` correctly creates a MultiPoint geometry.

caspervdw:
Closes #74 . I cannot reproduce locally, but maybe a CI picks it up

### DIFF
--- a/pygeos/test/test_creation.py
+++ b/pygeos/test/test_creation.py
@@ -131,9 +131,14 @@ def test_create_collection_skips_none():
     assert str(actual) == "MULTIPOINT (2 3, 2 3)"
 
 
-def test_create_collection_wrong_type():
-    a = pygeos.points(np.arange(10), np.arange(10))
-    pygeos.multipolygons(a)
+def test_create_collection_wrong_collection_type():
+    with pytest.raises(TypeError):
+        pygeos.lib.create_collection([point], 2)
+
+
+def test_create_collection_wrong_geom_type():
+    with pytest.raises(TypeError):
+        pygeos.multipolygons([point])
 
 
 def test_box():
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
So as a test it also passes for me. Which led me to find that it actually returns an object, but it then segfaults on the repr.

caspervdw:
@jorisvandenbossche  I could reproduce the issue, fixed it by checking the input types when creacting a geometrycollection.

jorisvandenbossche:
Thanks!

Issue #52, opened 2019-10-06
Port the spatial join algorithm from geopandas


jorisvandenbossche:
In the cython branch in GeoPandas, we had a C implementation of a spatial join algorithm (using the STRTree from GEOS). This was added in https://github.com/geopandas/geopandas/pull/475

Since it's written in C directly against the GEOS C API, it would be nice to include it in this package (primarily for use in geopandas, but it might also be useful in general).

One question is how it would fit in the current API of pygeos. 
The version in GeoPandas is currently written as a C function that gets 2 C arrays of geometry objects:

```
size_vector sjoin(GEOSContextHandle_t handle,
                  GEOSPreparedPredicate predicate,
                  GEOSGeometry **left, size_t nleft,
                  GEOSGeometry **right, size_t nright);
```

which is then wrapped in cython to provide a python interface with a signature of

```
def sjoin(left: np.ndarray[1D], right: np.ndarray[1D], predicate_name: str) -> (left_out, right_out)
```

where `left_out` and `right_out` are arrays of indices into left and out, respectively, for all the matches according to the predicate.

I think it is difficult to generalize this to multiple dimensions, as currently all other functions in pygeos do.

jorisvandenbossche:
@brendan-ward what were your ideas regarding spatial joins? Since you mentioned "This should make it easier to implement spatial joins on top of pygeos." in https://github.com/pygeos/pygeos/pull/87

I think we should still have a spatial join algorithm in pygeos itself (as having the outer loop in c as well will still be beneficial). 

It could reuse the `STRtree_query` function, and call that multiple times for all geometries However, the problem is that that is a "python function", so not ideal to reuse from C. So we might need to factor out the actual "query" part that doesn't need python interaction (eg keyword parsing).

I think the API we have had in the GeoPandas cython branch (see links above) returning two arrays of indices (indices into the first and second set of geometries, respectively) is probably the most general one.

brendan-ward:
@jorisvandenbossche agreed on all points.

The general thought was just what you said, we need a vectorized way to loop over queries to the tree.  Spatial join is basically a bulk query, with a predicate applied after, and possibly some smarts about the internal direction of the join to best leverage prepared geometries.

A couple ways we could achieve this (without thinking it through all the way):
1. create a function in C that implements the core of the query operation, and use this inside C functions exposed to Python for both singular and bulk queries to the tree (the latter within a loop in C).
[...]

PR #87, opened 2020-01-13
ENH: include predicates in STRtree.query()

brendan-ward:
This PR adds support for several geometric predicates into the STRtree `query()` function.  In addition to querying geometries from the tree that have envelopes that intersect the envelope of the input geometry, this now allows you to test the predicate relationships between the input geometry and the returned geometries from the tree:
* `intersects`
* `within`
* `contains`
* `overlaps`
* `crosses`
* `touches`

This intentionally does not include other binary predicates, since they don't seem useful in this context (e.g., `disjoint`).

This should make it easier to implement spatial joins on top of `pygeos`.

This could be further improved by using prepared geometries and vectorizing the query function, but those should be handled in a separate PR.

Most of the code here is in tests, to verify that the predicates are being evaluated correctly.  These could be improved further by adding more combinations of geometry types.

In order to preserve the existing behavior where querying the tree with a geometry of `None` returns an empty array, I specifically handle that in the wrapper function since there is no need to step into C code for that.

### DIFF
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -1,7 +1,23 @@
-from pygeos import lib
+from enum import IntEnum
 import numpy as np
+from pygeos import lib
+
 
 __all__ = ["STRtree"]
+
+
+class UnaryPredicate(IntEnum):
+    """The enumeration of GEOS unary predicates types"""
+
+    intersects = 1
+    within = 2
+    contains = 3
+    overlaps = 4
+    crosses = 5
+    touches = 6
+
+
+VALID_PREDICATES = {e.name for e in UnaryPredicate}
 
 
 class STRtree:
@@ -32,15 +48,36 @@ class STRtree:
     def __len__(self):
         return self._tree.count
 
-    def query(self, envelope):
-        """Return all items whose extent intersect the given envelope.
+    def query(self, geometry, predicate=None):
+        """Return all items whose extent intersect the envelope of the input
+        geometry.  If predicate is provided, these items are limited to those
+        that satisfy the predicate operation when compared against the input
+        geometry.
 
         Parameters
         ----------
-        envelope : Geometry
-            The envelope of the geometry is taken automatically.
+        geometry : Geometry
+            The envelope of the geometry is taken automatically for
+            querying the tree.
+        predicate : str, optional (default: None)
+            The predicate to use for testing geometries from the tree
+            that are within the input geometry's envelope.
         """
-        return self._tree.query(envelope)
+
+        if predicate is None:
+            predicate = 0
+
+        else:
+            if not predicate in VALID_PREDICATES:
+                raise ValueError(
+                    "Predicate {} is not valid; must be one of {}".format(
+                        predicate, ", ".join(VALID_PREDICATES)
+                    )
+                )
+
+            predicate = UnaryPredicate[predicate].value
+
+        return self._tree.query(geometry, predicate)
 
     @property
     def geometries(self):
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -54,6 +54,8 @@ class STRtree:
         that satisfy the predicate operation when compared against the input
         geometry.
 
+        If geometry is None, an empty array is returned.
+
         Parameters
         ----------
         geometry : Geometry
@@ -63,6 +65,9 @@ class STRtree:
             The predicate to use for testing geometries from the tree
             that are within the input geometry's envelope.
         """
+
+        if geometry is None:
+            return np.array([], dtype="int")
 
         if predicate is None:
             predicate = 0
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -6,8 +6,8 @@ from pygeos import lib
 __all__ = ["STRtree"]
 
 
-class UnaryPredicate(IntEnum):
-    """The enumeration of GEOS unary predicates types"""
+class BinaryPredicate(IntEnum):
+    """The enumeration of GEOS binary predicates types"""
 
     intersects = 1
     within = 2
@@ -17,7 +17,7 @@ class UnaryPredicate(IntEnum):
     touches = 6
 
 
-VALID_PREDICATES = {e.name for e in UnaryPredicate}
+VALID_PREDICATES = {e.name for e in BinaryPredicate}
 
 
 class STRtree:
@@ -50,9 +50,12 @@ class STRtree:
 
     def query(self, geometry, predicate=None):
         """Return all items whose extent intersect the envelope of the input
-        geometry.  If predicate is provided, these items are limited to those
-        that satisfy the predicate operation when compared against the input
         geometry.
+
+        If predicate is provided, a prepared version of the input geometry
+        is tested using that predicate function against each item whose
+        extent intersects the envelope of the input geometry:
+        predicate(geometry, tree_geometry).
 
         If geometry is None, an empty array is returned.
 
@@ -67,7 +70,7 @@ class STRtree:
         """
 
         if geometry is None:
-            return np.array([], dtype="int")
+            return np.array([], dtype=np.intp)
 
         if predicate is None:
             predicate = 0
@@ -80,7 +83,7 @@ class STRtree:
                     )
                 )
 
-            predicate = UnaryPredicate[predicate].value
+            predicate = BinaryPredicate[predicate].value
 
         return self._tree.query(geometry, predicate)
 
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -38,8 +38,14 @@ class STRtree:
     >>> import pygeos
     >>> geoms = pygeos.points(np.arange(10), np.arange(10))
     >>> tree = pygeos.STRtree(geoms)
-    >>> tree.query(pygeos.box(2, 2, 4, 4)).tolist()
+    >>> geom = pygeos.box(2, 2, 4, 4)
+    >>> # Query geometries that overlap envelope of `geom`:
+    >>> tree.query(geom).tolist()
     [2, 3, 4]
+    >>> # Query geometries that overlap envelope of `geom`
+    >>> # and are contained by `geom`:
+    >>> tree.query(geom, predicate='contains').tolist()
+    [3]
     """
 
     def __init__(self, geometries, leafsize=5):
@@ -53,7 +59,7 @@ class STRtree:
         geometry.
 
         If predicate is provided, a prepared version of the input geometry
-        is tested using that predicate function against each item whose
+        is tested using the predicate function against each item whose
         extent intersects the envelope of the input geometry:
         predicate(geometry, tree_geometry).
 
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
@brendan-ward thanks a lot for looking at this!

First gut reaction: is there a performance difference compared to doing this filtering in python using pygeos? 

```
geoms = ...
geometry = ...
tree = pygeos.STRtree(geoms)
indices = tree.query(geometry)
subset = pygeos.intersects(geoms[indices], geometry)
```

This PR avoids creating the full `indices` array (before further filtering) and the subsetting of the `geoms` numpy array in python. But I am wondering if this is significant performance wise to do it in C (as it adds complexity)

brendan-ward:
@jorisvandenbossche 

Right now, the performance (based on a few contrived tests) between either approach is nearly equivalent.  I was a little surprised, I was expecting a bit better performance by doing more of this in C.  There might be something I'm doing in C that is slowing it down.

I think the bigger benefits may be building off this to use prepared geometries after querying the tree and before testing the predicate functions.  I think it would be possible to first query the tree, then for any geometry that had > N (arbitrary threshold) of results from the tree, to create a prepared geometry and then use that for the predicate function.  My hope there was that the use of prepared geometries in that case was handled automatically (from the user's perspective), rather than something they needed to do before using the ufunc predicates.

But until I can get a working version of prepared geometries (needs more digging) as a ufunc, it's hard to compare the two approaches.

I was also hoping this would make it easier to implement #52, though the bigger issue here is how to effectively vectorize querying the tree and testing the predicate.

brendan-ward:
Using prepared geometries in the query with predicates is ~ 550x faster than without prepared geometries, for my contrived test case (polygon against 280,000 line segments, ~1,500 of which overlap the envelope of the polygon and ~500 intersect the polygon).

Comparison to ufunc predicate using prepared geometries depends on more work in #84.

Adding prepared geometries to this PR proved to be pretty easy.

caspervdw:
This is great, I really like your tests also @brendan-ward  thanks for being so thorough.

I am :+1: on merging and leaving the prepared geometry stuff for a later PR

brendan-ward:
Thanks @caspervdw 

Don't merge yet - I've got prepared geometries in place and now just trying to cleanup a few compiler warnings about pointer types...

It might be good to hold off on merging this until after I can get a first pass of #84 working so we can compare.  I fully take @jorisvandenbossche 's point about added complexity vs performance tradeoffs.   (it's easy to add complexity, it's harder to take it away later... though you aren't yet up against API compatibility promises)

brendan-ward:
@jorisvandenbossche thanks to your POC for prepared geometries, I was able to do a little testing between predicate + prepared geometries within `query()` in this PR vs `query()` as it exists currently followed by `intersects_prepared`.  For the latter, I prepared the geometries before creating the tree.

I might have been doing something wrong, but for a crude spatial join that produces ~149k intersections with the target geometries, I was seeing:
* this PR: 2.2 - 2.37s
* regular `query()` followed by `intersects_prepared`: 2.5 - 3s

It seems like the bit of extra complexity in C is reasonable given the performance (in this limited test) suggests it is a little better.  And using the prepared geometries here gives us a huge performance win.

pygeos/strtree.py
```diff
@@ -32,15 +48,41 @@ def __init__(self, geometries, leafsize=5):
...
         Parameters
         ----------
-        envelope : Geometry
-            The envelope of the geometry is taken automatically.
+        geometry : Geometry
+            The envelope of the geometry is taken automatically for
+            querying the tree.
+        predicate : str, optional (default: None)
+            The predicate to use for testing geometries from the tree
+            that are within the input geometry's envelope.
```
jorisvandenbossche: I think we should clarify here in which order the predicate is evaluated (predicate(passed_geometry, tree_geometry) vs predicate(tree_geometry, passed_geometry))

(it's not directly self-evident which of the two it would be I think)

pygeos/strtree.py
```diff
@@ -32,15 +48,41 @@ def __init__(self, geometries, leafsize=5):
...
         """
-        return self._tree.query(envelope)
+
+        if geometry is None:
+            return np.array([], dtype="int")
```
jorisvandenbossche: This is no longer handled on the C level?

pygeos/strtree.py
```diff
@@ -1,9 +1,25 @@
-from pygeos import lib
+from enum import IntEnum
 import numpy as np
+from pygeos import lib
+
 
 __all__ = ["STRtree"]
 
 
+class UnaryPredicate(IntEnum):
```
jorisvandenbossche: BinaryPredicate instead of UnaryPredicate?

pygeos/strtree.py
jorisvandenbossche: And if done here, you can use `dtype=np.intp` for consistency in the dtype

src/strtree.c
```diff
@@ -16,6 +16,39 @@
...
+    // in numpy.  There doesn't appear to be anything we can do to avoid it.
+    PyArrayObject *result = (PyArrayObject *) PyArray_SimpleNew(1, dims, NPY_INTP);
+    if (result == NULL) {
+        return NULL;
+    }
+
+    // iterate from last value to first to get back original order
+    for (i = size - 1; i >= 0; i--) {
+        ptr = PyArray_GETPTR1(result, i);
+        *ptr = (npy_intp)kv_a(npy_intp, *arr, i);
```
jorisvandenbossche: Is there a reason you use `kv_a` here instead of `kv_pop`? (I assume the use of `kv_pop` was also the reason to iterate backwards; I don't know if there was a performance reason to use kv_pop over kv_a)

jorisvandenbossche:
@brendan-ward thanks for the timings! And after thinking more about this, apart from a small speed-up, I think this also gives a nice usability improvement, of not needing to do the actual predicate check manually in a follow-up step. So certainly +1 on this PR.

I added some minor comments / questions. 

In principle, this doesn't need to be blocked by the prepared geoms PR https://github.com/pygeos/pygeos/pull/92 I think? Once that lands, we can update this to reuse cached prepared geometries if they exist. 
But thinking on that topic, what would you ideally need from https://github.com/pygeos/pygeos/pull/92? Something like a `get_geom_prepared` that gets the cached prepared pointer or if that doesn't exist yet prepares the geometry, caches it, and returns that? 
In psuedo code, I mean (+handling of None):

```
char get_geom_prepared(GeometryObject *obj, GEOSPreparedGeometry **out) {
    if obj->prepared_ptr != NULL:
        *out = obj->prepared_ptr
    else:
        obj->prepared_ptr = GEOSPrepare(handle, obj->ptr)
        *out = obj->prepared_ptr
}
```

and with that you don't need to prepare geometries and destroy the prepared geometries inside the STRtree query ?

pygeos/strtree.py
brendan-ward: I had assumed (but not sure I tested) that passing `None` would be problematic for parsing the function arguments via `PyArg_ParseTuple(args, "O!i", &GeometryType, &geometry, &predicate)`

According to the [docs](https://docs.python.org/3/c-api/arg.html) for `O!`, a `TypeError` would be raised if we pass in something of `NoneType` instead of `GeometryType`, and failing out here returns `NULL` instead of an empty array.

I'm not that familiar with `PyArg_ParseTuple`, so if there is a better way we can handle `None` we could deal with at the C level.

Can certainly change the `dtype` though.

src/strtree.c
brendan-ward: I did not test the performance of either variant.  I had assumed there might be allocations involved in `kv_pop` as you remove items from the vector, but looking more closely at the implementation I don't believe that happens.  I just wanted to access an element, not remove it from the vector.

I use `kv_A` below instead of `kv_a`, so I wasn't very consistent either.  

Looping backwards was probably a holdover from using `kv_pop`.

I'll standardize to `kv_pop` - the implementation is much simpler so it theoretically should be marginally faster.

pygeos/strtree.py
brendan-ward: Confirmed, passing `None` to the C function as it is now raises
`TypeError: argument 1 must be pygeos.lib.GEOSGeometry, not None`

src/strtree.c
brendan-ward: Changed my mind, went with `kv_A` in both cases.  It's function signature looks low complexity, so it should be as fast as `kv_pop`, and allows us to iterate forward in arrays.

pygeos/strtree.py
brendan-ward: I tried to clarify this, and also added an example to the docstring.

brendan-ward:
@jorisvandenbossche thanks for the review!

I think I have addressed everything from your comments.  I updated the documentation settings for `STRtree` so that it outputs what we want.

It looks like there is potentially a difference in behavior for `MultiPoint` intersections between GEOS 3.5 and 3.6+, since that is what is failing on Travis-CI.

brendan-ward:
If I follow [this issue](https://trac.osgeo.org/postgis/ticket/3437), this was an issue with prepared MultiPoints resolved in GEOS 3.6.

Not isolated to this PR, it got detected here only because we're using prepared geometries.

I marked them as `xfail` here, but once a minimum supported version of GEOS is set, those can be removed.

jorisvandenbossche:
I only noticed now when building locally, but this introduces a build warning:

```
src/strtree.c: In function ‘STRtree_query’:
/home/joris/miniconda3/envs/geo-dev/lib/python3.8/site-packages/numpy/core/include/numpy/__multiarray_api.h:840:10: warning: return from incompatible pointer type [-Wincompatible-pointer-types]
         (*(PyObject * (*)(PyTypeObject *, int, npy_intp const *, int, npy_intp const *, void *, int, int, PyObject *)) \
/home/joris/miniconda3/envs/geo-dev/lib/python3.8/site-packages/numpy/core/include/numpy/ndarrayobject.h:130:9: note: in expansion of macro ‘PyArray_New’
         PyArray_New(&PyArray_Type, nd, dims, typenum, NULL, NULL, 0, 0, NULL)
         ^~~~~~~~~~~
src/strtree.c:157:16: note: in expansion of macro ‘PyArray_SimpleNew’
         return PyArray_SimpleNew(1, dims, NPY_INTP);
                ^~~~~~~~~~~~~~~~~
```

Something to look into at some point.

src/strtree.c
```diff
@@ -16,6 +16,38 @@
...
+ * values onto the end. */
+
+static PyArrayObject *copy_kvec_to_npy(npy_intp_vec *arr)
+{
+    npy_intp i;
+    npy_intp size = kv_size(*arr);
+    npy_intp *ptr;
+
+    npy_intp dims[1] = {size};
+    // the following raises a compiler warning based on how the macro is defined
```
brendan-ward: @jorisvandenbossche I flagged the same compiler error you found.  I did a bit of digging into it, and couldn't resolve; it appears to come from `numpy` this way.

see [this](https://stackoverflow.com/a/32414359) and [this](https://stackoverflow.com/a/40900812)

I believe the compiler warning was there originally?

PR #91, opened 2020-01-25
Fix strtree header file 

### LINKED ISSUE #89: Compilation problem on Mac OS / clang
See https://github.com/pygeos/pygeos/pull/83

With the current codebase, @sgillies and @vincentsarago report they get the following error:

```
duplicate symbol '_STRtreeType' in:
    build/temp.macosx-10.14-x86_64-3.7/src/strtree.o
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: command 'clang' failed with exit status 1
```

Removing an `#include "strtree.h"` solves this, but then results in some warnings (see https://github.com/pygeos/pygeos/pull/83). @sgillies suggests that the better solution would be to have one source file per extension module.

caspervdw:
This fixes #89 theoretically.

The `strtree.h` file currently *defines* the symbol STRtreeType. Adding `extern` instead just *declares* STRtreeType, which is what you should do in a header file. Same issue as in #79 

I was looking for a gcc / ld flag that would detect this issue, to add on Travis so that I don't keep making this mistake. 

@sgillies please check if this works for you

### DIFF
--- a/src/strtree.h
+++ b/src/strtree.h
@@ -19,8 +19,8 @@ typedef struct
     npy_intp *a;
 } npy_intp_vec;
 
-PyTypeObject STRtreeType;
+extern PyTypeObject STRtreeType;
 
-int init_strtree_type(PyObject *m);
+extern int init_strtree_type(PyObject *m);
 
 #endif


jorisvandenbossche:
> I was looking for a gcc / ld flag that would detect this issue, to add on Travis so that I don't keep making this mistake.

We should probably try to set up CI for Mac to be able to catch this (I think Travis, Azure and Github Actions all have some options for that.

I can't test the PR, but looks good to me. Would be good if someone with Mac can confirm. cc @vincentsarago

caspervdw:
We have setup Travis on OSX already, see https://github.com/pygeos/pygeos/blob/8a7664b762327e801e0d7bbe70290604c59e4efa/.travis.yml#L13

sgillies:
Works for me on my macbook 👍

vincentsarago:
worked for me too, thanks a lot @caspervdw 
cc @jorisvandenbossche

jorisvandenbossche:
> We have setup Travis on OSX already, see 

Whoops, forgot that :) 
Do we know the reason it's not failing there?

vincentsarago:
@caspervdw @jorisvandenbossche as this fix a major problem (only for mac user?) is there a date for a feature release that could include this ?

jorisvandenbossche:
@vincentsarago it's not much work do make a release, and since we are still in early development mode, we can probably just cut one now 
(@caspervdw would you be OK with that? There are already some nice things in master since the last release, like the STRTree and from_shapely. I could cut a release tomorrow)

jorisvandenbossche:
@vincentsarago released (https://github.com/pygeos/pygeos/releases/tag/0.6), it's on PyPI, and conda-forge soon as well

Issue #75, opened 2019-11-11
ENH: get coordinates + offset arrays for "ragged array" representation


jorisvandenbossche:
Context: the "spatialpandas" project is trying to define a representation of geometries based on "ragged arrays" (nested arrays where subarrays can have varying length). This can give a representation of geometries that is independent of GEOS, and can have a faster access to the coordinates (as the coordinates of all geometries are stored as a flat array). The driving use case for this is in the datashader project (https://github.com/pyviz/datashader/), that is adding support for plotting polygons and other geometry types. Datashader is especially targetting visualizing big datasets, and then fast access to the coordinates is important. 
@jonmmease is implementing such a "ragged array" representation in https://github.com/jonmmease/spatialpandas/pull/2 (pandas ExtensionArrays (to store it in a dataframe) based on pyarrow's ListArrays). I am partly involved in that effort in giving feedback from a geopandas viewpoint (also for visualizing large geodataframes, datashader can be a very interesting approach). 

PyGEOS is of course tied to GEOS, and other representations as described above are clearly out of scope. 
But, I think it would be useful to have a fast conversion from a GEOS-based representation to such a "flat array + offsets"-based representation. And to have such conversion fast, it is probably easiest to have this live in PyGEOS (or at least provide all necessary building blocks to construct it).

We already have a `pygeos.get_coordinates`, which gives you access to the flat coordinates array, but in addition to that we would also need functionality to build up the offsets arrays.

Below follows an example of how to create such a "flat array + offsets" representation with the current pygeos API for the countries dataset of geopandas (mixture of polygon and multipolygons, but I first convert them to all multipolygons):

```python
import geopandas
import shapely.geometry
import pygeos

def ensure_multipolygon(geom): 
    if isinstance(geom, shapely.geometry.MultiPolygon): 
        return geom 
    elif isinstance(geom, shapely.geometry.Polygon): 
        return shapely.geometry.MultiPolygon([geom]) 
    elif geom is None: 
        return geom 
    else: 
        return ValueError 

df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))

# create pygeos array with all MultiPolygons from geopandas/shapely
arr = pygeos.from_wkb([ensure_multipolygon(geom).wkb for geom in df.geometry])

# the flat array of coordinates
coords = pygeos.get_coordinates(arr)
coords_flat = coords.ravel()

# the offsets into the multipolygon parts
[...]

PR #93, opened 2020-01-31
Add cython algo to get offset arrays for flat coordinates representation

jorisvandenbossche:
An experiment with cython to tackle https://github.com/pygeos/pygeos/issues/75: creating a representation with a flat array of coordinates and offset arrays. Focusing for now on the (Multi)Polygon use case. 
And trying it in cython, for this use case I think it simplifies it quite a bit compared to C (for array creation)

(note: the cython code certainly still needs to be cleaned up, if we want to use it, just dumped everything in a single file now)

See https://nbviewer.jupyter.org/gist/jorisvandenbossche/398ac85012e1c0214633e6d3cdd3ae70 for example of using this to convert a geopandas GeoDataFrame to this representation, to then visualize it with datashader (which needs this representation via spatialpandas).


jorisvandenbossche:
cc @jonmmease

caspervdw:
I have been thinking about this serialization format. I think it would make sense if we would come up with something that covers all different geometries and not just polygons.

One way might be using WKB without the actual coordinates. We could encode everything in one list of 4-byte integers:

 - for point: type
 - for linestring: type, number of coordinates
 - for polygon: type, coordinates in exterior ring, number of interior rings, [number of coordinates in each interior ring]
 - for collections: type, number of geometries, [wkb-without-coords of each geometry]

see https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry

Would this approach work for your application to connect a GeoDataFrame to datashader?

jorisvandenbossche:
To be clear, this PR only implements it for MultiPolygons, but it's not specific to polygons. All other types can be stored as (possibly) nested flat coordinate arrays as well. You only need to know which type of geometry it represents. But I assume it is this aspect (storing this information directly in the data) you are thinking about.

How would you store this type information with the coordinates?
Or do you mean storing something like this per geometry in a separate array, instead of the separate array(s) of indices (indices into the flat coordinates)?

> Would this approach work for your application to connect a GeoDataFrame to datashader?

Not directly, as we are using the Arrow memory layout (for list type: https://arrow.apache.org/docs/format/Columnar.html#variable-size-list-layout). 
At least, that's the memory layout that datashader can now understand and render (not saying this is the only possible layout, but changes to it will require (possibly non-trivial) changes in datashader / spatialpandas). 

One big advantage of using that layout (and that's the main reason I originally proposed that) is that it is also supported out of the box by parquet file format. 
Of course, something generic variable sized binary is also natively supported by Apache Arrow / Parquet, but the disadvantage is that this will always require custom code to interpret, while the list type can natively be inspected as numpy arrays (with only using pyarrow).

> One way might be using WKB without the actual coordinates. We could encode everything in one list of 4-byte integers:

Regarding efficient serialization formats, there is also https://github.com/bjornharrtell/flatgeobuf/blob/master/src/fbs/feature.fbs if we want to have something that includes geometry type information (but without inventing our own format)

jorisvandenbossche:
With the latest pygeos with the addition of `get_parts`, and potentially `get_rings` (https://github.com/pygeos/pygeos/pull/342), the functionality of this PR can also be mimicked with those functions:

```python
def get_flat_coords_offset_arrays(arr):
    # explode/flatten the MultiPolygons
    arr_flat, part_indices = pygeos.get_parts(arr, return_index=True)
    # the offsets into the multipolygon parts
    offsets1 = np.insert(np.bincount(part_indices).cumsum(), 0, 0)

    # explode/flatten the Polygons into Rings
    arr_flat2, ring_indices = pygeos.geometry.get_rings(arr_flat, return_index=True)
    # the offsets into the exterior/interior rings of the multipolygon parts 
    offsets2 = np.insert(np.bincount(ring_indices).cumsum(), 0, 0)

    # the coords and offsets into the coordinates of the rings
    coords, indices = pygeos.get_coordinates(arr_flat2, return_index=True)
    offsets3 = np.insert(np.bincount(indices).cumsum(), 0, 0)
    
    return coords, offsets1, offsets2, offsets3
```

(it's the same as what I showed in the issue (https://github.com/pygeos/pygeos/issues/75#issue-520893101), but then with faster `get_parts`/`get_rings` instead of custom python loops for those)

However, there is still a quite big difference in performance. 

First using a small example to test:
```python
import geopandas
import pygeos
from pygeos.flatcoords import get_offset_arrays

df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))
arr = df.geometry.array.data

coords = pygeos.get_coordinates(arr)
offsets1, offsets2, offsets3 = get_offset_arrays(arr)

coordsb, offsets1b, offsets2b, offsets3b = get_offset_arrays_pygeos(arr)

# check both are equivalent (this passes)
np.testing.assert_allclose(offsets1, offsets1b)
np.testing.assert_allclose(offsets2, offsets2b)
np.testing.assert_allclose(offsets3, offsets3b)
np.testing.assert_allclose(coords, coordsb)
```

Comparing both:

```
In [4]: %timeit get_flat_coords_offset_arrays(arr)
719 µs ± 24 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %%timeit
   ...: coords = pygeos.get_coordinates(arr)
   ...: offsets1, offsets2, offsets3 = get_offset_arrays(arr)
220 µs ± 39.3 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
So the single-pass cython algo it is considerably faster (well, still a different pass for the coordinates right now). And note that it's mainly getting the coordinates that takes time, as the offsets itself is quite fast:

```
In [6]: %timeit offsets1, offsets2, offsets3 = get_offset_arrays(arr)
27.6 µs ± 707 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

This is of course only a tiny toy dataset. 
Using a different / larger dataset:

```
In [3]: df_tl = geopandas.read_file("/home/joris/Downloads/tl_2019_us_zcta510.zip", rows=10000,
   ...:                             ignore_fields=['ZCTA5CE10', 'GEOID10', 'CLASSFP10', 'MTFCC10', 'FUNCSTAT10', 'ALAND10', 'AWATER10', 'INTPTLAT10', 'INTPTLON10'])

In [4]: arr = df_tl.geometry.array.data

In [5]: %timeit get_flat_coords_offset_arrays(arr)
574 ms ± 38.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %%timeit
   ...: coords = pygeos.get_coordinates(arr)
   ...: offsets1, offsets2, offsets3 = get_offset_arrays(arr)
223 ms ± 3.15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Relative speed-up is a bit smaller in this case, but the cython algo still gives a decent improvement.

As comparison, both are faster than conversion to WKB:

```
In [7]: %timeit pygeos.to_wkb(arr)
876 ms ± 14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #96, opened 2020-02-19
Benchmarking suite (asv)

caspervdw:
See here for an example result:

https://pygeos.github.io/pygeos/


caspervdw:
Test failure is unrelated ==> #97

jorisvandenbossche:
Thanks for getting this started! (now we should add some more benchmarks ;-))

Issue #97, opened 2020-02-19
BUG: Segfault in test_strtree.py::test_flush_geometries


caspervdw:
See https://travis-ci.org/pygeos/pygeos/jobs/652638801

The master build worked. I took a diff between the dependencies in the master build and the dependencies in the link and there was only one:
```
pycparser-2.19             |             py_2          88 KB  conda-forge
```
Instead of
```
pycparser-2.19             |           py37_1         171 KB  conda-forge
```

Which seems completely unrelated, although I don't know how pycparser is used in our test suite.

jorisvandenbossche:
Are you able to reproduce it locally with the new dependencies?

caspervdw:
No I haven’t tried to set up the environment yet.

brendan-ward:
@caspervdw this appears resolved by #100

PR #100, opened 2020-02-29
Fix segfault in strtree test_flush_geometries

caspervdw:
This reproduces #97  locally

It seems like we need to keep references to the geometries that were used to construct an STRTree.

### DIFF
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -53,7 +53,7 @@ class STRtree:
         self._tree = lib.STRtree(self.geometries, leafsize)
 
     def __len__(self):
-        return self._tree.get_size()
+        return self._tree.count
 
     def query(self, geometry, predicate=None):
         """Return all items whose extent intersect the envelope of the input
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -49,10 +49,11 @@ class STRtree:
     """
 
     def __init__(self, geometries, leafsize=5):
-        self._tree = lib.STRtree(np.asarray(geometries, dtype=np.object), leafsize)
+        self.geometries = np.asarray(geometries, dtype=np.object)
+        self._tree = lib.STRtree(self.geometries, leafsize)
 
     def __len__(self):
-        return self._tree.count
+        return self._tree.get_size()
 
     def query(self, geometry, predicate=None):
         """Return all items whose extent intersect the envelope of the input
@@ -92,8 +93,3 @@ class STRtree:
             predicate = BinaryPredicate[predicate].value
 
         return self._tree.query(geometry, predicate)
-
-    @property
-    def geometries(self):
-        """Return the array_like of geometries used to construct the STRtree."""
-        return self._tree.geometries
--- a/src/strtree.c
+++ b/src/strtree.c
@@ -73,7 +73,7 @@ static PyObject *STRtree_new(PyTypeObject *type, PyObject *args,
     void *tree, *ptr;
     npy_intp n, i, count = 0;
     GEOSGeometry *geom;
-    goes_geom_vec _geoms;
+    geom_obj_vec _geoms;
     GeometryObject *obj;
     GEOSContextHandle_t context = geos_context[0];
 
--- a/src/strtree.c
+++ b/src/strtree.c
@@ -58,7 +58,7 @@ static void STRtree_dealloc(STRtreeObject *self)
     // free the geometries
     size = kv_size(self->_geoms);
     for (i = 0; i < size; i++) {
-        Py_DECREF(kv_pop(self->_geoms));
+        Py_XDECREF(kv_pop(self->_geoms));
     }
     kv_destroy(self->_geoms);
     // free the PyObject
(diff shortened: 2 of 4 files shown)


caspervdw:
I traced the issue to:

https://github.com/pygeos/pygeos/blob/56ad4d09ae1fae1518dace7d8134c22a79e5c106/src/strtree.c#L56

Apparently, decreasing the refcount of `STRtree->geometries`  before deallocating the tree _sometimes_ leads to a segfault. I am quite sure that we should decrease the refcount here though. I also established that doesn't really matter what you put in to the array (you just have to replace the geometries with something else to get the issue).

caspervdw:
@jorisvandenbossche  This is done. The only way I could remove the segfault is by letting STRtree keep a reference to the geometries it is constructed with. Apparently this is needed. This is not clear from the docs nor from the GEOS source code.

https://geos.osgeo.org/doxygen/classgeos_1_1index_1_1strtree_1_1STRtree.html#a9ca89d79af23375493acbc8286e7e3e4

src/strtree.c
```diff
@@ -136,23 +154,22 @@ void query_callback(void *item, void *user_data)
...
-    npy_intp *geom_ptr;
     FuncGEOS_YpY_b *predicate_func;
     PyArrayObject *result;
 
     if (self->ptr == NULL) {
         PyErr_SetString(PyExc_RuntimeError, "Tree is uninitialized");
         return NULL;
     }
-    if (self->count == 0) {
+    if (kv_size(self->_geoms) == 0) {
```
caspervdw:
```suggestion
    if (self->count == 0) {
```

src/strtree.c
```diff
@@ -91,20 +99,31 @@ static PyObject *STRtree_new(PyTypeObject *type, PyObject *args,
...
+            // free the geometries
+            count = kv_size(_geoms);
+            for (i = 0; i < count; i++) { Py_XDECREF(kv_pop(_geoms)); }
+            kv_destroy(_geoms);
             return NULL;
         }
         /* skip incase obj was None */
-        if (geom == NULL) { continue; }
+        if (geom == NULL) {
+            kv_push(GeometryObject *, _geoms, NULL);
```
jorisvandenbossche: Do we need to keep nulls in the `_geoms` as well?

src/strtree.h
```diff
@@ -3,22 +3,32 @@
...
     npy_intp *a;
 } npy_intp_vec;
 
+
+/* A resizable vector with pointers */
+typedef struct
+{
+    size_t n, m;
+    GeometryObject **a;
+} goes_geom_vec;
```
jorisvandenbossche:
```suggestion
} geom_obj_vec;
```

? (it are not GEOSGeometries that get stored)

src/strtree.c
jorisvandenbossche: (ah, yes, because we later index into this to get the geometry)

src/strtree.c
```diff
@@ -91,20 +99,31 @@ static PyObject *STRtree_new(PyTypeObject *type, PyObject *args,
...
+        } else {
         /* perform the insert */
-        count++;
-        GEOSSTRtree_insert_r(context, tree, geom, (void *) i );
+            Py_INCREF(obj);
+            kv_push(GeometryObject *, _geoms, obj);
```
jorisvandenbossche: Would an alternative be to loop through `arr` and INCREF each element of it? (and keep it as a numpy array) Or is there a specific benefit of putting those in a separate vector?

jorisvandenbossche:
>  The only way I could remove the segfault is by letting STRtree keep a reference to the geometries it is constructed with. Apparently this is needed. This is not clear from the docs nor from the GEOS source code.

Something to report to GEOS?

src/strtree.c
caspervdw: The array itself is mutable, so in the case the array is changed after constructing the tree you might end up with a memory leak or a segfault

jorisvandenbossche: Ah, yes, good point

src/strtree.h
caspervdw: And a typo in geos ;) I renamed it.

src/strtree.c
```diff
@@ -52,8 +52,16 @@ static PyArrayObject *copy_kvec_to_npy(npy_intp_vec *arr)
...
 {
     void *context = geos_context[0];
+    size_t i, size;
+    // free the tree
     if (self->ptr != NULL) { GEOSSTRtree_destroy_r(context, self->ptr); }
-    Py_XDECREF(self->geometries);
+    // free the geometries
+    size = kv_size(self->_geoms);
+    for (i = 0; i < size; i++) {
+        Py_XDECREF(kv_pop(self->_geoms));
```
brendan-ward: @caspervdw 
I think this can be `Py_XDECREF(kv_A(self->_geoms, i));`, which avoids decrementing the size of the vector (one less operation).

I'll add this in #108

pygeos/strtree.py
```diff
@@ -49,7 +49,8 @@ class STRtree:
     """
 
     def __init__(self, geometries, leafsize=5):
-        self._tree = lib.STRtree(np.asarray(geometries, dtype=np.object), leafsize)
+        self.geometries = np.asarray(geometries, dtype=np.object)
```
brendan-ward: @caspervdw I'm bumping into issues merging this into #108 

```
  File "/Users/bcward/projects/pygeos/pygeos/strtree.py", line 52, in __init__
    self.geometries = np.asarray(geometries, dtype=np.object)
AttributeError: can't set attribute
```

This is conflicting with the `geometries` attribute.  How did you get past this?

brendan-ward: Oh - whoops!  Merge issue on my end!  Please ignore.

PR #102, opened 2020-03-02
ENH: Implement hash, eq, neq

caspervdw:
This implement hashing for geometry objects.

### DIFF
--- a/src/pygeom.c
+++ b/src/pygeom.c
@@ -110,6 +110,7 @@ static Py_hash_t GeometryObject_hash(GeometryObject *self)
 
 static PyObject *GeometryObject_richcompare(GeometryObject *self, PyObject *other, int op) {
   PyObject *result = NULL;
+  void *context = geos_context[0];
   if (Py_TYPE(self)->tp_richcompare != Py_TYPE(other)->tp_richcompare) {
       result = Py_NotImplemented;
   } else {
@@ -122,10 +123,10 @@ static PyObject *GeometryObject_richcompare(GeometryObject *self, PyObject *othe
         result = Py_NotImplemented;
         break;
       case Py_EQ:
-        result = (GeometryObject_hash(self) == GeometryObject_hash(other_geom)) ? Py_True : Py_False;
+        result = GEOSEqualsExact_r(context, self->ptr, other_geom->ptr, 0) ? Py_True : Py_False;
         break;
       case Py_NE:
-        result = (GeometryObject_hash(self) != GeometryObject_hash(other_geom)) ? Py_True : Py_False;
+        result = GEOSEqualsExact_r(context, self->ptr, other_geom->ptr, 0) ? Py_False : Py_True;
         break;
       case Py_GT:
         result = Py_NotImplemented;
--- a/src/pygeom.c
+++ b/src/pygeom.c
@@ -76,6 +76,69 @@ static PyObject *GeometryObject_str(GeometryObject *self)
     return GeometryObject_ToWKT(self, "%s");
 }
 
+static Py_hash_t GeometryObject_hash(GeometryObject *self)
+{
+    void *context = geos_context[0];
+    unsigned char *wkb;
+    size_t size;
+    Py_hash_t x;
+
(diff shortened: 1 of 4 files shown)


src/pygeom.c
```diff
@@ -76,6 +76,69 @@ static PyObject *GeometryObject_str(GeometryObject *self)
...
+{
+    void *context = geos_context[0];
+    unsigned char *wkb;
+    size_t size;
+    Py_hash_t x;
+
+    if (self->ptr == NULL) {
+        return -1;
+    }
+    GEOSWKBWriter *writer = GEOSWKBWriter_create_r(context);
```
brendan-ward: @caspervdw what about using [GEOSGeomToWKB_buf_r](https://github.com/libgeos/geos/blob/3.8/capi/geos_ts_c.cpp#L1370)?  It seems like that would avoid needing to manage the writer yourself.

Example [usage](https://github.com/libgeos/geos/blob/3.8/tests/thread/threadtest.c#L222-L223).

jorisvandenbossche: According to the C API header that is function is deprecated:

https://github.com/libgeos/geos/blob/c01c8c1c30091d655ba888790c2daeec1f33bae3/capi/geos_c.h.in#L220

(but not sure how/if this is enforced)

src/pygeom.c
```diff
@@ -76,6 +76,69 @@ static PyObject *GeometryObject_str(GeometryObject *self)
...
+    wkb = GEOSWKBWriter_write_r(context, writer, self->ptr, &size);
+    GEOSWKBWriter_destroy_r(context, writer);
+    if (wkb == NULL) {
+        return -1;
+    }
+    x = PyHash_GetFuncDef()->hash(wkb, size);
+    if (x == -1) {
+        x = -2;
+    } else {
+        x ^= 374761393UL;  // to make the result distinct from the actual WKB hash //
```
jorisvandenbossche: is this a random number? Or does it come from somewhere?

caspervdw: It is a prime that I stole from some other hash function in the Python source. I guess it could be any number if you use it in a XOR like I did.

caspervdw:
Thanks for the feedback. I compared 'my'  equals implementation with GEOS' equals_exact, and of course, the latter is a lot more efficient:

Going through WKB:
```In [1]: import numpy as np
In [2]: import pygeos
In [3]: a = pygeos.linestrings(np.arange(1000), np.arange(0, 2000, 2))
In [4]: b = pygeos.linestrings(np.arange(1000), np.arange(0, 2000, 2))
In [5]: %timeit a == b
82.6 µs ± 136 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
Using EqualsExact:
```
In [8]: %timeit a == b
6.32 µs ± 34.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

src/pygeom.c
caspervdw: I think it is no good to use deprecated functions.

caspervdw:
The test failure is unrelated (see #100)

jorisvandenbossche:
On the other hand, for scalar equality, we maybe don't care much about performance? 
When working with arrays, `pygeos.equals_exact(a, b)` will always be faster than `a == b` (even if the scalar comparison itself is as fast as possible)

jorisvandenbossche:
> One concern with EqualsExact might be whether we are sure this will always give the same result as hashing of WKB (to ensure equivalence between hash and eq). But I suppose this will be the case with tolerance of 0.

Maybe it would be good to check come corner cases: empty geometries, geometries with NaN coordinates.

caspervdw:
I added some tests and hit upon a peculiar bit of logic:
```
> set([float("nan"), float("nan")])
{nan, nan}
> a = float("nan"); set([a, a])
{nan}
```

Same can be done for `np.nan`. It has to do with the fact that python first does the `is` operator and then the `==`.

Related numpy issue: https://github.com/numpy/numpy/issues/9358

PR #107, opened 2020-03-06
ENH: Added total_bounds()

brendan-ward:
As mentioned in #69, `total_bounds` was deferred for later work.  I needed it, so it was a good time to add.

It collapses all dimensions down to `xmin, ymin, xmax, ymax` regardless of the shape of arrays passed to `total_bounds`.

If any inputs are `None`, empty, or otherwise returns `[nan, nan, nan,nan]` for their bounds, `total_bounds` returns `[nan, nan, nan,nan]`.

I consolidated the tests for `bounds` and added counterparts for `total_bounds`

### DIFF
--- a/pygeos/measurement.py
+++ b/pygeos/measurement.py
@@ -3,7 +3,7 @@ import numpy as np
 from . import lib
 from . import Geometry  # NOQA
 
-__all__ = ["area", "distance", "bounds","total_bounds", "length", "hausdorff_distance"]
+__all__ = ["area", "distance", "bounds", "total_bounds", "length", "hausdorff_distance"]
 
 
 def area(geometry, **kwargs):
@@ -99,7 +99,7 @@ def total_bounds(geometry, **kwargs):
     >>> total_bounds(Geometry("POLYGON EMPTY")).tolist()
     [nan, nan, nan, nan]
     >>> total_bounds([Geometry("POLYGON EMPTY"), Geometry("POINT (2 3)")]).tolist()
-    [nan, nan, nan, nan]
+    [2.0, 3.0, 2.0, 3.0]
     >>> total_bounds(None).tolist()
     [nan, nan, nan, nan]
     """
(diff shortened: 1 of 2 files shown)


martinfleis:
We have recently changed (fixed) behaviour of `total_bounds` in geopandas to ignore NaN values and return bounds even if there is empty of missing geometry - https://github.com/geopandas/geopandas/pull/1312. I would say these two should be in sync, if there is no other reason for returning `[nan, nan, nan, nan]` in such a case.

brendan-ward:
@martinfleis thanks for catching that!

What about for `bounds()` of individual geometries?  I assume that it should return valid values or `[nan, nan, nan,nan]` otherwise?  

What do we do here if all geometries return `[nan, nan, nan,nan]`?  I assume in that case, we still have to return `[nan, nan, nan,nan]` since there is no alternative.

It seems like the main change in response to your suggestion would be to make `total_bounds` return the total bounds of the underlying non-null geometries.

martinfleis:
> What about for `bounds()` of individual geometries? I assume that it should return valid values or `[nan, nan, nan, nan]` otherwise?

Yes, that is what we do in geopandas now.

> What do we do here if all geometries return `[nan, nan, nan, nan]`? I assume in that case, we still have to return `[nan, nan, nan, nan]` since there is no alternative.

Yes.

> It seems like the main change in response to your suggestion would be to make total_bounds return the total bounds of the underlying non-null geometries.

I guess it make sense to use ` np.nanmin` instead of `min` to take care of the issue. It might be easier and faster than filtering non-null and doing bounds then (but I might be wrong).

caspervdw:
Thanks for this contribution! 

In my view, the `total_bounds` returns the bounds of the union of the underlying geometries. Missing and empty geometries are basically skipped in the `union`. So I agree with @martinfleis , `total_bounds` should only return NaNs if all underlying geometries are missing or empty.

jorisvandenbossche:
Thanks!

PR #111, opened 2020-03-10
Added STRtree::nearest function

brendan-ward:
This is one of a few variants of nearest neighbor functionality based on the STRtree.

This version wraps the GEOS nearest neighbor functionality and returns the index of the nearest item in the tree.

Right now this is limited to a singular geometry operation.  I'm considering changing that to a bulk geometry operation (given array of source geometries, return nearest neighbor index for each), which would likely be more useful.

Other variants that return multiple neighbors for each source geometry could be handled in a later PR.


adriangb:
A couple of questions:
1) Are you sure that it returns the lowest index for equidistant geometries? This would be contrary to the behavior of [`libspatialindex`](https://libspatialindex.org/doxygen/classSpatialIndex_1_1RTree_1_1RTree.html#a40e20c572745a2340654b6c36e3ccd5c) used in [`Toblerity/rtree`](https://github.com/Toblerity/rtree/blob/439c009235183daf6dfd93f8950eabba2c964664/rtree/index.py#L826). This behavior is actually super important for spatial joins to be consistent. A distance-based search with 0 distance should be the same as an intersection search. Plus, it is always easy to drop the other indexes but impossible to recover them.
2) Is the distance being calculated on the bounding boxes or the geometries themselves? If the latter is possible, that would be super useful!

brendan-ward:
@adriangb good catch - I had falsely concluded the ordering of results based on conflating the order of insertion into the tree with the ordering of the extents in the evaluation of nearest neighbors.

I have a hard time following the implementation in `libspatialindex` but it looks like the case there is that it will return multiple matches if there are multiple with the same distance.

GEOS returns only one even in the case of multiple matches.  

It is a little hard to follow the [implementation](https://github.com/libgeos/geos/blob/3.8.1/src/index/strtree/STRtree.cpp#L216-L287), but it looks like it recursively tests the geometry extent against the extent of each node in the tree starting from the root.  Based on which ones have the minimum distance between extents, it then [picks](https://github.com/libgeos/geos/blob/3.8.1/src/index/strtree/BoundablePair.cpp#L90-L120) either the largest or the one with leaf nodes and continues down the tree.  At a leaf node, it then iterates over the child items and calculates their minimum distance; if the distance is lower than the minimum it has seen so far, it pushes them onto the priority queue.

Here's some examples (note: these are in the middle of a refactor of this PR to operate on multiple geoms and returns source index, tree index):

tree geoms created from bottom left to upper right:
```
>>> tree = STRtree(pygeos.points(np.arange(10), np.arange(10)))
>>> tree.nearest(pygeos.box(1, 1, 3, 3)).tolist()
[[0], [1]]   # point at bottom left of the box: <pygeos.Geometry POINT (1 1)>
```

tree geoms created from upper right to bottom left:
```
>>> tree = STRtree(pygeos.points(np.arange(10, -1, -1), np.arange(10, -1, -1)))
>>> tree.nearest(pygeos.box(1, 1, 3, 3)).tolist()
[[0], [9]]  # point at bottom left of the box: <pygeos.Geometry POINT (1 1)> 
```

However, this doesn't seem particularly consistent in more complex examples.
Example grid of points from 0,0 to 4,4 in columns:
```
>>> grid = np.mgrid[0:5, 0:5]
>>> geoms = pygeos.points(grid[0].flatten(), grid[1].flatten())
>>> tree = STRtree(geoms)
>>> tree.nearest(pygeos.box(1, 1, 3, 3)).tolist()
[[0], [7]]  # <pygeos.Geometry POINT (1 2)>
```
If we reverse the x direction so that these start at 4,4 and go to 0,0 in columns:
```
>>> grid = np.mgrid[0:5, 0:5]
>>> geoms = pygeos.points(grid[0].flatten()[::-1], grid[1].flatten())
>>> tree = STRtree(geoms)
>>> tree.nearest(pygeos.box(1, 1, 3, 3)).tolist()
[[0], [7]]  # <pygeos.Geometry POINT (3 2)>
```

These are also sensitive to the leaf size of the tree:
```
>>> grid = np.mgrid[0:5, 0:5]
>>> geoms = pygeos.points(grid[0].flatten(), grid[1].flatten())
>>> tree = STRtree(geoms, leafsize=2)
>>> tree.nearest(pygeos.box(1, 1, 3, 3)).tolist()
[[0], [6]]  # <pygeos.Geometry POINT (1 1)>
```


per your 1) We don't have any ability here to specify a minimum (or even maximum) distance, so we can't ensure that these will produce same indices as an intersection if distance is 0.  I think what you are looking for a is a more robust kNN search, but that isn't available in GEOS yet.

per your 2) Once individual items are tested against the input geometry, they are evaluated according to the distance function we provide, which calculates the distance between the geometries not their envelopes.

adriangb:
Hmm. 1) is a real deal breaker. To clarify, what I mean is that if you have intersecting geometries, ex `[bbox(-1, -1, 1, 1), bbox(-2, -2, 2, 2)]` and query for the nearest geometries to `Point(0,0)` I would expect to get both `bbox` back. That is what would happen if I did an intersection query. I think this behavior makes a lot more sense than returning an almost arbitrary index. If we can't fix that I wouldn't even bother implementing it, the behavior would be too confusing to use reliably.

brendan-ward:
@adriangb more testing or investigation into GEOS behavior certainly would be appreciated, I could be interpreting the source code or results above incorrectly.

If this is a bug in GEOS behavior, we should raise the issue over there.

Is the problem more around intersections (distance == 0) or is it also present for multiple tree geometries the same distance away, e.g., 3 geoms that are _exactly_ 5 meters away?

For the intersection case, we could theoretically try to handle this internally.  We could set a flag if our distance callback function detects that it calculated a distance of 0.  Then in this case, we could query the tree for any intersecting geometries and return all of them.

We could theoretically do something similar for a fixed distance away, but it would be more involved.  We would need to set the flag if distance function encounters the same minimum distance multiple times, then expand the envelope of the input geometry by that amount, and query the tree for those geometries, then calculate the distance to each _again_ ourselves.  Not feeling good about this option.

brendan-ward:
For what it's worth, `shapely` is also using the GEOS STRtree nearest neighbor function.

adriangb:
This applies to any equidistant geometries.

I am only using intersections as a reference because I think it is very logical that the results for that specific case should be consistent `query` with `op="intersects"`.  Then by the transitive property, the same should apply to any number equidistant geometries at any distance.

In terms of how it is implemented elsewhere, `libspatialindex` and by extension `Toblerity/rtree` (which is what `GeoPandas` uses at the moment) implement it as I described above. So for 3 tree geometries exactly 5 meters away, all 3 would be returned in no particular order.

adriangb:
Here's a fully working (as in I did 1 single test, did not try to optimize anything) example of the above in Python. Obviously this would be better to implement on the C side, and also we would want to make sure the distances aren't being computed in the callback, but anyway:
```python3
def nearest(self, geometry, return_distances=False):
    geometry = np.asarray(geometry)
    if geometry.ndim == 0:
        geometry = np.expand_dims(geometry, 0)

    input_idxs, tree_idxs = self._tree.nearest(geometry)
    # why do I need this?
    predicate = BinaryPredicate["intersects"].value
    tree_idxs_new, tree_idxs_equidistant = self._tree.query_bulk(
        self.geometries[tree_idxs], predicate
    )

    @lru_cache(maxsize=np.unique(tree_idxs).size)
    def find_equidistant(tree_idx):
        return tuple([tree_idx, *tree_idxs_equidistant[tree_idxs_new==tree_idx]])

    res = []
    for i, (input_idx, tree_idx) in enumerate(zip(input_idxs, tree_idxs)):
        all_matches = find_equidistant(tree_idx)
        res.extend(((inp, tree) for inp, tree in zip([input_idx]*len(all_matches), all_matches)))

    possible_nearest = np.unique(res, axis=0)

    distances = distance(geometry[possible_nearest[:, 0]], self.geometries[possible_nearest[:, 1]])
    min_dist = np.min(distances)
    accurate_match_indexes = distances==min_dist
    matches = possible_nearest[accurate_match_indexes]
    if return_distances:
        # return in a different array to keep dtype int?
        return np.array([matches[:, 0], matches[:, 1], distances[accurate_match_indexes]])
    else:
        return np.array([matches[:, 0], matches[:, 1]])
```

adriangb:
@brendan-ward I implemented the above in [`strtree.c`](https://github.com/adriangb/pygeos/blob/tree_nearest_agb/src/strtree.c). It seems to be working as expected. Some tests fail, but I think they just need to be fixed for the new behavior. I'm sure the implementation needs some cleanup, but maybe this can be a good plan B to getting this fixed in `GEOS`? I just don't know how feasible it is to get that fixed, I've never worked with their source.

A couple quick notes on the implementation:

1. I didn't edit the docs or test cases.
2. If possible and if it would be faster, we should change it so that `GEOSSTRtree_nearest_generic_r` operates on the bounds instead of using `distance_callback`. This wouldn't change the end result since the distances are always re-calculated.
3. It would be nice to optionally return the distances alongside the matches.

brendan-ward:
@adriangb thanks for experimenting with that!

Note: some of the tests for `nearest` will fail without being updated; some were failing anyway because multiple equidistant hits were returning non-deterministically between different CI runs (proving your point above).

I've been looking into creating a test case in GEOS to demonstrate the issue.  However, given their promise of API stability, I don't know that the existing nearest neighbor functions there can be adapted to address these issues.  Even if nearest neighbors can be returned more deterministically, the API is limited to returning a single result.  Likewise in JTS (which drives the design for GEOS), the API for [nearestNeighbor](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/index/strtree/STRtree.html) returns a single result. 

It seems like we are touching on some higher-order issues here, so it would be good to get some broader input.

@caspervdw @jorisvandenbossche As described above, the existing GEOS nearest neighbor function returns a single result, nondeterministically, in the case of multiple intersecting features with the input geometry or multiple equidistant features.  A k nearest neighbor function has not yet been implemented in GEOS (but has been added to JTS and could be ported over), but theoretically could help mitigate this issue (provided that k > equidistant neighbors).

Questions:
1. Should we add this as-is, and do our best to state clearly what users should expect in case of equidistant results?
2. Should we try to overcome this issue on the pygeos side by returning all equidistant results instead of just one?   (see above for some of the considerations here)
3. The original approach here focused on returning indexes of a similar nature as `STRtree::query_bulk`; should we extend that further to return an array of distances for each item?  Assume that we either keep the original distances as calculated by the tree callback function or that we have to calculate them for 2), rather than calculating them purely for returning distances.
4. Should we hold off on adding a nearest neighbor function against the tree until a kNN implementation is available from GEOS?

adriangb:
@brendan-ward what are your thoughts on continuing this? Now that geopandas has pygeos, this would be really useful. As per the [comment above](https://github.com/pygeos/pygeos/pull/111#issuecomment-599414972), I have a working version using the workaround of checking for intersections. Do you think that is a workable solution? If so, should I open a seperate PR or PR my branch into your existing branch that is tied with this PR?

brendan-ward:
@adriangb I wanted to pause until we got broader feedback.  There is a very reasonable tension around how far to go beyond what GEOS natively provides (tradeoffs re: complexity, maintainability, etc), so I want to make sure we're collectively in sync.

As such, I've held off on offering comments on your implementation.  I think there are some things we could optimize there, such as having the distance callback cache indexes and distances.  But for now, my recommendation would be to hold off until we have a bit more consensus.

adriangb:
Now that we are using pygeos's STRtree pretty extensively in geopandas, I think we should revive this discussion. Implementing it here would make it trivial to add an `sjoin_nearest` feature to geopandas.

brendan-ward:
I think the core question is how much we want to extend beyond the core functionality available in GEOS, which is pretty limited in this case.

Specifically:
- returning > 1 result when results are equidistant
- sorting of equidistant results regardless of number of results returned (apparently non-deterministic across GEOS versions and number of leaves setting in STRtree)

Another option would be to move forward with this more or less as-is, but noting equidistant results and sorting as a known issue to be solved later.  In the case of 1 nearest neighbor out of multiple equidistant neighbors, which one to return seems to be undefined behavior.  Is there any other library that handles that in a predictable way?

adriangb:
Agreed. I think in an ideal world, pygeos would not extend core functionality at all. But in practice, the implementation in C wouldn't be too bad, and it would be a _lot_ faster than implementing in geopandas/Python.

To address your points in particular:

1. > 1 equidistant results: this is pretty intensive processing that can easily be implemented here but would be hard to do in geopandas/python. I would do this here.
2. sorting: this can easily be implemented in geopandas/python and it would be fast with numpy, so I wouldn't worry about it here.

I personally think that returning a single nearest result in a non-deterministic fashion is could lead to a lot of confusion. It is however better than nothing. If that is the route chosen, I would suggest renaming to `sample_nearest` or something that makes it clear what the expectations are.

adriangb:
Pinging @jorisvandenbossche and @caspervdw for some input on this.

carlosg-m:
This is the most interesting discussion I've seen in the last months.

I've tried implementing a nearest neighbor join in python using [Toblerity/rtree](https://github.com/Toblerity/rtree) but failed because their implementation uses bounding boxes and only returns as nearest neighbors geometries that intersect, even if there is a closer geometry. 

There are two workarounds that consist in increasing K, or number of neighbors returned or expanding the bounding box.

The first does guarantee exact result or full recall, the second does not scale.

Shapely's STRtree has other properties that can be used to achieve this goal but unfortunately it's very slow.

Let me just ask a dumb question: is this an implementation problem or it's an algorithm problem?

Are there efficient algorithms to achieve exact nearest neighbor search between geometries without any constraint (e.g. radius) in the current literature?

What do the PostGIS and Esri people do to solve this problem? 

What about geohashes and fancy indexes like S2 and H3, do they solve this problem?

adriangb:
> the second does not scale

Just curious, have you tried the algo in https://github.com/pygeos/pygeos/pull/111#issuecomment-599414972 with rtree?

Basically you first query with `k=1`, record that distance, then use that as your bounding box. You could then do a nearest radius search or you could do what I proposed there which is to buffer the input geometry by that radius and do an interests search.

brendan-ward:
Superseded by #272

PR #113, opened 2020-03-14
Eliminate global GEOS context and release GIL

caspervdw:
Inspired by #80 I took another take at releasing the GIL. I think we do not need to have 2 inner loops as in the `get_geom` func, only static attributes are accessed and the CPython API is not used (except for the exception handling)

However, both this PR and #80 suffer from the same condition: the GeometryObjects (and contained GEOSGeometry objects) could be deallocated while not holding the GIL. I added a unittest that shows this issue.

Luckily, the GeometryObjects are immutable, so we only have to be wary of them being deleted. The way out of this seems: making a copy of the ndarray (and make sure nobody can access it) before releasing the GIL. This would add some overhead, so I guess we want to have some heuristics on the array size before doing this. Or we leave it up to the user.


jorisvandenbossche:
Thanks for looking into this! (I was wondering the same based on your comments in the subclass PR)

> The way out of this seems: making a copy of the ndarray (and make sure nobody can access it) before releasing the GIL. This would add some overhead, so I guess we want to have some heuristics on the array size before doing this. Or we leave it up to the user.

Another alternative I have been thinking about is setting the writable flag to False before passing the arrays into the pygeos.lib ufunc (a user could still change this again to modify the array while the ufunc is running, but then they are clearly on their own) 
We would need a way to reset the flag afterwards, also when an error get raised, but that might be possible to solve quite cleanly with a context manager.

I am also fine with leaving it up to the user for now.

caspervdw:
I didn't know that existed! It solves this issue nicely. Now you get a `ValueError` when trying to edit the ndarray while concurrently computing the ufunc.

pygeos/test/test_linear.py
```diff
@@ -81,5 +81,4 @@ def test_shared_paths_none():
 def test_shared_paths_non_linestring():
     g1 = pygeos.linestrings([(0, 0), (1, 0), (1, 1)])
     g2 = pygeos.points(0, 1)
-    with pytest.raises(pygeos.GEOSException):
-        pygeos.shared_paths(g1, g2)
+    raise Exception(pygeos.shared_paths(g1, g2))
```
jorisvandenbossche: Why the change here? (it looks a bit strange)

pygeos/test/test_threading.py
```diff
@@ -0,0 +1,33 @@
...
+            arr[i] = None
+
+    with ThreadPoolExecutor(max_workers=2) as executor:
+        future = executor.submit(pygeos.distance, a, b)
+        with pytest.raises(ValueError):
+            executor.submit(edit_arr, a).result()
+        future.result()
+
+
+def test_editing_ok_when_single_worker():
```
jorisvandenbossche: How does this work? It's not that we only set the writable flag if there are multiple threads?

src/pygeom.c
```diff
@@ -216,12 +216,12 @@ PyTypeObject GeometryType = {
...
 Py_None. Returns 0 on error, 1 on success. */
 char get_geom(GeometryObject *obj, GEOSGeometry **out) {
-    if (!PyObject_IsInstance((PyObject *) obj, (PyObject *) &GeometryType)) {
+    PyTypeObject *type = ((PyObject *)obj)->ob_type;
+    if ((type != &GeometryType) & (type->tp_base != &GeometryType)) {
         if ((PyObject *) obj == Py_None) {
             *out = NULL;
             return 1;
         } else {
-            PyErr_Format(PyExc_TypeError, "One of the arguments is of incorrect type. Please provide only Geometry objects.");
```
jorisvandenbossche: Other alternative would be also to re-acquire the gil here, I suppose

jorisvandenbossche: (athough reading below, I suppose the templating makes this a bit more unified)

src/ufuncs.c
```diff
@@ -398,22 +465,26 @@ static void YY_Y_func(char **args, npy_intp *dimensions,
                       npy_intp *steps, void *data)
 {
     FuncGEOS_YY_Y *func = (FuncGEOS_YY_Y *)data;
-    void *context_handle = geos_context[0];
     GEOSGeometry *in1, *in2, *ret_ptr;
 
+    GEOS_INIT;
```
jorisvandenbossche: I suppose the reason here that those can't be threaded is because of `OUTPUT_Y`? 

Specifically for those binary operations, it would be nice to get those multi-threaded as well (as those can typically be costly). 
I suppose a future improvement could be to do this in two loops (first one to gather the pointers, second one to create the geometry objects). Since I don't think python interaction can be avoided here to achieve a single loop (as you did for `get_geom`)? As we have to manage the reference counts.

src/ufuncs.c
```diff
@@ -446,19 +517,23 @@ static void Y_d_func(char **args, npy_intp *dimensions,
                      npy_intp *steps, void *data)
 {
     FuncGEOS_Y_d *func = (FuncGEOS_Y_d *)data;
-    void *context_handle = geos_context[0];
     GEOSGeometry *in1;
 
+    GEOS_INIT;
```
jorisvandenbossche: area/length could be threaded?

pygeos/test/test_linear.py
caspervdw: I must have wanted to see the result of `pygeos.shared_paths(g1, g2)` in the console. It should go out

pygeos/test/test_threading.py
caspervdw: As you have only 1 worker, the two functions cannot be ran concurrently. I think this test is actually not necessary.

src/pygeom.c
caspervdw: The python macro actually includes a brace `}` so it seems to be intended to have the GIL release and re-acquisition in the same function.

src/ufuncs.c
caspervdw: The geometries cannot be created without the GIL. There is really no way around that, as the python memory management (which allocates memory for the PyObjects) needs the GIL.

The two-step loop is a logical choice, I think we would benefit from that. There might be trade-offs however, I wonder with how many geometries this starts to pay off.

pygeos/utils.py
```diff
@@ -1,14 +1,27 @@
...
+    ndarrays to False.
+
+    NB: multithreading also requires the GIL to be released, which is done in
+    the C extension (ufuncs.c)."""
+    @wraps(func)
+    def wrapped(*args, **kwargs):
+        array_args = [
+            arg for arg in chain(args, kwargs.values())
+            if isinstance(arg, np.ndarray) and arg.dtype == object
+        ]
```
jorisvandenbossche: Alternatively, we could also specify the keywords which needs to be treated this way in the decorator. Like:

```
@multithreading_enabled(['a', 'b'])
def distance(a, b, **kwargs):
...

@multithreading_enabled(['geometry'])
def length(geometry, **kwargs):
...
``` 

That's a bit more work to write, but on the other hand is more explicit and avoids inference based on the dtype

jorisvandenbossche: Although, since you can also pass a single scalar, you would still need to check whether it is an array anyway .. 
And since the arguments can both be passed as arg or kwarg, getting the values out of that based on the position or name, will also get a bit complicated ..

(python only supports creating position only arguments starting from 3.8, otherwise we could have used it here)

src/ufuncs.c
jorisvandenbossche: The constructive ufuncs like intersection and union are typically expensive functions (and thus where pygeos does not give that much of speedup, as the looping part is small compared to the actual GEOS algo). So I think those could especially benefit from parallelizing. But indeed, would need to check in practice.

pygeos/utils.py
caspervdw: Yes, followed the same train of thought. So leave it like this? What about the name? It feels kind of misleading now.

May be `lock_object_ndarrays`?

jorisvandenbossche: Yeah, leaving it like this is fine for me.

About the name, I like the fact that it is clear what the *reason* is that we use this (because we "allow threads", as the pythn C api says), but it's true it doesn't really reflect *what it does*, so can go both ways

src/ufuncs.c
caspervdw: I like to leave this for a new issue - the PR is already large enough as is

caspervdw:
@jorisvandenbossche  I finished this PR, to enable multithreading on all functions that do not produce geometries.

Also, I eliminated the global geos context. At the end I realized it might be a good idea to have a global context after all that is only useable if you have the GIL. For instance, that would save a call to GEOSInit in the deallocation of geometries. At the other hand, it is good good to follow one pattern everywhere.

src/coords.c
```diff
@@ -286,29 +287,36 @@ PyObject *GetCoords(PyArrayObject *arr)
...
+    
+    finish:
 
-    fail:
-        Py_DECREF(result);
+        GEOS_FINISH;
         NpyIter_Deallocate(iter);
-        return NULL;
+
+        if (errstate == PGERR_SUCCESS ) {
```
brendan-ward, line 314: Suggestion: since both conditions return, make the bare return the successful one:
```
if (errstate != PGERR_SUCCESS) {
    Py_DECREF(result);
    return NULL;
}

return (PyObject *) result;
```

pygeos/test/test_misc.py
```diff
@@ -43,3 +46,65 @@ def foo():
 
     with pytest.raises(pygeos.UnsupportedGEOSOperation):
         foo()
+
+
+@multithreading_enabled
+def set_first_element(*args, **kwargs):
+    for arg in chain(args, kwargs.values()):
+        if hasattr(arg, "__setitem__"):
+            arg[0] = 42
```
brendan-ward, line 55: Suggestion: consider using a global constant, otherwise it gets a little harder to follow that 42 is being used as one (as in, it is an arbitrary rather than meaningful value).

src/coords.c
```diff
@@ -238,19 +237,22 @@ npy_intp CountCoords(PyArrayObject *arr)
     if (iternext == NULL) { NpyIter_Deallocate(iter); return -1; }
     dataptr = NpyIter_GetDataPtrArray(iter);
 
+    GEOS_INIT;
+
     do {
         /* get the geometry */
         obj = *(GeometryObject **) dataptr[0];
-        if (!get_geom(obj, &geom)) { result = -1; goto finish; }
+        if (!get_geom(obj, &geom)) { errstate = PGERR_NOT_A_GEOMETRY; result = -1; goto finish; }
```
brendan-ward, line 245: Suggestion: here and line 250, can this get expanded to multiple lines for readability?

src/geos.h
```diff
@@ -13,13 +13,92 @@
...
+        PyErr_SetString(PyExc_TypeError, "One of the Geometry inputs is of incorrect geometry type."); break;\
+      default:\
+        PyErr_Format(PyExc_RuntimeError, "Pygeos ufunc returned with unknown error state code %d.", errstate); break;\
+    }
+
+
+// Define initialization / finalization macros
+#define _GEOS_INIT_DEF\
+    char errstate = PGERR_SUCCESS;\
+    char last_error[1024] = "";\
```
brendan-ward, line 65: Are we at risk of buffer overflows here?  It looks like GEOS errors are getting set here directly without checking their length first and truncating.

src/strtree.c
```diff
@@ -208,25 +218,25 @@ void query_callback(void *item, void *user_data)
  * Returns the number of geometries that met the predicate or -1 in case of error.
  * */
 
-static int evaluate_predicate(FuncGEOS_YpY_b *predicate_func,
+static char evaluate_predicate(void* context,
```
brendan-ward, line 221: Note: the docstring above no longer matches the function signature.  Should be 
`Returns PGERR_GEOS_EXCEPTION if an error was encountered or PGERR_SUCCESS otherwise`

src/geos.h
```diff
@@ -13,13 +13,92 @@
 #pragma GCC diagnostic pop
 
 
+/* Macros to setup GEOS Context and error handlers 
+
+Typical PyGEOS pattern in a function that uses GEOS:
+
+GEOS_INIT;   // or GEOS_INIT_THREADS if you use no CPython calls
```
brendan-ward, line 20: Suggestion, it might be useful to note that `GEOS_INIT` makes `ctx` available, and that `ctx` is only valid between `GEOS_INIT` and `GEOS_FINISH`.

jorisvandenbossche:
@caspervdw thanks for your huge effort here! I will try to take another look one of the coming days

src/geos.h
caspervdw: I think not, I used `snprintf` in `geos_error_handler` and `geos_notice_handler`, which can limit the amount of characters written to the buffer.

caspervdw:
@brendan-ward  Thanks for your review, I addressed all of your comments but one, see above for specifics.

src/geos.h
```diff
@@ -13,13 +13,102 @@
...
+
+*/
+
+// Define the error states
+enum {
+	PGERR_SUCCESS,
+	PGERR_NOT_A_GEOMETRY,
+	PGERR_GEOS_EXCEPTION,
+  PGERR_NO_MALLOC,
+  PGERR_GEOMETRY_TYPE
```
jorisvandenbossche, line 50: minor style nit: some indentation misalignment above

src/coords.c
```diff
@@ -238,19 +237,30 @@ npy_intp CountCoords(PyArrayObject *arr)
...
+            result = -1;
+            goto finish;
+        }
         result += ret;
     } while(iternext(iter));
 
     finish:
+    GEOS_FINISH;
     NpyIter_Deallocate(iter);
     return result;
```
jorisvandenbossche, line 265: Another small style nit: in other places we indent the part after `finish: ` ?

src/geos.h
```diff
@@ -13,13 +13,102 @@
...
+#define GEOS_FINISH\
+    GEOS_finish_r(ctx);\
+    GEOS_HANDLE_ERR
+
+#define GEOS_FINISH_THREADS\
+    GEOS_finish_r(ctx);\
+    Py_END_ALLOW_THREADS\
+    GEOS_HANDLE_ERR
+
 #define RAISE_ILLEGAL_GEOS if (!PyErr_Occurred()) {PyErr_Format(PyExc_RuntimeError, "Uncaught GEOS exception");}
```
jorisvandenbossche, line 102: this one can be removed now? (unused as far as I see)

src/ufuncs.c
```diff
@@ -42,7 +49,6 @@ static void *is_empty_data[1] = {GEOSisEmpty_r};
 static char GEOSisSimpleAllTypes_r(void *context, void *geom) {
     int type = GEOSGeomTypeId_r(context, geom);
     if (type == -1) {
-        RAISE_ILLEGAL_GEOS;
```
jorisvandenbossche, line 45: maybe add a comment here that "2" is used to catch the error in `Y_b_func` below (as my first reaction was "don't we need to set an error?")

src/ufuncs.c
```diff
@@ -222,21 +236,25 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
...
+        if (!get_geom(*(GeometryObject **)ip1, &in1)) { errstate = PGERR_NOT_A_GEOMETRY; goto finish; }
         if (in1 == NULL) {
-            /* in case of a missing value: return NULL (NaG) */
+            /* in case of a missing value: return NULL (None) */
             ret_ptr = NULL;
         } else {
-            ret_ptr = func(context_handle, in1);
-            /* trust that GEOS calls HandleGEOSError on error */
+            ret_ptr = func(ctx, in1);
+            if ((ret_ptr == NULL) && (last_error[0] != 0)) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
```
jorisvandenbossche, line 251: why is the extra check of `last_error[0]` needed in this case? (some times it is done, and sometimes not)
(maybe add a comment about it)

src/ufuncs.c
```diff
@@ -1006,39 +1096,39 @@ static void bounds_func(char **args, npy_intp *dimensions,
...
+                coord_seq = GEOSGeom_getCoordSeq_r(ctx, ring);
+                if (coord_seq == NULL) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+                if (!GEOSCoordSeq_getX_r(ctx, coord_seq, 0, x1)) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+                if (!GEOSCoordSeq_getY_r(ctx, coord_seq, 0, y1)) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+                if (!GEOSCoordSeq_getX_r(ctx, coord_seq, 2, x2)) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+                if (!GEOSCoordSeq_getY_r(ctx, coord_seq, 2, y2)) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
             } else {
-                PyErr_Format(PyExc_ValueError, "Could not determine bounds from an envelope with %d coordinate pairs", size);
-                goto fail;
+                errstate = PGERR_GEOMETRY_TYPE;
```
jorisvandenbossche, line 1121: Can be done as a follow-up if we would want this, but I am wondering: would it be possible to set a custom error message as well with the new error handling machinery? 
Since before, in the above case a more specific (informative) error message was raised.

src/pygeom.c
```diff
@@ -78,39 +89,43 @@ static PyObject *GeometryObject_str(GeometryObject *self)
...
-    GEOSFree_r(context, wkb);
-    return x;
+    GEOSFree_r(ctx, wkb);
+
+    finish:
+        GEOS_FINISH;
+        if (errstate == PGERR_SUCCESS) {
+            return x;
+        } else {
+            return -1;
```
jorisvandenbossche, line 122: Is there a reason we return -1 here and NULL in the other type member functions? That's just what python expects for those functions?

src/coords.c
jorisvandenbossche: Although in other files we actually also don't do this consistently, so no matter

src/pygeom.c
caspervdw: That's what python expects for exceptions:

https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_hash

src/ufuncs.c
caspervdw: I think that is possible, although I am not sure if it is necessary. I'd like to keep it out of scope now.

caspervdw:
@jorisvandenbossche  Thanks for your review, I addressed your points.
@brendan-ward  @chaitan94  I touched the Frechet distance code in this PR.

Because the functions are eligible for GIL release, we can't set the PyError state from within the loop. So instead of implementing a new C-error code for the specific "out of range" error, I left it to GEOS after all. Please check if you agree on my refactored implementation.

src/ufuncs.c
```diff
@@ -222,21 +235,30 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
...
         if (in1 == NULL) {
-            /* in case of a missing value: return NULL (NaG) */
+            /* in case of a missing value: return NULL (None) */
             ret_ptr = NULL;
         } else {
-            ret_ptr = func(context_handle, in1);
-            /* trust that GEOS calls HandleGEOSError on error */
+            ret_ptr = func(ctx, in1);
+            // Check last_error if the result is NULL.
+            // We can't be sure otherwise if it is an exception or a 'missing' geometry
```
jorisvandenbossche, line 251: Isn't the case of a missing geometry already handled in the `if (in1 == NULL) {` branch? (so you shouldn't end up here in the `else` when you have a missing geometry?)

chaitan94:
Just went through https://github.com/pygeos/pygeos/commit/0f4082dac4d4af53991af18f2bfc6b30ebf07535. It looks much better now. From this it also seems - in regards to https://github.com/pygeos/pygeos/issues/150 - that to rely on `GEOSException` whenever we can is a better choice than rethrowing them unnecessarily as `ValueError`. This keeps our code simple, and we can have generic functionality like we did here with `YYd_d_func`.

src/ufuncs.c
caspervdw: I delved it a bit and it seems that for some of my own-written functions, NULL can either signify a GEOSException or a missing value. This is the case for:

 - GetPointN, GetInteriorRingN, GetGeometryN when N is out of bounds or the geometry is of incorrect type
 - GetExteriorRing if geometry is not a polygon
 - GEOSBoundaryAllTypes_r when the geometry is a collection

Maybe it is better to set an exception instead in these cases. I'll make a separate issue.

jorisvandenbossche:
Woohoo ;) We should probably do a new minor release now this is merged.

src/strtree.c
```diff
@@ -290,41 +301,46 @@ static PyObject *STRtree_query(STRtreeObject *self, PyObject *args) {
...
     }
 
+    if (predicate_id != 0) {
+        predicate_func = get_predicate_func(predicate_id);
+        if (predicate_func == NULL) {
+            return NULL;
+        }
+    }
+
+    GEOS_INIT;
```
jorisvandenbossche, line 311: @caspervdw did you try to use `GEOS_INIT_THREADS` here? 

As far as I can see, it's only `copy_kvec_to_npy` that uses Python APIs, that the calls to this can always be moved after `GEOS_FINISH_THREADS` I think. 

But if I change it to `_THREADS`, I get compilation errors.

Or is there something else I am missing that requires Python in the inner query loop? (also the GeometryObjects are put in a C vector)

caspervdw: I think we should try to implement the `_THREADS` versions in more places indeed. Let's treat that as separate issues.

caspervdw:
> Woohoo ;) We should probably do a new minor release now this is merged.

Indeed! Do you have time to wrap up https://github.com/pygeos/pygeos/pull/117 for the release?

src/strtree.c
jorisvandenbossche: Yes, separate issues/PRs for sure, but I was mainly wondering if you already tried/experienced this for the STRtree

-> https://github.com/pygeos/pygeos/issues/154

PR #116, opened 2020-03-15
Bump minimum required version of GEOS to 3.6

brendan-ward:
Updates test matrix to support only GEOS >= 3.6.  Also drops tests for Python 3.5, but that is probably OK.

Drops the xfail for `STRtree` tests that failed on GEOS 3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,29 @@
 language: minimal
 
 matrix:
-  include:
-  - os: linux  # 2015
-    env: PYTHON_VERSION="3.5" RESTORE_FREE_CHANNEL=1 DEPS="numpy=1.10 geos=3.5"
-  - os: linux  # 2017
-    env: PYTHON_VERSION="3.6" DEPS="numpy=1.13 geos=3.6"
-  - os: linux  # 2018
-    env: PYTHON_VERSION="3.7" DEPS="numpy=1.15 geos=3.7"
-  - os: linux  # 2019
-    env: PYTHON_VERSION="3.8" DEPS="numpy=1.17 geos=3.8"
-  - os: linux  # now
-    env: PYTHON_VERSION="3.8" CONDA_FORGE=1 DEPS="numpy geos"
-  - os: osx
-    env: PYTHON_VERSION="3.8" DEPS="numpy geos"
+    include:
+        - os: linux # 2017
+          env: PYTHON_VERSION="3.6" DEPS="numpy=1.13 geos=3.6"
+        - os: linux # 2018
+          env: PYTHON_VERSION="3.7" DEPS="numpy=1.15 geos=3.7"
+        - os: linux # 2019
+          env: PYTHON_VERSION="3.8" DEPS="numpy=1.17 geos=3.8"
+        - os: linux # now
+          env: PYTHON_VERSION="3.8" CONDA_FORGE=1 DEPS="numpy geos"
+        - os: osx
+          env: PYTHON_VERSION="3.8" DEPS="numpy geos"
 
 install:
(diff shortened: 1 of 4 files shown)


caspervdw:
Dropping support for 3.5 would raise issues for people on Ubuntu 16. As it's still one year in official support, many are still using it. So I'd prefer working with version conditionals (I set up already some macros for that) instead of just dropping support.

Same goes for Python 3.5.

brendan-ward:
Ok.  Maybe we could make the recommended versions more clear up front and the deprecation plan for versions of GEOS (i.e., that it is tied to the version of GEOS and Python available on the currently-supported LTS of Ubuntu).  I.e., "While GEOS 3.5 will remain minimally supported until Ubuntu 16 reaches end of life, we recommend you use at least GEOS 3.6 to take advantage of newer features or bug fixes..."

Then within individual functions that depend on newer versions, we document such in their docstrings and explicitly check version and raise appropriate errors.  Other than bug fixes that were introduced into GEOS 3.6, I think the only new feature pending is `nearest` that depends on >=3.6, but there are other things on the horizon that may need at least that version or newer.

caspervdw:
There are also a couple of functions (isCCW, and others) that were introduced to the C API after 3.5. We should probably work with a `@minimum_geos_version("3.5")` decorator that replaces the function with a dummy if necessary.

jorisvandenbossche:
In numpy / pydata land they made an information proposal with advice on how long to support versions (here specifically python and numpy), which opted for a shorter support period: https://numpy.org/neps/nep-0029-deprecation_policy.html

(not that all projects in the pydata community (already) follow that, though)

Now, as long as in practice is becomes not a too big burden to support the older versions, I am fine with keeping them (but would also be happy with dropping. But I mainly live in conda-land, so there I have access to recent version anyway).

adriangb:
I'll also add that I think `GEOSSTRtree_nearest_generic_r` was added in 3.6, so I think  #111 would require this change (or some type of decorator, as @caspervdw suggested).

caspervdw:
It would make our lives easier for sure to drop support for earlier versions. On the python and numpy side, I am :+1: on following the suggestions done by https://numpy.org/neps/nep-0029-deprecation_policy.html

On the GEOS side, I am still a bit doubtful. I know for a fact that many webservers are still on Ubuntu 16, which is fine, but they also have GEOS 3.5 there. I think it is logical that the "website world" has the tendency to "stick with what works", more than the pydata community.

At the same time, we will need the decorator approach anyway for GEOS differences between 3.6 and 3.8. So I don't think it will cost us a lot to keep 3.5 there.

jorisvandenbossche:
> At the same time, we will need the decorator approach anyway for GEOS differences between 3.6 and 3.8. So I don't think it will cost us a lot to keep 3.5 there.

+1

caspervdw:
@brendan-ward Do you agree with deciding on Python >=3.6+, Numpy >=1.15+, and GEOS >=3.5 ?

I am actually very happy with dropping Py3.5. f-strings and type annotations, here we come!

brendan-ward:
Yes - agree on those minimums.  Feel free to close this PR at any point; it's superseded by your #120

jorisvandenbossche:
For the python / numpy versions, we might want to check with Shapely before actually updating code to use those new features (if there is no agreement there about this, that would then only make it harder to upstream the code).

(will open an issue about that there)

caspervdw:
Closing through #125

mwtoews:
With Ubuntu 16.04 about to reach EOL in a few days, it's time to bump the minimum required version of GEOS to 3.6.

@brendan-ward do you want to revise this PR?

Issue #114, opened 2020-03-14
Release 0.7


caspervdw:
After the last release, we had some significant bugfixes / improvements:

 -  Include predicates in STRtree.query() (# 87)
 -  Add any pygeos/*.dll files for Windows binary .whl files (# 103) 
 -  BUG: Check types in create_collections (# 86) 
 -  ENH: Added total_bounds() (# 107) 
 -  Fix segfault in strtree test_flush_geometries (# 100) 
 -  ENH: Implement hash, eq, neq (# 102)

I think now is a good time for a release. Or is there anything that should go in before? @jorisvandenbossche

brendan-ward:
@caspervdw do we want to try and get #108 into the upcoming release, or should that get held back a little bit to further refine / discuss it's API?  (some investigations and possible changes still underway there - but wondering if I need to try and fast-track additional changes).

It might also be a good time to require GEOS >= 3.6.

jorisvandenbossche:
#108 would indeed be nice to get in (in my case to start using it in geopandas), but releases are cheap, so also fine with already releasing now.

caspervdw:
It's done now, ready to release.

Some points:
 - put a changelog somewhere in the docs
 - put a list of contributors (@brendan-ward @mwtoews #mattijn please supply your full name)
 - after the release, the first wheelbuild on pygeos-wheels can be done (deploy-test should deploy automatically to test - pypi, after that, I have to configure some stuff for the real pypi)

caspervdw:
@mattijn

brendan-ward:
@caspervdw what about using [All Contributors](https://allcontributors.org/) to help track contributors in a standardized way?

jorisvandenbossche:
OK, I already made a release: https://github.com/pygeos/pygeos/releases/tag/0.7
Did not yet upload sdist to PyPI, waiting for the wheels for that first (for which I need merge rights in pygeos-wheels)

---

I also didn't yet do the changelog / list of contributors yet (I thought that should be fine to do after the release as well, since we can update release notes after the fact).

So how do we want to acknowledge contributors? There are several options:

* Maintain an overall list of contributors in some file
  * Eg shapely does this (https://github.com/Toblerity/Shapely/blob/master/CREDITS.txt)
* List all contributors to a certain release
[...]

PR #117, opened 2020-03-19
Add release notes to the repo / docs

jorisvandenbossche:
Trying one possible way to do this (cfr https://github.com/pygeos/pygeos/issues/114#issuecomment-600866024), have the actual content in a top-level file (easily accessible when browsing on github) and including that content in the docs (easily accessible when browsing the docs).

Currently it doesn't yet render well (since the markdown is interpreted as rst). 
Options are to use rst-compatible syntax in the markdown file (eg convert single backticks to double backticks), just make CHANGELOG an rst file, or use the recommonmark sphinx extension to parse markdown (but then we still need a symbolic link to include the markdown file into the toctree, because sphinx cannot link to a file outside its root, see eg https://github.com/readthedocs/recommonmark/tree/master/docs). cfr https://github.com/sphinx-doc/sphinx/issues/7000

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,61 @@
+# Changelog
+
+## Version 0.7 (2020-03-18)
+
+**Highlights of this release**
+
+* STRtree improvements for spatial indexing:
+  * Directly include predicate evaluation in `STRtree.query()` (#87)
+  * Query multiple input geometries (spatial join style) with `STRtree.query_bulk` (#108)
+* Addition of a `total_bounds()` function (#107)
+* Geometries are now hashable, and can be compared with `==` or `!=` (#102)
+* Fixed bug in `create_collections()` with wrong types (#86) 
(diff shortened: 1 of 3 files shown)


caspervdw:
Let's be pragmatic here and keep it in the doc directory (while I would prefer a top-level CHANGES)

Why not converting to rst? That will fix the issues with rendering right?

brendan-ward:
@jorisvandenbossche thanks for working on this!

Converting to rst seems reasonable.  My preference is top-level file since that seems to be a very common convention on Github.

I think shapely gets around the imports outside root issue by updating the import path [here](https://github.com/Toblerity/Shapely/blob/1.7.0/docs/conf.py#L22).  However, I ran your PR locally without first making this change, and it seemed to work fine for the Changes as you have them setup here, so perhaps I don't understand the issue?

caspervdw:
This needs an update - I'd like for this to go in so that every PR can actually start adding to the changelog. @jorisvandenbossche  Do you have time to finish this?

PR #118, opened 2020-03-21
Include GEOS license info

caspervdw:
I read up on the LGPL 2.1 and in section 6 it is basically listed that we have to supply the LGPL license + refer to it in the docs.

There is no copyright notice in GEOS so we can't put that in our readme. (I grepped `opyright` in the whole geos distribution)

As we use dynamic linking, section 6b applies and we do not have to do anything more.

### DIFF
--- a/GEOS_LICENSE
+++ b/GEOS_LICENSE
@@ -1,0 +1,502 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
+ 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+[This is the first released version of the Lesser GPL.  It also counts
+ as the successor of the GNU Library Public License, version 2, hence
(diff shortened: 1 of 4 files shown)


caspervdw:
You're correct, it is not necessary for the sdist. 

Another route would be inserting the license file inside pygeos-wheels (and patching the manifest)

caspervdw:
I think that's not worth the effort. There is really no straightforward way to do that (including non-code files in setuptools is hard anyway).
So I rather go with this solution.

jorisvandenbossche:
Yeah, I have no problem with including the license anyway. As long as it is clear enough that PyGEOS itself is BSD

caspervdw:
So I tested it and it appears to be now only included in the sdist, and not in the bdist_wheel. The wrong way around.

After reading a very informative SO question: https://stackoverflow.com/questions/24347450/how-do-you-add-additional-files-to-a-wheel

And the python docs: https://docs.python.org/3.6/distutils/setupscript.html#installing-additional-files

I found that by removing it from the manifest, it will be included in the sdist, and by adding it to the data_files in the `setup`, it will be added to the bdist.

jorisvandenbossche:
The differences between what is includes in sdists and wheels with manifest, setup.py data_files, package_data, etc ... is annoyingly hard even for a simple package (I also ran into that with another package I was creating last week)

PR #120, opened 2020-03-22
Version conditional: make_valid

caspervdw:
Following up on #116 

MakeValid is available since GEOS 3.8.0

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -2,6 +2,7 @@ from enum import IntEnum
 import numpy as np
 from . import Geometry  # NOQA
 from . import lib
+from .geometry import needs_geos_version
 
 
 __all__ = [
@@ -14,6 +15,7 @@ __all__ = [
     "delaunay_triangles",
     "envelope",
     "extract_unique_points",
+    "make_valid",
     "point_on_surface",
     "simplify",
     "snap",
@@ -283,6 +285,26 @@ def extract_unique_points(geometry, **kwargs):
     <pygeos.Geometry MULTIPOINT EMPTY>
     """
     return lib.extract_unique_points(geometry, **kwargs)
+
+
+@needs_geos_version((3, 8, 0))
+def make_valid(geometry, **kwargs):
+    """Repairs invalid geometries.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+
+    Examples
+    --------
+    >>> make_valid(Geometry("POINT (0 0)"))
+    <pygeos.Geometry POINT (0 0)>
+    >>> make_valid(Geometry("POLYGON((0 0, 1 1, 1 2, 1 1, 0 0))"))
+    <pygeos.Geometry POLYGON((0 0, 1 1, 1 2, 1 1, 0 0))>
+    >>> make_valid(None) is None
+    True
+    """
+    return lib.make_valid(geometry, **kwargs)
 
 
 def point_on_surface(geometry, **kwargs):
--- a/pygeos/test/test_constructive.py
+++ b/pygeos/test/test_constructive.py
@@ -75,3 +75,26 @@ def test_snap_none():
 def test_snap_nan_float(geometry):
     actual = pygeos.snap(geometry, point, tolerance=np.nan)
     assert actual is None
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 8, 0), reason="GEOS < 3.8")
+def test_make_valid_none():
+    actual = pygeos.make_valid(None)
+    assert actual is None
+
+
+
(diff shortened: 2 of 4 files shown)


pygeos/geos.py
```diff
@@ -0,0 +1,26 @@
...
+from functools import wraps
+
+
+class UnsupportedGEOSOperation(ImportError):
+    pass
+
+
+class requires_geos:
+    def __init__(self, version):
+        if version.count(".") == 1:
```
brendan-ward: @caspervdw since this annotation is used by pygeos contributors (rather than users), it seems reasonable to insist on standard usage (always <major>.<minor>.<patch>) instead of inferring patch component, e.g., 
```
if version.count(".") != 2:
    raise ValueError("version must be <major>.<minor>.<patch> format")
```

otherwise as written, this doesn't handle missing minor component...

Also seems like it would be good to parse version into a tuple here for internal use, so that `TypeError`s are caught in the annotation constructor (e.g., if can't be coerced to int) rather than in `__call__`

pygeos/constructive.py
```diff
@@ -285,6 +287,22 @@ def extract_unique_points(geometry, **kwargs):
     return lib.extract_unique_points(geometry, **kwargs)
 
 
+@requires_geos("3.8.0")
+def make_valid(geometry, **kwargs):
+    """Repairs invalid geometries.
+
```
brendan-ward: Would be good to add "Requires GEOS 3.8.0" to the docstring.

pygeos/test/test_constructive.py
```diff
@@ -75,3 +75,25 @@ def test_snap_none():
...
     actual = pygeos.snap(geometry, point, tolerance=np.nan)
     assert actual is None
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 8, 0), reason="GEOS < 3.8")
+def test_make_valid_none():
+    actual = pygeos.make_valid(None)
+    assert actual is None
+
+
```
brendan-ward: Would be good to add test case for empty geometry (0d and 1d cases):

```
@pytest.mark.skipif(pygeos.geos_version < (3, 8, 0), reason="GEOS < 3.8")
@pytest.mark.parametrize(
    "geom,expected", [(empty, empty), ([empty], [empty])],
)
def test_make_valid_empty(geom, expected):
    assert pygeos.make_valid(geom) == expected
```

pygeos/test/test_constructive.py
```diff
@@ -75,3 +75,25 @@ def test_snap_none():
...
+@pytest.mark.skipif(pygeos.geos_version < (3, 8, 0), reason="GEOS < 3.8")
+@pytest.mark.parametrize(
+    "geom,expected",
+    [
+        (pygeos.points(0, 0), pygeos.points(0, 0)),
+        (
+            Geometry("POLYGON((0 0, 1 1, 1 2, 1 1, 0 0))"),
+            Geometry("MULTILINESTRING ((0 0, 1 1), (1 1, 1 2))"),
```
brendan-ward: A short comment noting what is not valid about the polygon would be helpful.

pygeos/test/test_constructive.py
```diff
@@ -75,3 +75,25 @@ def test_snap_none():
...
+        ),
+    ],
```
brendan-ward: A couple more test cases:

```
        # polygon with self-intersection (bowtie) converted to individual polygons
        (
            Geometry("POLYGON((0 0, 2 2, 2 0, 0 2, 0 0))"),
            Geometry("MULTIPOLYGON (((1 1, 0 0, 0 2, 1 1)), ((1 1, 2 2, 2 0, 1 1)))"),
        ),
        # first polygon is valid, second polygon has self-intersection
        (
            [
                Geometry("POLYGON((0 0, 2 2, 0 2, 0 0))"),
                Geometry("POLYGON((0 0, 2 2, 2 0, 0 2, 0 0))"),
            ],
            [
                Geometry("POLYGON((0 0, 2 2, 0 2, 0 0))"),
                Geometry(
                    "MULTIPOLYGON (((1 1, 0 0, 0 2, 1 1)), ((1 1, 2 2, 2 0, 1 1)))"
                ),
            ],
        ),

```

pygeos/test/test_constructive.py
```diff
@@ -75,3 +75,25 @@ def test_snap_none():
...
+)
+def test_make_valid(geom, expected):
+    actual = pygeos.make_valid(geom)
+    assert actual == expected
```
brendan-ward: Using numpy `assert_array_equal` here would let you handle 0d and 1d test cases above:

```
assert_array_equal(actual, expected)
```

caspervdw:
Thanks for the review, I updated the PR.

Issue #88, opened 2020-01-17
Standardize geometry predicate test suite


brendan-ward:
It would be good to have a common test suite for validating geometry predicate functions used in various places, and for these to have good coverage of each of the geometry types.

Using well-known and standardized geometry relations would be best, if we can find a good reference.  However, I did not find an existing test suite of geometry relationships that we could easily use as reference here.  

Some of the specialized relationships require more careful testing, as the relations are not always as obvious: contains, overlaps, touches.

It might be possible to draw from some of the test suites in JTS, PostGIS, or GEOS:
* JTS: https://github.com/locationtech/jts/tree/master/modules/tests/src/test/resources/testxml/general
* PostGIS: https://github.com/postgis/postgis/tree/master/regress/core
* GEOS: https://github.com/libgeos/geos/tree/master/tests/unit/geom

If you are open to it, I might be able to start chipping away at a more holistic geometry test suite.

PR #121, opened 2020-03-23
Softcode STRTree tests

adriangb:
I wanted to add some more tests, but it is a bit hard since the results are currently hard-coded. Since the brute force algorithms for these searches are not that hard, I propose checking the tree results in C against a brute force search in Python.

Some other ideas to think about for the future:
1. Parametrizing predicates.
2. Parametrizing tree types.
3. Parametrizing test geometries.

There are trade offs: more parametrization means more tests being run (ex: no current tests for bulk queries with "contains" parameter), but also losing directly visible information (ex: the comments on each test geometry that explain what they are testing in each case).

### DIFF
--- a/pygeos/test/test_strtree.py
+++ b/pygeos/test/test_strtree.py
@@ -11,6 +11,63 @@ from .common import point, empty, assert_increases_refcount, assert_decreases_re
 HALF_UNIT_DIAG = math.sqrt(2) / 2
 EPS = 1e-9
 
+def do_predicate(geometry, right, predicate):
+    """Applies predicate to a single geometry.
+    """
+    if not geometry or pygeos.is_empty(geometry):
+        return np.array([], dtype=int)
+    if predicate:
+        return np.flatnonzero(getattr(pygeos.predicates, predicate)(geometry, right))
+    else:
+        # no predicate means check extents intersection
+        extent_geo = pygeos.box(*pygeos.bounds(geometry))
+        if not pygeos.area(extent_geo): # point or points
+            extent_geo = geometry
+        extents_right = []
+        for geo_right in right:
+            extent = pygeos.box(*pygeos.bounds(geo_right))
+            if pygeos.area(extent):
+                extents_right.append(extent)
+            else:
+                extents_right.append(geo_right)
+        return np.flatnonzero(getattr(pygeos.predicates, "intersects")(extent_geo, extents_right))
+
+def brute_force_query_bulk(geometry, tree, predicate):
+    """Brute force search for nearest by checking all geometries.
+    """
+    left = np.atleast_1d(geometry)
+    right = np.atleast_1d(tree.geometries)
+    inds_left = []
+    inds_right = []
+    for ind_left, geom in enumerate(left):
+        matches = do_predicate(geom, right, predicate)
+        inds_right.extend(matches)
+        inds_left.extend(ind_left for _ in range(matches.size))
+    return np.array([inds_left, inds_right], dtype=int).reshape(2, -1)
+
+def brute_force_query(geometry, tree, predicate):
+    """Brute force search for nearest by checking all geometries.
+    """
+    if not geometry or pygeos.is_empty(geometry):
+        return None
+    right = np.atleast_1d(tree.geometries)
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Personally, I think there is still value in (some) hardcoded tests, for simple cases where it is relatively easy to write the expected result down (and in that way also have base tests that don't depend on yet another implementation).

So maybe we could add this *in addition* to the existing tests?

jorisvandenbossche:
Let's maybe wait a moment to hear other's thoughts before putting more work into it.

brendan-ward:
My 2 cents:

I think there is a lot of value in a trusted set of hard-coded tests, provided they are actually correct and cover the boundaries of likely errors (not saying the tests do yet).  They need to be informative and not just chasing code coverage; these are a safety net to catch breaking changes or logic errors later.

The tests (plus associated comments) also serve as a source of documentation about expected behavior that are well beyond the examples we put into docstrings or documentation.  There is also value in being able to explicitly see pairs of expected inputs and outputs; by moving expected outputs to other functions, we're hiding them behind layers of abstraction.

There is also value in [hypothesis](https://hypothesis.readthedocs.io/en/latest/) testing to further push at the boundaries of expected results vs potential coding errors.  However, I've had a bit harder time knowing how to apply hypothesis testing here, since we need to test known relations between inputs.


I'm 👎 on changes that make it harder to easily see what is going on in the test suite.  

I'm 👎 on changes that rely on the correctness of other functions to verify the correctness of a new function (e.g., relying on pygeos.intersects to validate STRtree intersects); this obfuscates the source of error when something doesn't work out as expected.  (exceptions are OK where you are testing child functions that derive from a common ancestor)

I'm 👍 on fixtures that help reduce developer effort to create trusted test suites.  E.g., per #88, perhaps there are composable sets of predefined relationships between things that we can leverage to make it easier to write new tests.

I'm 👍 on approaches to testing that help us uncover logic errors at the boundary conditions between the hard-coded manual tests and real-world (or randomly generated) data, since these are often where our logic errors are hiding. 😄 


@adriangb I'd like a better sense of where the friction is here.  Is it that the existing test cases are hard to understand or wrong?  Is it that it is hard to formulate similar pairs of expected inputs and outputs?  Something else?  Knowing where the friction is will help us better optimize next steps.

If we left the existing test suite intact, is there a way that we could leverage some of your work here for hypotethsis-oriented testing to supplement the existing tests, where the inputs are generated during the test suite rather than hard-coded?  Your approach seems like it would work well in that case, and help us better push the boundaries of what is manually tested.

adriangb:
@brendan-ward I hear you, those are good points.

As to where this came from, I was writing tests for nearest queries, and in some cases it is easy to calculate a distance, but in other cases the distances are something like 0.753873, which can only be reasonably obtained by leveraging existing functions. I do agree that relying on other functions to validate results is not ideal, but that's what I was doing to get the expected results anyway. And my hope was that if we are testing those other functions separately and they are relatively simple, the trade-off of trusting them might be worth it. But maybe I am just lazy, or maybe there is a better way I do not know about.

Maybe because I'm biased because I wrote them, but my thought with these brute force functions is that they are simple enough that it might make it _easier_ to see where the expected results come from, or test a new edge-case. That is, having a simple set of functions that define the expected behavior might make it easier to understand what is going on.

Maybe a good compromise is to keep the existing test suite, but add something like:


```python3
predicates = ("intersects", "within", "contains", "overlaps", "crosses", "touches")

@pytest.fixture
def generate_geometries():
    # generates geometries

@pytest.fixture
def point_tree(generate_geometries):
    # generates a tree

@pytest.mark.parametrize(
    "predicate", predicates,
    ids=predicates,
)
def test_query_bulk_all(generate_tree, generate_geometries, predicate):
    check_query(geometry, tree, predicate=predicate)
```

We could make the generated geometries of a certain type (points, lines, polygons, etc. kind of how the tests currently are) and check all and/or make a random assortment of geometries.

That would allow for example to add several dozen tests for `query_bulk` with `predicate="touches"`,  etc.

adriangb:
Closing for now, no need for immediate action.

Issue #71, opened 2019-10-30
Function to check or correct the "orient" of polygons


jorisvandenbossche:
Shapely has a `orient` function to make a copy of a Polygon with "proper" orientation of the rings: https://shapely.readthedocs.io/en/latest/manual.html#shapely.geometry.polygon.orient  (definition in the code: https://github.com/Toblerity/Shapely/blob/b21a964a56a86cf77c67b477db7e7c2e81d2ccb5/shapely/geometry/polygon.py#L391-L404)

In addition, Shapely also has a `is_ccw` (is counter clock wise) attribute on LinearRings: https://shapely.readthedocs.io/en/latest/manual.html#object.is_ccw 

For this, there is a GEOS C API function: https://github.com/libgeos/geos/blob/eb4da1440836fedf503e1eb2c245595a0f33db5e/capi/geos_c.h.in#L351-L357 (although Shapely implements it manually by calculating the signed area)

jorisvandenbossche:
Postgis also has a set of methods related to this: `ST_ForcePolygonCCW`, `ST_ForcePolygonCW` , `ST_IsPolygonCCW` , `ST_IsPolygonCW` (see eg https://postgis.net/docs/ST_ForcePolygonCCW.html)

jorisvandenbossche:
Possible related to this is also GEOSNormalize_r (https://postgis.net/docs/ST_Normalize.html)

caspervdw:
#126 --> implement `is_ccw`

jorisvandenbossche:
It's potentially a bit more than just `is_ccw`, eg both Shapely and Postgis also have functions to convert to consistent orientation, in addition to checking the orientation.

caspervdw:
Ok sorry I was a bit fast closing this then. Let’s investigate!

jorisvandenbossche:
I opened an equivalent function in Shapely: https://github.com/shapely/shapely/issues/1366, so let's close here.

Issue #94, opened 2020-02-19
BUG: get/set_coordinates modifies the original geometries


jorisvandenbossche:
The `set_coordinates` docstring indicates that it "Returns a copy of a geometry array with different coordinates."

However, this doesn't seem to be working correctly:

```In [31]: arr = pygeos.points([1, 2, 3], [1, 2, 3])

In [32]: arr
Out[32]: 
array([<pygeos.Geometry POINT (1 1)>, <pygeos.Geometry POINT (2 2)>,
       <pygeos.Geometry POINT (3 3)>], dtype=object)

In [33]: coords = pygeos.get_coordinates(arr) 

In [35]: new_coords = coords * 2 

In [36]: arr2 = pygeos.set_coordinates(arr, new_coords) 

In [37]: arr2  
Out[37]: 
array([<pygeos.Geometry POINT (2 2)>, <pygeos.Geometry POINT (4 4)>,
       <pygeos.Geometry POINT (6 6)>], dtype=object)

In [39]: arr  
Out[39]: 
array([<pygeos.Geometry POINT (2 2)>, <pygeos.Geometry POINT (4 4)>,
       <pygeos.Geometry POINT (6 6)>], dtype=object)
```

where the original array was modified in place.

jorisvandenbossche:
@caspervdw actually, there is a comment in the code about this function being inplace:

https://github.com/pygeos/pygeos/blob/56ad4d09ae1fae1518dace7d8134c22a79e5c106/src/coords.c#L326-L330

And it is also purposefully assigning the new geometry to a pointer in the original array:

jorisvandenbossche:
I think what makes this confusing / surprising, is: 1) I am not sure that in general you want to overwrite the existing geometries (eg in projections ("to_crs"), you might want to keep the original geometries), 2) the fact that it returns a "new" array (although this is identical (in python sense) to the passed one) also doesn't give the impression it is inplace

I think I would rather change the behaviour (although then you don't have the ability to do it inplace if you actually want), but I am also fine with keeping as is if there are good reasons for that, in which case I can do docstring update.

caspervdw:
@jorisvandenbossche This also surprises me. But on second thought, I think this is good behaviour. I would like to leave it up to the user if he wants to copy the array. I really see these `get_coordinates` and `set_coordinates` as low-level functions.

Of course the docstrings are wrong. That the array is updated inplace should be documented (with a nice red 'warning' box).

[...]

PR #123, opened 2020-03-24
ENH: add normalize

jorisvandenbossche:
xref https://github.com/pygeos/pygeos/issues/71

This is exposing `GEOSNormalize` (postgis equivalent: https://postgis.net/docs/ST_Normalize.html).

Questions:

- This is a "inplace" function in GEOS. Do we want to act it inplace as well in our python API? Otherwise I can first clone the geometry inside the ufunc function.

(still need to add tests)

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -309,6 +309,10 @@ def make_valid(geometry, **kwargs):
 def normalize(geometry, **kwargs):
     """Converts Geometry to normal form (or canonical form).
 
+    This method orders the coordinates, rings of a polygon and parts of
+    multi geometries consistently. Typically useful for testing purposes
+    (for example in combination with `equals_exact`).
+
     Parameters
     ----------
     geometry : Geometry or array_like
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -297,11 +297,9 @@ def normalize(geometry, **kwargs):
     --------
     >>> p = Geometry("MULTILINESTRING((0 0, 1 1),(2 2, 3 3))")
     >>> normalize(p)
-    >>> p
     <pygeos.Geometry MULTILINESTRING ((2 2, 3 3), (0 0, 1 1))>
     """
-    lib.normalize(geometry, **kwargs)
-    return
+    return lib.normalize(geometry, **kwargs)
 
 
 def point_on_surface(geometry, **kwargs):
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -14,6 +14,7 @@ __all__ = [
     "delaunay_triangles",
     "envelope",
     "extract_unique_points",
+    "normalize",
     "point_on_surface",
     "simplify",
     "snap",
@@ -283,6 +284,24 @@ def extract_unique_points(geometry, **kwargs):
     <pygeos.Geometry MULTIPOINT EMPTY>
     """
     return lib.extract_unique_points(geometry, **kwargs)
+
+
+def normalize(geometry, **kwargs):
+    """Converts Geometry to normal form (or canonical form).
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+
+    Examples
+    --------
+    >>> p = Geometry("MULTILINESTRING((0 0, 1 1),(2 2, 3 3))")
+    >>> normalize(p)
+    >>> p
+    <pygeos.Geometry MULTILINESTRING ((2 2, 3 3), (0 0, 1 1))>
+    """
+    lib.normalize(geometry, **kwargs)
+    return
 
 
 def point_on_surface(geometry, **kwargs):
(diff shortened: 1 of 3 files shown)


pygeos/constructive.py
```diff
@@ -285,6 +286,24 @@ def extract_unique_points(geometry, **kwargs):
     return lib.extract_unique_points(geometry, **kwargs)
 
 
+def normalize(geometry, **kwargs):
+    """Converts Geometry to normal form (or canonical form).
```
brendan-ward: Could you expand on the docstring?  I don't think that the terms normal or canonical are sufficiently well-understood by all users to know what this means (I don't know what they mean).  In particular, how it impacts the ordering of geometries within a collection / Multi or ordering of inner rings within a polygon.

jorisvandenbossche:
> Other than your work on prepared geometries, are there other instances of in-place operations in ufuncs where we perform the operation in-place rather than on a clone? 

Not a ufunc, but the `PySetCoords` / `set_coordinates` is also inplace (although the higher level `apply` that uses this is not)

pygeos/constructive.py
jorisvandenbossche: Yeah, this is just a copy of the GEOS docstring for now. 

I also don't know the details, but it might suffice to say that the coordinates / parts of Multi geometries / innter rings of a Polygon get ordered *consistently*?

jorisvandenbossche:
See also https://github.com/pygeos/pygeos/issues/94 for the confusion that `set_coordinates` being inplace caused

brendan-ward:
What about adding an `inplace=True` parameter like in pandas to all ufuncs where there may be some advantage to performing the operation in place, starting with those like `set_coordinates` that already do this.  By default, it would be `False`.  This makes the expected behavior very explicit from a user API standpoint.

caspervdw:
Thanks @jorisvandenbossche for adding this function

We should however be careful about this inplace acting of Normalize. `pygeos.Geometry `objects are *immutable* by convention, which is an essential element. This is assumed throughout the library. Copying arrays, threading, hashing, reference counting etc. will get an awful lot harder if you drop this assumption. So please, don't ever change a Geometry object inplace. The Clone overhead is just something we have to take.

So the `Normalize` can best be implemented following the pattern from `SetSRID` which also acts inplace on the GEOS side:

https://github.com/pygeos/pygeos/blob/b176e297c673671cd78f82738f1a899310501dd9/src/ufuncs.c#L361

> Other than your work on prepared geometries, are there other instances of in-place operations in ufuncs where we perform the operation in-place rather than on a clone? 

No there are none (and if you find one please tell me)

> Not a ufunc, but the PySetCoords / set_coordinates is also inplace (although the higher level apply that uses this is not)

set_coordinates acts inplace on the ndarray, not on the Geometries. It creates/destroys Geometry objects over here:

https://github.com/pygeos/pygeos/blob/b176e297c673671cd78f82738f1a899310501dd9/src/coords.c#L356

> What about adding an inplace=True parameter like in pandas to all ufuncs where there may be some advantage to performing the operation in place, starting with those like set_coordinates that already do this. By default, it would be False. This makes the expected behavior very explicit from a user API standpoint.

I am not fundamentally against this, in pandas I am actually very used to using `inplace`. But we need sufficient reason to do this work (which is quite significant). For us, we have to copy the Geometry objects anyway, so the only difference would be instantiating the ndarray. From a performance point of view, the difference would be negligible. What remains is an API design choice, and these are very opinionated. Personally I would say, not worth the effort, let's keep the API precisely as large as needed.

jorisvandenbossche:
> We should however be careful about this inplace acting of Normalize. pygeos.Geometry objects are immutable by convention, which is an essential element. This is assumed throughout the library. 

To be clear, I am fully on board with not doing this function inplace. But I raised the question because from  the set_coordinates discussion I seemed to remember we deviated from that for certain algos (but got confused about the mutating array vs mutating geometries!)

Now, about mutability, my prepared geometry PR does mutate the GeometryObject .. (but it doesn't change the actual geometry object)

> So the Normalize can best be implemented following the pattern from SetSRID which also acts inplace on the GEOS side:

Thanks for the pointer, will update the PR to take a similar approach.

I would also not add an inplace keyword now, if someone comes up with a compelling use case / provable benefit, we can always reconsider that.

caspervdw:
Great! Thanks

> Now, about mutability, my prepared geometry PR does mutate the GeometryObject .. (but it doesn't change the actual geometry object)

That might become problematic. But I'm not sure about this, because you're not changing the geometry but just adding a cache. It is a compelling solution to pack the prepared geometry and use it when present.

jorisvandenbossche:
I have been mainly thinking about it becoming problematic for the multithreading, but let's discuss that on the PR. There is still some open discussion there on what the best route is to make use of prepared geometries.

src/ufuncs.c
```diff
@@ -192,6 +192,16 @@ static void *GetExteriorRing(void *context, void *geom) {
     return ret;
 }
 static void *get_exterior_ring_data[1] = {GetExteriorRing};
+/* the normalize funcion acts inplace */
+static void *GEOSNormalize_r_with_clone(void *context, void *geom) {
+    int ret;
+    void *new_geom = GEOSGeom_clone_r(context, geom);
+    if (new_geom == NULL) { return NULL; }
+    ret = GEOSNormalize_r(context, new_geom);
+    if (ret == -1) { return NULL; }
```
caspervdw, line 201: You should destroy the ‘new_geom’ in this case

jorisvandenbossche: Ah, yes, thanks for the catch, updated.

Issue #51, opened 2019-10-06
Mix with use of cython?


jorisvandenbossche:
One thing that I was wondering: would we be fine with also writing some things in Cython instead of C (especially things that would not use the ufunc machinery but need to written more manually) ?

Some things (eg the shapely interaction I was doing in https://github.com/caspervdw/pygeos/pull/47) might be simpler to do in cython. 
On the other hand, that creates more different ways to do things, making the library as whole more complex.

caspervdw:
I agree with you on both arguments. Cython code is for sure more easy to read and write, but it will make this library more complex to manage and to build.

So I'd rather stay away from it now. Maybe at some point there will be a good reason like: we don't have the time to rewrite a certain heap of cython code.

jorisvandenbossche:
Meanwhile, I now lean towards using cython as well. 

When we add some more custom algorithms (that don't fit in either a ufunc, or are not STRtree related), I think often the benefit of coding those in C disappears. And the added complexity of also adding cython into the mix is relatively limited IMO (at least when it comes to building, it's of course a different syntax to write). 

For example, things like the PR at https://github.com/pygeos/pygeos/pull/93 for getting coordinates or the proposal in https://github.com/pygeos/pygeos/issues/127 / https://github.com/pygeos/pygeos/issues/128 for getting parts / exploding, will IMO be more approachable when we can write them in cython instead of C (especially if we limit those to 1D arrays).

jorisvandenbossche:
Any more thoughts on this? 

cc @brendan-ward

brendan-ward:
Agreed on lowering the barrier to add in new code that doesn't strictly need to be written in C.

I'm open to adapting #130 to be in Cython instead.  Should we try it there first, or is there a simpler bit of functionality where we could test out the approach?

I am a bit concerned about the additional complexities around packaging, but we can leverage some of the work already done in this space from `fiona` / `rasterio`.

jorisvandenbossche:
What *additional* complexities are you thinking of? (we already need to deal with building against and possibly including GEOS anyway) 
It's of course not battle tested at all, but in https://github.com/pygeos/pygeos/pull/93, I didn't need to do that much to get cython set up in addition to our C code to build locally.

brendan-ward:
Sorry, I forgot about #93.

There might not be any additional complexities, other than including Cython as a build dependency (needs to be added to #93 to build on CI); the hardest parts are likely already solved by having a build environment for the C extensions...

[...]

PR #130, opened 2020-03-25
WIP: get parts of a geometry collection or multi* geometries

### LINKED ISSUE #127: Function to "flatten" or "explode" multi-geometries
For several use cases in geopandas (the explode method, for plotting), it would be useful to have a function that explodes multi-geometries. 

In pseudo-python (and only for an array with only multi-geometries, would of course this needs to be generalized, but for illustration purposes), this could look like:

```
def flatten_multipolygons(arr):
   indices = [] 
   res = [] 
  
   for i, geom in enumerate(arr): 
        for j in range(pygeos.get_num_geometries(geom)): 
            poly = pygeos.get_geometry(geom, j) 
            res.append(poly)
            indices.append(i)
    
   return np.array(res, dtype=object), np.array(indices)
```

(the indices are already possible to calculate based on get_num_geometries(arr))

This is something that could be fairly easily implemented in cython, I think. Although we should check what performance benefit it can give. Thoughts?

Questions are *what* to explode though. MultiPoint/MultiLineString/MultiPolygon would be clear candidates, but GeometryCollection is maybe less clear (what with Multi geoms inside a GeometryCollection ?).

### LINKED ISSUE #128: PROPOSAL: function to return all parts of a multi-geometry as ndarray
Extracting the individual geometries from a multi geometry / collection is a fairly common task.

In shapely this is available using the `geoms` attribute on multi / collection geometry object.

There doesn't appear to be any counterpart in GEOS, nor does there need to be: it gives us the count of parts and an accessor given an index (plus, it would be hard to consume an array of geometries from GEOS).

Right now, we have to do this in a Python loop, e.g.,
```
parts = [pygeos.get_geometry(g, i) for i in range(0, pygeos.get_num_geometries(g))]
```

Which is certainly not optimal for large numbers of geometries.

There are a number of cases where we would also want to do this for all geometries in an ndarray of source geometries.

A standard ufunc approach doesn't work for this, because the output shape is greater than or equal to the input shape.  It also doesn't look like the generalized ufunc expressions get us here either, though I haven't been able to fully get my head around those.  Is there a way to do this as a ufunc?

Instead, I think we could expose a new function implemented in C that provides this functionality.

Given a singular geometry or 1D array_like of geometries, which returns a 1D array of all parts:

```
>>> geom = pg.Geometry("MULTIPOLYGON (((0 0, 0 10, 10 10, 0 0)), ((1 1, 1 10, 10 10, 1 1)))")
>>> parts = pygeos.get_parts(geom)
array([<pygeos.Geometry POLYGON ((0 0, 0 10, 10 10, 0 0))>, <pygeos.Geometry POLYGON ((1 1, 1 10, 10 10, 1 1))>])
```

For the 1D input case, we may want to know the indexes of the input geometry as well as the parts in order to relate back to the inputs (e.g., for other attributes within a DataFrame).  

In that case, we may want to limit `pygeos.get_parts()` to the single geometry case, and provide `pygeos.get_parts_bulk()` (better name welcome!!  Maybe `get_parts_indexed`?) which returns an array of shape (2,n) where the first array is indexes into source and the second is array of geometries.  

This could leverage some of the existing code used in the ufuncs, so the new code here is mostly in wrapping a bit of C as python functions.

brendan-ward:
Closes #127, closes #128 

This includes a bit of refactoring to make it easier to reuse our functions and types in C.

`vector`: types and functions for working with resizable vectors
`geom_op`: arbitrary place to put operations on geometries that are not ufuncs and not at the core of creating / converting geometries

This lifted the `GetGeometryN` from `ufuncs.c` to `geom_op.c` for reusibility in both.  There are a few other C functions in `ufuncs` that could move here as well (to keep the ufunc code focused on ufuncs), but weren't directly related to changes here.

We clearly need a better name for this function, `get_parts` is just a placeholder for now.

This takes as input a geometry or 1d array of geometry, and returns a tuple of indexes and new geometries for the parts.

GeometryCollection is not yet implemented.  There are recursive aspects of that which will require a bit more refactoring here.

I did not implement the single geometry case.  There might be a better way to handle that in a different function or method.

Key things to focus on here:
* function name
* return values

### DIFF
--- a/pygeos/geometry.py
+++ b/pygeos/geometry.py
@@ -20,6 +20,7 @@ __all__ = [
     "get_point",
     "get_interior_ring",
     "get_geometry",
+    "get_parts",
 ]
 
 
@@ -434,3 +435,33 @@ def get_num_geometries(geometry):
     1
     """
     return lib.get_num_geometries(geometry)
+
+
+def get_parts(geometry):
+    """Gets parts of each GeometryCollection or Multi* geometry object; returns
+    a copy of each singular geometry type.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+
+    Returns
+    -------
+    tuple of (indexes, parts)
+        Returns ndarray of indexes into the source geometry array and ndarray
+        of new geometries for parts or copies of the original geometries.  Arrays
+        are of equal length.
+
+    Examples
+    --------
+    >>> idx, parts = get_parts(Geometry("MULTIPOLYGON (((0 0, 0 10, 10 10, 0 0)), ((1 1, 1 10, 10 10, 1 1)))"))
+    >>> idx.tolist()
+    [0, 0]
+    >>> parts.tolist()
+    [<pygeos.Geometry POLYGON ((0 0, 0 10, 10 10, 0 0))>, <pygeos.Geometry POLYGON ((1 1, 1 10, 10 10, 1 1))>]
+    """
+    geometry = np.asarray(geometry, dtype=np.object)
+    if geometry.ndim == 0:
+        geometry = np.expand_dims(geometry, 0)
+
+    return lib.get_parts(geometry)
--- a/setup.py
+++ b/setup.py
@@ -109,7 +109,16 @@ class build_ext(_build_ext):
 
 module_lib = Extension(
     "pygeos.lib",
-    sources=["src/lib.c", "src/geos.c", "src/pygeom.c", "src/ufuncs.c", "src/coords.c", "src/strtree.c"],
+    sources=[
+        "src/lib.c",
+        "src/geos.c",
+        "src/geom_op.c",
+        "src/pygeom.c",
+        "src/ufuncs.c",
+        "src/coords.c",
+        "src/strtree.c",
+        "src/vector.c",
+    ],
     **get_geos_paths()
 )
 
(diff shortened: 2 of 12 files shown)


caspervdw:
@brendan-ward  I think we reached a consensus on the return value in https://github.com/pygeos/pygeos/issues/128

Looking at this draft PR, the work is almost done. What I think needs to be changed (functionally) is:

 - a `return_indices` argument that defaults to False (can be implemented on the Python level, just discard the second element)
 - Naming. the PostGIS equivalent is ` ST_Dump`. can we name this `dump`?

brendan-ward:
Agreed on the use of `return_indices`.

Personally, I don't like `ST_Dump` or `dump`.  It is ambiguous and doesn't give much information about what the function actually does. In ArcGIS and geopandas this is called `explode`, and it seems similar in intent to `explode` in pandas.

Could we go with `explode` or `explode_parts` instead?

I also need to add support for `GeometryCollection` via recursion, unless we want to handle that in a future PR?

jorisvandenbossche:
I agree on "dump" not being a clear name IMO (I know the analogy with PostGIS is generally nice, but here I would personally not follow it)

jorisvandenbossche:
Also, I personally think we should consider doing functions like `PyGetParts` in cython (once you need to start dealing with python function arguments, things get a lot simpler in cython), xref https://github.com/pygeos/pygeos/issues/51

brendan-ward:
Superseded by #197

PR #136, opened 2020-04-20
DOC: add missing modules

### LINKED ISSUE #133: Missing modules in docs website?
It looks like there are a few modules in `pygeos/` that aren't in the public documentation website. Specifically I was looking for documentation on the creation methods in [`creation.py`](https://github.com/pygeos/pygeos/blob/master/pygeos/creation.py), but it's missing from the [Sphinx docs](https://pygeos.readthedocs.io/en/latest/index.html).

Is that intentional?

martinfleis:
Added `pygeos.creation` and `pygeos.io` modules to API reference.

Closes #133

### DIFF
--- a/docs/creation.rst
+++ b/docs/creation.rst
@@ -1,0 +1,9 @@
+:mod:`pygeos.creation`
+======================
+
+.. automodule:: pygeos.creation
+   :members:
+   :exclude-members:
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Thanks!

PR #140, opened 2020-05-07
Ufunc atribute wrappers and documentation

adriangb:
I'm opening this to possibly discuss in the future, but mainly I want to see what happens with the docs. I pinged ReadTheDocs to get [build on pull request](https://docs.readthedocs.io/en/latest/guides/autobuild-docs-for-pull-requests.html) enabled, we'll see if it works.

### DIFF
--- a/pygeos/set_operations.py
+++ b/pygeos/set_operations.py
@@ -1,5 +1,6 @@
 import numpy as np
 from . import lib, Geometry, GeometryType
+from .ufunc_tools import ufunc_attributes_wrapper, ufunc_methods_wrapper
 
 __all__ = [
     "difference",
@@ -12,6 +13,8 @@ __all__ = [
 ]
 
 
+@ufunc_attributes_wrapper(lib.difference)
+@ufunc_methods_wrapper(lib.difference)
 def difference(a, b, **kwargs):
     """Returns the part of geometry A that does not intersect with geometry B.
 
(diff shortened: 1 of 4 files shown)


adriangb:
Interesting that all of the tests passed. @caspervdw, do you think that you could check if the webhook is properly configured, just so I can play around with these options some more and see how that impacts the documentation?

caspervdw:
@adriangb  Thanks for looking into this. The RTD webhook is setup properly, but I don't know how to setup builds for PRs.

Maybe you can try installing `sphinx` and running `make html` yourself?

About this issue: I actually took a great deal of time myself on this problem. At the end I concluding it wasn't worth the complexity. You went quite far down the rabbit hole already, but still, I think we need to discuss if this is worth it.

Complexity brings extra bugs, extra maintenance on numpy releases. Also the code becomes harder to understand. While we can also direct users that need ufunc functionality to `pygeos.lib`.

adriangb:
> The RTD webhook is setup properly, but I don't know how to setup builds for PRs.

I think (not sure) that it might be related to what is checked in the "Which events would you like to trigger this webhook?" setting.

> Complexity brings extra bugs, extra maintenance on numpy releases. Also the code becomes harder to understand. While we can also direct users that need ufunc functionality to `pygeos.lib`.

Agreed. I do see that any of these solutions is pretty complex. There are pros and cons, which you can probably judge much better than I can. I'll just leave this here in case there is any future interest, at least we know what some of the options are and what their implementation might look like.

adriangb:
To answer the documentation question: as is, the callable classes do not show up in the documentation.

Issue #126, opened 2020-03-24
Functions in GEOS > 3.5.0 to implement


caspervdw:
From the GEOS changes at https://github.com/libgeos/geos/blob/master/NEWS

Could be exposed as new functions:
- [x] GEOSBuildArea
- [x] GEOSMakeValid
- [ ] GEOSPolygonize_valid
- [x] GEOSCoverageUnion
- [x] GEOSMinimumBoundingCircle
- [x] GEOSCoordSeq_isCCW
- [x] GEOSFrechetDistance
- [x] GEOSReverse --> #243 
- [ ] GEOSSegmentIntersection
- [x] GEOSGeomGetZ
- [ ] GEOSGeom_{get,set}UserData (Rashad Kanavath)
- [x] GEOSGeom_{set,get}Precision (#257)
- [x] GEOSMinimumRotatedRectangle
- [ ] GEOSMaximumInscribedCircle
- [ ] GEOSLargestEmptyCircle
- [ ] GEOSMinimumWidth
- [x] GEOSMinimumClearance
- [ ] GEOSMinimumClearanceLine
- [x] GEOSDistanceWithin —> #417 
- [x] GEOSDensify —> #299 
- [ ] GEOSMakeValidWithParams
- [ ] GEOSWKBWriter_getFlavor, GEOSWKBWriter_setFlavor
- [ ] GEOSConstrainedDelaunayTriangulation

Having to do with strtree:
- [x] GEOSSTRtree_nearest  --> #111 
- [ ] GEOSDistanceIndexed

Functions that we could use internally:
- [ ] GEOSCoordSeq_setXY, GEOSCoordSeq_setXYZ, GEOSCoordSeq_getXY, GEOSCoordSeq_getXYZ
- [ ] GEOSGeom_createPointFromXY
- [ ] GEOSGeom_getXMin, GEOSGeom_getXMax, GEOSGeom_getYMin, GEOSGeom_getYMax 
- [x] GEOSCoordSeq_copyFromArrays, GEOSCoordSeq_copyFromBuffer,  GEOSCoordSeq_copyToArrays, GEOSCoordSeq_copyToBuffer

brendan-ward:
Crossref: `GEOSSTRtree_nearest` is in #111 with related WIP by @adriangb.

@caspervdw is there a priority order to the above?

PR #142, opened 2020-05-16
Implement GEOSCoverageUnion_r under set_operations (GEOS 3.8.0+)

chaitan94:
Implement `GEOSCoverageUnion_r` under `set_operations`, while keeping the module API similar to others, by introducing two functions `coverage_union` and `coverage_union_all`. With the latter being a reducer function of the former.

The only slightly hacky thing is that I had to test `coverage_union` slightly different from other functions of the set operations module, as it is different from them, added a comment about this in code.

See related issue https://github.com/pygeos/pygeos/issues/126.

### DIFF
--- a/pygeos/set_operations.py
+++ b/pygeos/set_operations.py
@@ -1,5 +1,6 @@
 import numpy as np
 from . import lib, Geometry, GeometryType
+from .geos import requires_geos
 
 __all__ = [
     "difference",
@@ -9,6 +10,8 @@ __all__ = [
     "symmetric_difference_all",
     "union",
     "union_all",
+    "coverage_union",
+    "coverage_union_all",
 ]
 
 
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
Thanks!

General question (more about GEOS' behaviour, but relevant for pygeos): what happens if you don't provide polygons or provide overlapping polygons? Do you get an error, or silently wrong results, or .. ?

chaitan94:
In case of overlapping polygons, it raises an error:

```py
pygeos.GEOSException: TopologyException: CoverageUnion cannot process incorrectly noded inputs.
```

I have added a doctest for the same - so it should be clear to the user is what I am expecting.

In case of us giving an input of non-polygon geometry, we get:

```py
pygeos.GEOSException: IllegalArgumentException: Unhandled geometry type in CoverageUnion.
```

However, I did not add a doctest for this. I think I should add one, what do you suggest?

pygeos/set_operations.py
```diff
@@ -193,3 +196,93 @@ def union_all(geometries, axis=0, **kwargs):
...
+    See also
+    --------
+    coverage_union_all
+
+    Examples
+    --------
+    >>> polygon = Geometry("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))")
+    >>> actual = coverage_union(polygon, Geometry("POLYGON ((1 0, 1 1, 2 1, 2 0, 1 0))"))
+    >>> from .predicates import equals
+    >>> expected = Geometry("POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))")
```
brendan-ward, line 222: I wanted to suggest avoiding `equals` to simplify the doctests, but what I'm seeing indicates that the vertices in the polygon are reordered by the operation.  E.g., 

```
>>> coverage_union(polygon, Geometry("POLYGON ((1 0, 1 1, 2 1, 2 0, 1 0))"))
<pygeos.Geometry POLYGON ((2 1, 2 0, 1 0, 0 0, 0 1, 1 1, 2 1))>

Union with None returns same polygon
>>> coverage_union(polygon, None)
<pygeos.Geometry POLYGON ((1 1, 1 0, 0 0, 0 1, 1 1))>
```

Is that why you went with `equals` here?

brendan-ward: Nevermind, just saw 5cb06b0b3cdfd70ba7cb70e267b3558616c29694

However, it would be good to use fully qualified import for `pygeos` in docstrings: `from pygeos.predicates import equals`

chaitan94: Yes, that is the reason I had to use `equals`.

I have now changed it to use absolute import instead of relative import.

chaitan94:
@jorisvandenbossche I have added tests for non-polygon inputs as well. Let me know if there's a better approach.

chaitan94:
Makes sense. I have made the changes.

However, it seems  AppVeyor and Travis have different expectations in terms of ordering. Hence I am not able to satisfy them both 😅

And it does feel a bit awkward copying the expected ordering from CI results. I feel it would be nice if there any way we can make this predictable beforehand, and maintain the readability at the same time.

Was this not a problem in any of the other doctests?

brendan-ward:
@chaitan94 [`normalize`](https://pygeos.readthedocs.io/en/latest/constructive.html#pygeos.constructive.normalize) might help here.  I'd run it manually on your input polygons then paste the outputs of that in as the inputs in your docstring tests.
(not tested...)

chaitan94:
@brendan-ward it seems that the inputs I am giving are already in their normalized version

```python
>>> from pygeos import Geometry
>>> from pygeos.constructive import normalize
>>> polygon_1 = Geometry("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))")
>>> normalize(polygon_1)
<pygeos.Geometry POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))>
>>> polygon_2 = Geometry("POLYGON ((1 0, 1 1, 2 1, 2 0, 1 0))")
>>> normalize(polygon_2)
<pygeos.Geometry POLYGON ((1 0, 1 1, 2 1, 2 0, 1 0))>
```

chaitan94:
However, I pushed a solution that meets midway using `normalize` on the outputs within the examples instead. It's not as verbose as `equals` - hence not polluting the examples too much, and still a bit readable. At the same time, it produces consistent results in both CI and local environments. I don't see any better approach otherwise as such.

caspervdw:
@chaitan94  Thanks for finishing this, I am merging this PR, it is long due. Sorry for the big delay.

PR #144, opened 2020-05-16
Implement wrappers for Fréchet distance under measurement module (3.7.0+)

chaitan94:
Code is pretty much similar to the implementation of wrappers for Hausdorff distance, except for:

1. Fréchet distance is only supported in 3.7.0, hence appropriate checks are additionally added.
2. When testing Hausdorff distance, there are tests with empty linestrings, but in the case of Fréchet distance, the GEOS library seems to be segfaulting when such inputs are passed. Hence I have skipped tests with empty linestrings for now. Any alternate suggestions over this are welcome.

Also, this is part of #126.

### DIFF
--- a/pygeos/measurement.py
+++ b/pygeos/measurement.py
@@ -2,8 +2,10 @@ import numpy as np
 
 from . import lib
 from . import Geometry  # NOQA
+from .geos import requires_geos
 
-__all__ = ["area", "distance", "bounds", "total_bounds", "length", "hausdorff_distance"]
+
+__all__ = ["area", "distance", "bounds", "total_bounds", "length", "hausdorff_distance", "frechet_distance"]
 
 
 def area(geometry, **kwargs):
@@ -172,3 +174,40 @@ def hausdorff_distance(a, b, densify=None, **kwargs):
         return lib.hausdorff_distance(a, b, **kwargs)
     else:
         return lib.haussdorf_distance_densify(a, b, densify, **kwargs)
+
+
+@requires_geos("3.7.0")
+def frechet_distance(a, b, densify=None, **kwargs):
+    """Compute the discrete Fréchet distance between two geometries.
+
+    The Fréchet distance is a measure of similarity: it is the greatest
+    distance between any point in A and the closest point in B. The discrete
+    distance is an approximation of this metric: only vertices are considered.
+    The parameter 'densify' makes this approximation less coarse by splitting
+    the line segments between vertices before computing the distance.
+
+    Fréchet distance sweep continuously along their respective curves
+    and the direction of curves is significant. This makes it a better measure
+    of similarity than Hausdorff distance for curve or surface matching.
+
+    Parameters
+    ----------
+    a, b : Geometry or array_like
+    densify : float, array_like or None
+        The value of densify is required to be between 0 and 1.
+
+    Examples
+    --------
+    >>> line_1 = Geometry("LINESTRING (0 0, 100 0)")
+    >>> line_2 = Geometry("LINESTRING (0 0, 50 50, 100 0)")
+    >>> frechet_distance(line_1, line_2)  # doctest: +ELLIPSIS
+    70.71...
+    >>> frechet_distance(line_1, line_2, densify=0.5)
+    50.0
+    >>> frechet_distance(line_1, None)
+    nan
+    """
+    if densify is None:
+        return lib.frechet_distance(a, b, **kwargs)
+    else:
+        return lib.frechet_distance_densify(a, b, densify, **kwargs)
--- a/pygeos/measurement.py
+++ b/pygeos/measurement.py
@@ -211,10 +211,4 @@ def frechet_distance(a, b, densify=None, **kwargs):
     """
     if densify is None:
         return lib.frechet_distance(a, b, **kwargs)
-
-    if not isinstance(densify, (int, float)) or np.isnan(densify) or densify <= 0 or densify > 1:
-        raise ValueError(
-                "Densify must be in range (0.0 - 1.0], got {} instead".format(densify)
-            )
-
     return lib.frechet_distance_densify(a, b, densify, **kwargs)
--- a/pygeos/measurement.py
+++ b/pygeos/measurement.py
@@ -204,6 +204,8 @@ def frechet_distance(a, b, densify=None, **kwargs):
     70.71...
     >>> frechet_distance(line_1, line_2, densify=0.5)
     50.0
+    >>> frechet_distance(line_1, Geometry("LINESTRING EMPTY"))
+    nan
     >>> frechet_distance(line_1, None)
     nan
     """
--- a/pygeos/measurement.py
+++ b/pygeos/measurement.py
@@ -212,4 +212,8 @@ def frechet_distance(a, b, densify=None, **kwargs):
     if densify is None:
         return lib.frechet_distance(a, b, **kwargs)
     else:
+        if not isinstance(densify, (int, float)) or np.isnan(densify) or densify <= 0 or densify > 1:
+            raise ValueError(
+                    "Densify must be in range (0.0 - 1.0], got {} instead".format(densify)
+                )
         return lib.frechet_distance_densify(a, b, densify, **kwargs)
--- a/pygeos/measurement.py
+++ b/pygeos/measurement.py
@@ -211,9 +211,10 @@ def frechet_distance(a, b, densify=None, **kwargs):
     """
     if densify is None:
         return lib.frechet_distance(a, b, **kwargs)
-    else:
-        if not isinstance(densify, (int, float)) or np.isnan(densify) or densify <= 0 or densify > 1:
-            raise ValueError(
-                    "Densify must be in range (0.0 - 1.0], got {} instead".format(densify)
-                )
-        return lib.frechet_distance_densify(a, b, densify, **kwargs)
+
+    if not isinstance(densify, (int, float)) or np.isnan(densify) or densify <= 0 or densify > 1:
+        raise ValueError(
+                "Densify must be in range (0.0 - 1.0], got {} instead".format(densify)
+            )
+
+    return lib.frechet_distance_densify(a, b, densify, **kwargs)
--- a/pygeos/test/test_measurement.py
+++ b/pygeos/test/test_measurement.py
@@ -241,11 +241,11 @@ def test_frechet_distance_densify(geom1, geom2, densify, expected):
 @pytest.mark.parametrize(
     "geom1, geom2",
     [
-        (point, None),
-        (None, point),
+        (line_string, None),
+        (None, line_string),
         (None, None),
-        (point, empty),
-        (empty, point),
+        (line_string, empty),
+        (empty, line_string),
         (empty, empty),
     ],
 )
@@ -266,11 +266,11 @@ def test_frechet_distance_nan_for_invalid_geometry_inputs(geom1, geom2):
     ]
 )
 def test_frechet_densify_invalid_values(densify):
-    with pytest.raises(ValueError):
-        actual = pygeos.frechet_distance(point, point, densify=densify)
+    with pytest.raises(ValueError, match=r"Densify must be in range \(0.0 - 1.0], got .* instead"):
+        actual = pygeos.frechet_distance(line_string, line_string, densify=densify)
 
 
 @pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
 def test_frechet_distance_densify_empty():
-    actual = pygeos.frechet_distance(point, empty, densify=0.2)
+    actual = pygeos.frechet_distance(line_string, empty, densify=0.2)
     assert np.isnan(actual)
--- a/pygeos/test/test_measurement.py
+++ b/pygeos/test/test_measurement.py
@@ -175,3 +175,42 @@ def test_haussdorf_distance_empty():
 def test_haussdorf_distance_densify_empty():
     actual = pygeos.hausdorff_distance(point, empty, densify=0.2)
     assert np.isnan(actual)
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
+def test_frechet_distance():
+    # example from GEOS docs
+    a = pygeos.linestrings([[0, 0], [50, 200], [100, 0], [150, 200], [200, 0]])
+    b = pygeos.linestrings([[0, 200], [200, 150], [0, 100], [200, 50], [0, 0]])
+    actual = pygeos.frechet_distance(a, b)
+    assert actual == pytest.approx(200, abs=1e-7)
+    c = pygeos.linestrings([[0, 0], [200, 50], [0, 100], [200, 150], [0, 200]])
+    actual = pygeos.frechet_distance(a, c)
+    assert actual == pytest.approx(282.842712474619, abs=1e-12)
+
+    # example from GEOS tests
+    a = pygeos.linestrings([[0, 0], [100, 0]])
+    b = pygeos.linestrings([[0, 0], [50, 50], [100, 0]])
+    actual = pygeos.frechet_distance(a, b)
+    assert actual == pytest.approx(70.7106781186548, abs=1e-12)
+
+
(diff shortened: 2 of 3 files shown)


caspervdw:
Thanks for this PR!

The reason for our extensive testing especially with empty geometries is that we avoid segfaults at all costs.

The way to solve this is to wrap the function from the GEOS CAPI on the C level. First check if the geometry is an empty line, and raise a PyError if so.

chaitan94:
Done, handled the empty geometry cases which were causing segfaults, and added tests as well.

Also, during this change I noticed 'hausdorff' was misspelled as 'haussdorf' in a few places, making it hard to be discovered through Ctrl+F. I will submit a different PR for it.

src/ufuncs.c
```diff
@@ -530,6 +530,18 @@ static PyUFuncGenericFunction Y_i_funcs[1] = {&Y_i_func};
 /* Define the geom, geom -> double functions (YY_d) */
 static void *distance_data[1] = {GEOSDistance_r};
 static void *hausdorff_distance_data[1] = {GEOSHausdorffDistance_r};
+#if GEOS_SINCE_370
+  static int GEOSFrechetDistanceWrapped_r(void *context, void *a,  void *b, double *c) {
+      /* Handle empty geometries (they give segfaults) */
+      if (GEOSisEmpty_r(context, a) | GEOSisEmpty_r(context, b)) {
+          *c = NPY_NAN;
```
brendan-ward, line 537: Suggestion: shift the logic around to keep the bare return the happy path:

```
if (GEOSisEmpty_r(context, a) | GEOSisEmpty_r(context, b)) {
    *c = NPY_NAN;
    return 1
}

return GEOSFrechetDistance_r(context, a, b, c);

File: pygeos/test/test_measurement.py [hunk 2]
```diff
@@ -175,3 +175,54 @@ def test_haussdorf_distance_empty():
 def test_haussdorf_distance_densify_empty():
     actual = pygeos.hausdorff_distance(point, empty, densify=0.2)
     assert np.isnan(actual)
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
+def test_frechet_distance():
```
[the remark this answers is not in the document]
chaitan94: Makes sense. Done, made the changes.

pygeos/test/test_measurement.py
```diff
@@ -175,3 +175,54 @@ def test_haussdorf_distance_empty():
...
+    # example from GEOS tests
+    a = pygeos.linestrings([[0, 0], [100, 0]])
+    b = pygeos.linestrings([[0, 0], [50, 50], [100, 0]])
+    actual = pygeos.frechet_distance(a, b, densify=0.001)
+    assert actual == pytest.approx(50, abs=0.1)
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
+def test_frechet_distance_missing():
+    actual = pygeos.frechet_distance(point, None)
```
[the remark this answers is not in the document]
chaitan94: Yes, originally I was thinking of adding too, but after looking at Hausdorff's tests I though maybe I am going overboard. Made the changes here for now. If it makes sense will add it for Hausdorff as well after this in a new PR.

pygeos/test/test_measurement.py
```diff
@@ -175,3 +175,54 @@ def test_haussdorf_distance_empty():
...
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
+def test_frechet_densify_nan():
+    actual = pygeos.frechet_distance(point, point, densify=np.nan)
+    assert np.isnan(actual)
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
+def test_frechet_distance_empty():
+    actual = pygeos.frechet_distance(point, empty)
```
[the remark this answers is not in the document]
chaitan94: Done.

src/ufuncs.c
chaitan94: Done.

pygeos/test/test_measurement.py
```diff
@@ -175,3 +175,79 @@ def test_haussdorf_distance_empty():
...
+            pygeos.linestrings([[0, 0], [200, 50], [0, 100], [200, 150], [0, 200]]),
+            282.842712474619
+        ),
+        # example from GEOS tests
+        (
+            pygeos.linestrings([[0, 0], [100, 0]]),
+            pygeos.linestrings([[0, 0], [50, 50], [100, 0]]),
+            70.7106781186548
+        ),
+    ],
```
brendan-ward, line 202: @chaitan94 Thank you for adding `parametrize`d tests.  What do you think about adding a 0-distance example, to validate that two identical geometries return 0 distance?

```
        # identical geometries should have 0 distance
        (
            pygeos.linestrings([[0, 0], [100, 0]]),
            pygeos.linestrings([[0, 0], [100, 0]]),
            0,
        ),
```

Likewise, one that demonstrates that direction really does matter:
```
        # same geometries but different curve direction results in maximum
        # distance between vertices on the lines.
        (
            pygeos.linestrings([[0, 0], [50, 200], [100, 0], [150, 200], [200, 0]]),
            pygeos.linestrings([[200, 0], [150, 200], [100, 0], [50, 200], [0, 0]]),
            200,
        ),
```

src/ufuncs.c
```diff
@@ -722,6 +734,28 @@ static void haussdorf_distance_densify_func(char **args, npy_intp *dimensions,
...
+{
+    void *context_handle = geos_context[0];
+    GEOSGeometry *in1, *in2;
+
+    TERNARY_LOOP {
+        /* get the geometries: return on error */
+        if (!get_geom(*(GeometryObject **)ip1, &in1)) { return; }
+        if (!get_geom(*(GeometryObject **)ip2, &in2)) { return; }
+        double in3 = *(double *) ip3;
+        if ((in1 == NULL) | (in2 == NULL) | npy_isnan(in3) | GEOSisEmpty_r(context_handle, in1) | GEOSisEmpty_r(context_handle, in2)) {
```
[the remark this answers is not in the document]
brendan-ward: If `densify` is negative, we get this:

```
pygeos.GEOSException: IllegalArgumentException: Fraction is not in range (0.0 - 1.0]
```

Personally, I think we should specifically check for this case in Python / C and raise a `ValueError` instead.  (and add a corresponding test to validate it is raised)

pygeos/test/test_measurement.py
chaitan94: Makes sense. I have added both of these test cases.

src/ufuncs.c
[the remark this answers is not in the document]
chaitan94: Sure, I think we can check if the value is specifically in the range (0.0 - 1.0] in Python, and raise a ValueError if not.

I have just pushed a commit for the same. I also excluded nan from valid values of `densify` in this commit. @caspervdw if you have any reason why we should allow `nan`, please let me know I will adjust the code accordingly. Otherwise, @brendan-ward I think we can go ahead with any other suggestions or for a merge.

pygeos/measurement.py
```diff
@@ -172,3 +174,46 @@ def hausdorff_distance(a, b, densify=None, **kwargs):
...
+    >>> frechet_distance(line_1, line_2, densify=0.5)
+    50.0
+    >>> frechet_distance(line_1, Geometry("LINESTRING EMPTY"))
+    nan
+    >>> frechet_distance(line_1, None)
+    nan
+    """
+    if densify is None:
+        return lib.frechet_distance(a, b, **kwargs)
+    else:
```
brendan-ward, line 214: Suggestion: since `if` condition returns, you can drop this `else` statement.

pygeos/test/test_measurement.py
```diff
@@ -175,3 +175,102 @@ def test_haussdorf_distance_empty():
...
+    [
+        np.nan,
+        0,
+        -1,
+        2,
+        "0.5"
+    ]
+)
+def test_frechet_densify_invalid_values(densify):
+    with pytest.raises(ValueError):
```
brendan-ward, line 269: Suggestion: to make sure we're getting the right `ValueError`:

```
with pytest.raises(ValueError, match="Densify must be in range (0.0 - 1.0]"):
```

src/ufuncs.c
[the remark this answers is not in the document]
caspervdw: The reason for allowing nan in `haussdorf_distance_densify` is that GEOS handles that situation correctly: it returns `nan`.

Whenever GEOS handles the situation correctly, there is no reason to put up another Python layer of exception handling.

pygeos/measurement.py
```diff
@@ -172,3 +174,46 @@ def hausdorff_distance(a, b, densify=None, **kwargs):
...
+        if not isinstance(densify, (int, float)) or np.isnan(densify) or densify <= 0 or densify > 1:
+            raise ValueError(
```
caspervdw, line 216: You should also take into account that `densify` could by an ndarray.

About this error handling: is it really necessary? Doesn't GEOS neatly raise an error if its out of range? If so, you don't need to check an additional time.

In addition, NaN should be propagated here into the result (and not raise an exception).

pygeos/test/test_measurement.py
```diff
@@ -175,3 +175,102 @@ def test_haussdorf_distance_empty():
...
+def test_frechet_distance_densify(geom1, geom2, densify, expected):
+    actual = pygeos.frechet_distance(geom1, geom2, densify=densify)
+    assert actual == pytest.approx(expected, abs=1e-12)
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
+@pytest.mark.parametrize(
+    "geom1, geom2",
+    [
+        (point, None),
```
brendan-ward, line 244: I also found that `(point, point)` results in `nan` too.  That seems counterintuitive; shouldn't it be 0?

src/ufuncs.c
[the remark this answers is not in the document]
brendan-ward: @caspervdw GEOS isn't handling `np.nan` handled specifically in the C code (i.e., above) - which means it wasn't a choice made inside GEOS, right?

Personally, I think while there are a number of valid cases where the inputs can be a mix of `nan` and valid values, it feels like a validation problem for parameters that control the output of the function - and are parameters that are unlikely to vary based on your input geometries.

Example: I can see cases where I could pass in an array of `densify` values directly or via broadcasting; maybe some geometries need to be densified more than others.  I'm not deriving `densify` at runtime from something else that may return `nan` (though in theory someone could; e.g., as a function fo distance), so it means that I goofed up if `nan`s are in the list, and I'd rather have to handle that specifically rather than silently getting `nan` in my output.

For example, numpy doesn't let me get away with it for parameters that control the output in a roughly similar way:
```
>>> np.repeat([1], [np.nan])
...
ValueError: cannot convert float NaN to integer
```

But - these are bigger issues we should probably take up elsewhere...

If your sense is that `nan` is a valid input in this case and the output should deterministically be `nan`, instead of indicating a validation problem, then my apologies to @chaitan94 for suggesting otherwise!


As for exceptions, I realize this is very much my own opinion, but my suggestion is that `GEOSException` indicates that something went wrong inside the GEOS function when we otherwise expected it to succeed.  There are a number of places in C that we're returning `ValueError` for invalid inputs (also `TypeError`), and that seems like a reasonable pattern to follow elsewhere: it indicates we deliberately looked for invalid values and handled them before ever calling into GEOS.  The text of the GEOS error in this particular case is sufficiently clear - but I'm not sure we can always count on GEOS to give us back meaningful exceptions for data validation issues.

pygeos/measurement.py
chaitan94: Oh, just to understand better, what's the reasoning behind nan not raising an error?

pygeos/test/test_measurement.py
chaitan94: You are right. Actually, these tests should be over line_string to be more precise, as Fréchet distance works over curves and surfaces, not points. Maybe that is also the reason why we are getting a NaN here.

chaitan94: I have changed all Fréchet distance tests to use line_strings for now.

pygeos/measurement.py
chaitan94: Done.

pygeos/test/test_measurement.py
chaitan94: Done.

pygeos/measurement.py
chaitan94: Let's use the other thread (with @brendan-ward) for any further discussion on this so that we have a single place of reference.

src/ufuncs.c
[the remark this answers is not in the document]
chaitan94: Got it, I made changes to return NaN when densify is passed as NaN.
I have also handled ndarray case of densify and added a test for the same.

Apart from that, I see two options as of now:

1. Handle values <= 0 and >1 ourselves and raise ValueError
2. Just rely on GEOS exception, add tests for that directly

I pushed a commit for the implementation of option 1.

For option 2, we won't have an additional check in the C code, and we will instead do this in the tests:

```py
with pytest.raises(pygeos.GEOSException, match=r"IllegalArgumentException: Fraction is not in range \(0.0 - 1.0]"):
    actual = pygeos.frechet_distance(line_string, line_string, densify=densify)
```

Pros of option 1: We can add more information like for example when I did with `got 2 instead`.
Pros of option 2: No need for additional code and redundant checks in pygeos and it will stay in sync with GEOS going forward as well.

Based on your input I can add another commit to change the implementation to option 2 instead.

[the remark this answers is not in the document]
brendan-ward: @chaitan94 thanks for making the updates!  And apologies for introducing confusion where we still have some different ideas about how to approach validation / error handling.  We'll get there!

👍 on your option 1.  In my personal opinion, as a general rule, it seems risky to rely on the exception messages from GEOS as always fitting our usage here; while the code level dependency on GEOS for implementation is unavoidable, the dependency on exception messages and validation doesn't have to be.

That said, we could go with option 2 as a general rule throughout the codebase, so long as we are diligent about adding tests for those exceptions, and manually verify that the exception we get back from GEOS is good enough (i.e., via the `match` parameter as you've suggested in 2)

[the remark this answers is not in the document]
caspervdw: Sorry for catching on again so late. Thanks for the thorough discussion on this subject already.

> @caspervdw GEOS isn't handling np.nan handled specifically in the C code (i.e., above) - which means it wasn't a choice made inside GEOS, right?

You are correct, sorry, I didn't actually check.

> Personally, I think while there are a number of valid cases where the inputs can be a mix of nan and valid values, it feels like a validation problem for parameters that control the output of the function - and are parameters that are unlikely to vary based on your input geometries.

Using broadcasting is a general design pattern in PyGEOS. We sometimes diverge from it (see the `buffer` implementation), but that's because of technical difficulties on the GOES CAPI side. .

Propagating missing is also a general design pattern in PyGEOS. For Geometries, `None` indicates "missing" , for floats `NaN` does it. For integers and booleans, there is no value to indicate "missing".

The two points distinguished by @chaitan94 do need some more discussion.

> 1. Handle values <= 0 and >1 ourselves and raise ValueError
> 2. Just rely on GEOS exception, add tests for that directly

My preference is: if it is done by GEOS, PyGEOS should not do it again. More code means more maintenance, more tests, more bugs, etc. We need focus.

>  +1 on your option 1. In my personal opinion, as a general rule, it seems risky to rely on the   exception messages from GEOS as always fitting our usage here; while the code level dependency on GEOS for implementation is unavoidable, the dependency on exception messages and validation doesn't have to be.

So that is the exact opposite. I'd like option 2. Discussion is good! Let's continue this in a separate issue. For now, leave it as is, I'll approve so this can be merged.

On a side note, please don't mistake my shortish replies as a lack of appreciation of the work you are doing on this PR. I really appreciate the contributions to PyGEOS.

chaitan94:
Now that the error validation has its own issue (https://github.com/pygeos/pygeos/issues/150) I think this can be merged. If not, please let me know if anything else is needed.

caspervdw:
Done, thanks again!

Issue #155, opened 2020-06-01
ENH: release the GIL in spatial operations (intersection, union, etc) that return geometries


jorisvandenbossche:
This needs an extension to the current `GEOS_INIT_THEADS` / `GEOS_FINISH_THREADS` mechanism that is already used for eg spatial predicates like `intersects` etc (ufuncs that do not return geometries).

For ufuncs returning geometries (eg the `YY_Y` ufuncs), this needs some additional logic, possibly a two-step loop (first loop creating the GEOSGeometry objects, second loop wrapping them in python objects). See also https://github.com/pygeos/pygeos/pull/113#discussion_r393069605 for a brief discussion about this.

jorisvandenbossche:
This was mostly closed by https://github.com/pygeos/pygeos/pull/156. But @caspervdw I think there are some geometry-creating functions remaining to do? (like `points` and some other creation functions)

Further, the `from_wkb` / `from_wkt` are probably hard to do given they deal with Python objects as input (we would need to extract all `char`s in a first loop .. or something like that).

Further one that seems to be missing: `shared_paths` (but this might only be the decorator that is missing)

PR #156, opened 2020-06-01
Release the GIL for geometry-creating operations

caspervdw:
This addresses #155 , now only for the geometry-to-geometry operations (like `envelope` and `convex_hull`)

The same pattern could be repeated for all other geometry-creating operations.

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -260,6 +260,11 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
     FuncGEOS_Y_Y *func = (FuncGEOS_Y_Y *)data;
     GEOSGeometry *in1;
 
+    if ((steps[1] == 0) && (dimensions[0] > 1)) {
+        PyErr_Format(PyExc_NotImplementedError, "Cannot handle ufunc mode with args=[%p, %p], steps=[%ld, %ld], dimensions=[%ld].", args[0], args[1], steps[0], steps[1], dimensions[0]);
+        return;
+    }
+
     // allocate a temporary array to store output GEOSGeometry objects
     GEOSGeometry **geom_arr = malloc(sizeof(void *) * dimensions[0]);
     if (geom_arr == NULL) {
@@ -339,6 +344,11 @@ static void Yd_Y_func(char **args, npy_intp *dimensions,
 {
     FuncGEOS_Yd_Y *func = (FuncGEOS_Yd_Y *)data;
     GEOSGeometry *in1;
+
+    if ((steps[2] == 0) && (dimensions[0] > 1)) {
+        PyErr_Format(PyExc_NotImplementedError, "Cannot handle ufunc mode with args=[%p, %p, %p], steps=[%ld, %ld, %ld], dimensions=[%ld].", args[0], args[1], args[2], steps[0], steps[1], steps[2], dimensions[0]);
+        return;
+    }
 
     // allocate a temporary array to store output GEOSGeometry objects
     GEOSGeometry **geom_arr = malloc(sizeof(void *) * dimensions[0]);
@@ -473,6 +483,11 @@ static void Yi_Y_func(char **args, npy_intp *dimensions,
     FuncGEOS_Yi_Y *func = (FuncGEOS_Yi_Y *)data;
     GEOSGeometry *in1;
 
+    if ((steps[2] == 0) && (dimensions[0] > 1)) {
+        PyErr_Format(PyExc_NotImplementedError, "Cannot handle ufunc mode with args=[%p, %p, %p], steps=[%ld, %ld, %ld], dimensions=[%ld].", args[0], args[1], args[2], steps[0], steps[1], steps[2], dimensions[0]);
+        return;
+    }
+
     // allocate a temporary array to store output GEOSGeometry objects
     GEOSGeometry **geom_arr = malloc(sizeof(void *) * dimensions[0]);
     if (geom_arr == NULL) {
@@ -524,9 +539,73 @@ static void *union_data[1] = {GEOSUnion_r};
 static void *shared_paths_data[1] = {GEOSSharedPaths_r};
 typedef void *FuncGEOS_YY_Y(void *context, void *a, void *b);
 static char YY_Y_dtypes[3] = {NPY_OBJECT, NPY_OBJECT, NPY_OBJECT};
+
+/* There are two inner loop functions for the YY_Y. This is because this
+ * pattern allows for ufunc.reduce.
+ * A reduce operation requires special attention, because we output the
+ * result in a temporary array. NumPy expects that the output array is
+ * filled during the inner loop, so that it can take the first input of
+ * the reduction operation to be the output array. Easiest to show by
+ * example:
+
+ * function called:         out = intersection.reduce(in)
+ * initialization by numpy: out[0] = in[0]; in1 = out; in2[:] = in[:]
+ * first loop:              out[0] = func(out[0], in[1])
+ * second loop:             out[0] = func(out[0], in[2])
+ */
+static void YY_Y_func_reduce(char **args, npy_intp *dimensions,
+                             npy_intp *steps, void *data)
+{
+    FuncGEOS_YY_Y *func = (FuncGEOS_YY_Y *)data;
+    GEOSGeometry *in1, *in2, *out;
+
+    GEOS_INIT_THREADS;
+
+    if (!get_geom(*(GeometryObject **)args[0], &out)) {
+        errstate = PGERR_NOT_A_GEOMETRY;
+    } else if (out != NULL) {
+        BINARY_LOOP {
+            if (!get_geom(*(GeometryObject **)ip2, &in2)) {
+                errstate = PGERR_NOT_A_GEOMETRY;
+                break;
+            }
+            if (in2 == NULL) {
+                // we can break the loop as we know the NULL will propagate
+                break;
+            } else {
+                in1 = out;
+                out = func(ctx, in1, in2);
+                if (i > 0) {
+                    // in1 was intermediate; clean it up
+                    GEOSGeom_destroy_r(ctx, in1);
+                }
+                if (out == NULL) {
+                    errstate = PGERR_GEOS_EXCEPTION;
+                    break;
+                }
+            }
+        }
+    }
+
+    GEOS_FINISH_THREADS;
+
+    // fill the numpy array with a single PyObject while holding the GIL
+    if (errstate == PGERR_SUCCESS) {
+        geom_arr_to_npy(&out, args[2], steps[2], 1);
+    }
+}
+
 static void YY_Y_func(char **args, npy_intp *dimensions,
                       npy_intp *steps, void *data)
 {
+    // A reduce is characterized by the step of the output array being 0:
+    if ((steps[2] == 0) && (args[0] == args[2])) {
+        return YY_Y_func_reduce(args, dimensions, steps, data);
+    } else if ((steps[2] == 0) && (dimensions[0] > 1)) {
+        PyErr_Format(PyExc_NotImplementedError, "Cannot handle ufunc mode with args=[%p, %p, %p], steps=[%ld, %ld, %ld], dimensions=[%ld].", args[0], args[1], args[2], steps[0], steps[1], steps[2], dimensions[0]);
+        return;
+    }
+
     FuncGEOS_YY_Y *func = (FuncGEOS_YY_Y *)data;
     GEOSGeometry *in1, *in2;
 
@@ -540,14 +619,14 @@ static void YY_Y_func(char **args, npy_intp *dimensions,
     GEOS_INIT_THREADS;
 
     BINARY_LOOP {
-        /* get the geometries: return on error */
+        // get the geometries: return on error
         if (!get_geom(*(GeometryObject **)ip1, &in1) || !get_geom(*(GeometryObject **)ip2, &in2)) {
             errstate = PGERR_NOT_A_GEOMETRY;
             destroy_geom_arr(ctx, geom_arr, i - 1);
             break;
         }
         if ((in1 == NULL) | (in2 == NULL)) {
-            /* in case of a missing value: return NULL (None) */
+            // in case of a missing value: return NULL (None)
             geom_arr[i] = NULL;
         } else {
             geom_arr[i] = func(ctx, in1, in2);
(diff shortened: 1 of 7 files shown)


jorisvandenbossche:
Cool. I assume the python wrappers still need to be decorated with `multithreading_enabled` ?

Now, I am trying this out, and I don't see any speedup. I actually don't see usage of multiple threads on my laptop (I was testing with `make_valid`, both with dask as with stdlib `threading`). Which seems to point that there is still somewhere the GIL being reacquired in the loop, but don't directly see where that would be.
(but would need to do a more elaborate profile to be sure where it is spending time)

jorisvandenbossche:
.. ahum, most stupid error ever ;) I forgot to restart my python kernel after building this branch

Now I see a good speedup. For `make_valid` (an expensive operation), I see around 3x speedup by running in parallel on my laptop

Also no noticeable slowdown in the single threaded setting compared to master.

caspervdw:
> .. ahum, most stupid error ever ;) I forgot to restart my python kernel after building this branch
> 
> Now I see a good speedup. For `make_valid` (an expensive operation), I see around 3x speedup by running in parallel on my laptop
> 
> Also no noticeable slowdown in the single threaded setting compared to master.

Good! Thanks for checking :) I'll proceed with the other geometry-returning functions. And indeed, I should add the decorators as well.

caspervdw:
@jorisvandenbossche  I am running into issues with the set operations. If you do this two-step procedure (create GEOSGeometry without GIL, create PyObject with GIL), the reduce functions do not work, as they expect the full PyObject to be created in the ufunc loop.

src/ufuncs.c
```diff
@@ -340,25 +339,47 @@ static void Yd_Y_func(char **args, npy_intp *dimensions,
...
+                if (geom_arr[i] == NULL) {
+                    errstate = PGERR_GEOS_EXCEPTION;
+                    destroy_geom_arr(ctx, geom_arr, i - 1);
+                    break;
+                }
+            }
         }
-        OUTPUT_Y;
+    } else {
+        errstate = PGERR_NO_MALLOC;
```
brendan-ward, line 371: @caspervdw what do you think about moving this up below line 345, with a return immediately after?  Something like:
```
GEOSGeometry **geom_arr = malloc(sizeof(void *) * dimensions[0]);
if (geom_arr == NULL) {
    errstate = PGERR_NO_MALLOC;
    return
}
```

Then the "happy path" becomes simpler:
```
GEOS_INIT_THREADS;
BINARY_LOOP { ... }
GEOS_FINISH_THREADS;

if (errstate == PGERR_SUCCESS) {
    geom_arr_to_npy(geom_arr, args[2], steps[2], dimensions[0]);
}
free(geom_arr);
```

jorisvandenbossche:
>  I am running into issues with the set operations. If you do this two-step procedure (create GEOSGeometry without GIL, create PyObject with GIL), the reduce functions do not work, as they expect the full PyObject to be created in the ufunc loop.

Hmm. So you mean that they expect that at the end of each loop iteration, there is a PyObject that can then be used as starting point for the next iteration?

Basically, a reduce operation looks like (so where the result gets reused as input)

```
r = identity
for i in range(len(A)):
  r = func(r, A[i])
return r
```

and currently, we have the `OUTPUT_Y` inside the loop (which converts the GEOSGeometry pointer to a PyObject and assigns this to `op1`). 
And I assume that numpy then sets the input pointers and strides in such a way that `op1` gets used as input again? (just guessing here, to try to understand what is exactly going on).

So then I was wondering: can't we solve it with adding yet another loop? First a loop unpacking the input objects into an array pointers. Then the loop with the actual operation only dealing with GEOSGeometry pointers, and then afterwards the loop wrapping in python objects. 
But I suppose this actually doesn't help at all since when doing a reduce, you don't have two input arrays that can be unwrapped (do you know what `ip1` and `ip2` point to when it is called from `reduce` ?)

jorisvandenbossche:
> And I assume that numpy then sets the input pointers and strides in such a way that op1 gets used as input again? (just guessing here, to try to understand what is exactly going on).

That indeed seems to be the case:

https://github.com/numpy/numpy/blob/a5d021a1b6f439a19812926bc4d796ef5f346c44/numpy/core/src/umath/ufunc_object.c#L3500-L3508

```
            /* Turn the two items into three for the inner loop */
            dataptrs_copy[0] = dataptrs[0];
            dataptrs_copy[1] = dataptrs[1];
            dataptrs_copy[2] = dataptrs[0];
            strides_copy[0] = strides[0];
            strides_copy[1] = strides[1];
            strides_copy[2] = strides[0];
            innerloop(dataptrs_copy, &count,
                        strides_copy, innerloopdata);
```

src/ufuncs.c
caspervdw: Thanks for the suggestion! The `errstate` variable doesn't exist outside of the `GEOS_INIT` and `GEOS_FINISH`, but I'll just set the python errorstate directly and return instead.

caspervdw:
> > And I assume that numpy then sets the input pointers and strides in such a way that op1 gets used as input again? (just guessing here, to try to understand what is exactly going on).
> 
> That indeed seems to be the case:

That is indeed the case. I added 
```
printf("Args: %d, %d, %d Steps: %d, %d, %d\n", args[0], args[1], args[2], steps[0], steps[1], steps[2]);
```
into the ufunc and I got

```
>>> pygeos.lib.intersection.reduce(arr)
Args: 30428160, 26932968, 30428160 Steps: 0, 8, 0
```

I presume that the output array is filled with the first geometry in the array, else this would never work ;)


> So then I was wondering: can't we solve it with adding yet another loop? First a loop unpacking the input objects into an array pointers. Then the loop with the actual operation only dealing with GEOSGeometry pointers, and then afterwards the loop wrapping in python objects.

I think this is our safest option. This might be tricky to get right though.

caspervdw:
@jorisvandenbossche  I tried 2 approaches, one with the temporary array, and one with some custom logic inside the inner loop. The second one turned out to be much cleaner than the first.

src/ufuncs.c
```diff
@@ -448,30 +539,119 @@ static void *union_data[1] = {GEOSUnion_r};
...
+    if (errstate == PGERR_SUCCESS) {
+        geom_arr_to_npy(&out, args[2], steps[2], 1);
+    }
+}
+
 static void YY_Y_func(char **args, npy_intp *dimensions,
                       npy_intp *steps, void *data)
 {
+    // A reduce is characterized by the step of the output array being 0:
+    if ((steps[2] == 0) && (args[0] == args[2])) {
```
brendan-ward, line 609: Suggestion: convert this to nested conditionals, which more clearly associates the comment with the primary condition:

```suggestion
    if (steps[2] == 0) {
        if (args[0] == args[2]) {...}
        else if (dimensions[0] > 1) {...}
    }
```

src/ufuncs.c
```diff
@@ -448,30 +539,119 @@ static void *union_data[1] = {GEOSUnion_r};
...
+        return YY_Y_func_reduce(args, dimensions, steps, data);
```
brendan-ward, line 610: Suggestion: since `YY_Y_func_reduce` doesn't return a value, split the call to it from the return here to make that more obvious.

```suggestion
        YY_Y_func_reduce(args, dimensions, steps, data);
        return;
```

src/ufuncs.c
```diff
@@ -235,31 +258,52 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
                      npy_intp *steps, void *data)
 {
     FuncGEOS_Y_Y *func = (FuncGEOS_Y_Y *)data;
-    GEOSGeometry *in1, *ret_ptr;
+    GEOSGeometry *in1;
 
-    GEOS_INIT;
+    if ((steps[1] == 0) && (dimensions[0] > 1)) {
```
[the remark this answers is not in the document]
caspervdw: I do not know of any case that this would occur (except for the YY_Y operations). So I am doubting to do this check at all. Still, we might forget something and better to have a Python exception than a segfault or a silent memory leak.

src/ufuncs.c
```diff
@@ -235,31 +258,52 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
                      npy_intp *steps, void *data)
 {
     FuncGEOS_Y_Y *func = (FuncGEOS_Y_Y *)data;
-    GEOSGeometry *in1, *ret_ptr;
+    GEOSGeometry *in1;
```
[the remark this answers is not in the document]
caspervdw: Agree on the `geom_arr` declaration, I am not sure I follow on the NULL initialization. Both `in1` and `geom_arr` are assigned to below. What is the use of the `NULL`? Just good practice?

caspervdw:
Thanks for your review @brendan-ward , I gave especially the reduce function some additional attention.

src/ufuncs.c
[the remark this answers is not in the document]
brendan-ward: I think in general it is a good practice to assign pointers to `NULL` esp. where they are used in a conditional where it is possible they might not have been assigned.  I think the GCC compiler emits warnings about that.

Since there is `in1 == NULL` below, even though it follows assignment by way of `get_geom(*(GeometryObject **)ip1, &in1)`, there is always the chance that it isn't assigned by that function to `NULL` or a valid value by that function.  Of course, in this specific case, if that function wasn't setting it to `NULL` there would be bigger problems (since it is in a loop where some items may be `NULL`).

src/ufuncs.c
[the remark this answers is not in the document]
brendan-ward: Seems OK to leave this in and just note it guards against unexpected behavior.  Always nice to avoid segfaults, though if there is never a case where this would get called, I guess we don't need it...

```
// Fail if inputs do not have the expected structure.
```

src/ufuncs.c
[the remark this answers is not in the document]
caspervdw: We shouldn't rely on a function to initialize the variable, that is true. I added `= NULL` at a couple of places in the code

jorisvandenbossche:
I was initially going to say: we can also simply "punt" on having a no-gil reduce version if it turns out to be too difficult / complex. 
But it seems you already found a relatively clean solution in the meantime! ;) Taking a look now

src/ufuncs.c
```diff
@@ -448,30 +545,129 @@ static void *union_data[1] = {GEOSUnion_r};
...
+            } else {
+                out = func(ctx, in1, in2);
+                if (out == NULL) {
+                    errstate = PGERR_GEOS_EXCEPTION;
+                    break;
+                }
+            }
+        }
+        // We need to cleanup the previous in1, but not if it is owned by python
+        // Get the first in1 from the ufunc input again (misuse in2 for that) and compare
```
jorisvandenbossche, line 600: I don't fully understand this comment. In what kind of cases is in1 owned by python? If it is the original input geometry?

src/ufuncs.c
```diff
@@ -448,30 +545,129 @@ static void *union_data[1] = {GEOSUnion_r};
...
+        geom_arr_to_npy(&out, args[2], steps[2], 1);
+    }
+}
+
 static void YY_Y_func(char **args, npy_intp *dimensions,
                       npy_intp *steps, void *data)
 {
+    // A reduce is characterized by
+
+    if ((steps[2] == 0) && (dimensions[0] > 1)) {
```
jorisvandenbossche, line 620: What is the meaning of this `dimension` actually? Or more specifically, why is it larger than in case of `reduce` ?

src/ufuncs.c
```diff
@@ -235,31 +258,54 @@ static void Y_Y_func(char **args, npy_intp *dimensions,
...
 
-    finish:
-    GEOS_FINISH;
+    GEOS_FINISH_THREADS;
+
+    // fill the numpy array with PyObjects while holding the GIL
+    if (errstate == PGERR_SUCCESS) {
+        geom_arr_to_npy(geom_arr, args[1], steps[1], dimensions[0]);
+    }
+    free(geom_arr);
```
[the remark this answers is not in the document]
caspervdw: That is done inside the loop, whenever an error is set (lines 283 and 296)

caspervdw:
Thanks @jorisvandenbossche, I tried to clarify the docstrings a bit.

Also, this PR is now tested for memory leaks using valgrind (see #159). It passes the tests.

caspervdw:
@jorisvandenbossche and @brendan-ward , I think I addressed the issues. OK to merge?

src/ufuncs.c
```diff
@@ -448,30 +548,132 @@ static void *union_data[1] = {GEOSUnion_r};
...
+                    errstate = PGERR_GEOS_EXCEPTION;
+                    break;
+                }
+            }
+        }
+        // We need to cleanup the intermediate geometry stored in in1.
+        // However, in some cases the intermediate geometry equals the first input, which is
+        // 'owned' by python. Destoying that would lead to a segfault when the python object
+        // is dereferenced.
+        // We check for that situation explicitely from the ufunc input (misusing the variable 'in2').
```
brendan-ward, line 606:
```suggestion
        // We check for that situation explicitly from the ufunc input (misusing the variable 'in2').
```

src/ufuncs.c
```diff
@@ -767,7 +1003,7 @@ static char snap_dtypes[4] = {NPY_OBJECT, NPY_OBJECT, NPY_DOUBLE, NPY_OBJECT};
 static void snap_func(char **args, npy_intp *dimensions,
```
brendan-ward, line 1003: Should this be updated to use `GEOS_INIT_THREADS` / `GEOS_FINISH_THREADS`?  If not, it might be good to add a comment indicating why it is not compatible with multiple threads.

caspervdw: Thanks for spotting that omission, I added it. Also, the creation functions like `points` need attention still, but I'd like to leave that for later.

Care to take a look at the new `snap_func`?

jorisvandenbossche:
Thanks a *lot* @caspervdw !

jorisvandenbossche:
(BTW, I recently experimented with those multithreading capabilities again a bit in dask-geopandas -> https://github.com/jsignell/dask-geopandas/)

caspervdw:
Thanks! @jorisvandenbossche maybe we should have a hangouts session sometime about this dask integration. I have this work project at github.com/nens/dask-geomodeling that also parallelizes geopandas.

jorisvandenbossche:
@caspervdw for some reason I only notice your comment now, but: certainly interested in a hangout chat about dask integration!

PR #157, opened 2020-06-04
ENH: Add more predicates to STRtree query

### LINKED ISSUE #148: ENH: add additional predicates to STRtree query
We can add additional predicates to STRtree query to enable better parity of spatial indexes in geopandas (based on this) and `sf` in R.

Crossref: https://github.com/geopandas/geopandas/pull/1271#issuecomment-632271525

These should be easy to add:
* covers
* covered by
* contains properly

May require a bit more care:
* equals / equals exact (in part due to extra tolerance parameter for equals exact)

Would require more logic to invert the results:
* disjoint  (internally would run intersects, then would return these subtracted from the tree indexes)

brendan-ward:
Resolves #148 for the easy predicates

* covers
* covered by
* contains properly

Pretty much all the work here is in the tests.

### DIFF
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -15,6 +15,9 @@ class BinaryPredicate(IntEnum):
     overlaps = 4
     crosses = 5
     touches = 6
+    covers = 7
+    covered_by = 8
+    contains_properly = 9
 
 
 VALID_PREDICATES = {e.name for e in BinaryPredicate}
@@ -138,7 +141,8 @@ class STRtree:
         geometry : Geometry or array_like
             Input geometries to query the tree.  The envelope of each geometry
             is automatically calculated for querying the tree.
-        predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+        predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses',\
+            'touches', 'covers', 'covered_by', 'contains_properly'}, optional
             The predicate to use for testing geometries from the tree
             that are within the input geometry's envelope.
 
--- a/pygeos/test/test_strtree.py
+++ b/pygeos/test/test_strtree.py
@@ -410,7 +410,7 @@ def test_query_contains_lines(line_tree, geometry, expected):
 @pytest.mark.parametrize(
     "geometry,expected",
     [
-        # point does not contain any polygs (not valid relation)
+        # point does not contain any polygons (not valid relation)
         (pygeos.points(0, 0), []),
         # box overlaps envelope of 2 polygons but contains neither
         (box(0, 0, 1, 1), []),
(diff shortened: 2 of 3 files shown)


pygeos/strtree.py
```diff
@@ -138,7 +141,8 @@ def query_bulk(self, geometry, predicate=None):
         geometry : Geometry or array_like
             Input geometries to query the tree.  The envelope of each geometry
             is automatically calculated for querying the tree.
-        predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+        predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses',\
+            'touches', 'covers', 'covered_by', 'contains_properly'}, optional
```
jorisvandenbossche, line 145:
```suggestion
 'touches', 'covers', 'covered_by', 'contains_properly'}, optional
```

(it's a bit ugly in the code, but that's what is needed to not get empty space in the actual docstring

pygeos/test/test_strtree.py
```diff
@@ -629,6 +629,256 @@ def test_query_touches_polygons(poly_tree, geometry, expected):
...
+def test_query_contains_properly_points(tree, geometry, expected):
+    assert_array_equal(tree.query(geometry, predicate="contains_properly"), expected)
+
+
+@pytest.mark.parametrize(
+    "geometry,expected",
+    [
+        # None of the following conditions satisfy the relation for linestrings
+        # because they have no interior:
+        # "a contains b iff no points of b lie in the exterior of a, and at least one
```
jorisvandenbossche, line 836:
```suggestion
        # "a contains b if no points of b lie in the exterior of a, and at least one
```

brendan-ward: @jorisvandenbossche in this case, it was copied from the definition on the [wikipedia page](https://en.wikipedia.org/wiki/DE-9IM), with a special meaning of [`iff`](https://en.wikipedia.org/wiki/If_and_only_if) but perhaps the mathematical definition isn't as helpful as something that reads nicely in English?

jorisvandenbossche: I had never heard of the abbreviation "iff" for "if and only if" ;) 
So personally I would say it's not that helpful here

brendan-ward:
@jorisvandenbossche this didn't seem like the place to document difference between `contains` and `contains_properly` although it is the first place to introduce `contains_properly` to `pygeos`.

Instead, perhaps we can document that in the docstring of #180 where it would be more proximate, and make reference to it from the `STRtree` docstrings?

Alternatively, we could create a separate issue to add a documentation page to describe each predicate in a bit more detail than is done in their predicate function docstrings now (e.g., similar to what is done in [PostGIS](https://postgis.net/docs/ST_ContainsProperly.html))

PR #159, opened 2020-06-21
Memory leak testing using valgrind

caspervdw:
Valgrind is a C utility to track memory leak issues. See description of how to use the included docker in the README.

Currently, 322 out of 1124 tests are failing, so we have some work to do! I'll go through the logs and make separate issues.

The whole run takes 20 minutes so I did not make it part of the CI.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,19 @@
+# Ignore everything
+**
+
+# Allow files and directories
+!/*.py
+!/*LICENSE*
+!/setup.cfg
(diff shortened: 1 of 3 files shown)


caspervdw:
@jorisvandenbossche I think I'd like to merge this as is. It is a handy tool to use once in a while.

jorisvandenbossche:
Yes, sure! I won't have time on the short term to test out the instruction, so go ahead

PR #168, opened 2020-07-23
FIX: Accept linearrings in pygeos.multilinestrings 

### LINKED ISSUE #153: pygeos.multilinestrings doesn't accept linearrings
`pygeos.multilinestrings` appears to reject input containing `LINEARRING`s. Is that on purpose? This differs from shapely's `MultiLineString()` behaviour.

caspervdw:
Closes #153 

Bug introduced in #86

### DIFF
--- a/pygeos/test/test_creation.py
+++ b/pygeos/test/test_creation.py
@@ -185,10 +185,13 @@ def test_create_collection_skips_none(func, sub_geom):
     [
         (pygeos.multipoints, line_string),
         (pygeos.multipoints, geometry_collection),
+        (pygeos.multipoints, multi_point),
         (pygeos.multilinestrings, point),
         (pygeos.multilinestrings, polygon),
+        (pygeos.multilinestrings, multi_line_string),
         (pygeos.multipolygons, linear_ring),
         (pygeos.multipolygons, multi_point),
+        (pygeos.multipolygons, multi_polygon),
     ],
 )
 def test_create_collection_wrong_geom_type(func, sub_geom):
--- a/pygeos/test/test_creation.py
+++ b/pygeos/test/test_creation.py
@@ -2,7 +2,16 @@ import pygeos
 import pytest
 import numpy as np
 
-from .common import point
+from .common import (
+    point,
+    line_string,
+    linear_ring,
+    polygon,
+    multi_point,
+    multi_line_string,
+    multi_polygon,
+    geometry_collection,
+)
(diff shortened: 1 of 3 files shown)


pygeos/test/test_creation.py
```diff
@@ -121,24 +130,70 @@ def test_2_polygons_with_different_holes():
...
+
+@pytest.mark.parametrize(
+    "func,sub_geom",
+    [
+        (pygeos.multipoints, line_string),
+        (pygeos.multipoints, geometry_collection),
+        (pygeos.multilinestrings, point),
+        (pygeos.multilinestrings, polygon),
+        (pygeos.multipolygons, linear_ring),
+        (pygeos.multipolygons, multi_point),
```
brendan-ward, line 191: In addition to the wrong fundamental geometry type, it would also be good to add check for fail on multi geometries to a multi geometry (here or a different test function):

e.g.,
```
        (pygeos.multipoints, multi_point),
        (pygeos.multilinestrings, multi_line_string),
        (pygeos.multipolygons, multi_polygon),
```

src/ufuncs.c
```diff
@@ -1363,7 +1363,7 @@ static void create_collection_func(char **args, npy_intp *dimensions,
 {
     GEOSGeometry *g, *g_copy;
     int n_geoms, type;
-    char actual_type, expected_type;
+    char actual_type, first_expected_type, last_expected_type;
```
brendan-ward, line 1366: @caspervdw do you foresee cases where testing against a broader range of geometry types will be useful?  As written, testing this as a range between first and last expected type feels awkward (it just so happens that these are in sequence), when there is at most 2 possible expected types.  I think equality would be better in this case.


Since you compare against them below, I'd suggest initializing all to `-1`.

```suggestion
    char actual_type = -1;
    char expected_type = -1;
    char alt_expected_type = -1;
```

Then, I think you could drop the assignments to the same types on lines 1378, 1386 below, and replace 1381-1382 with
```
expected_type = GEOS_LINESTRING;
alt_expected_type = GEOS_LINEARRING;
```

To make the final comparison on 1403 a bit more straightforward:
```
if ((actual_type != expected_type) & (actual_type!= alt_expected_type)) {
```

caspervdw: That's better indeed, thanks for the suggestion.

The `alt_expected_type` has to be set in each iteration though.

PR #171, opened 2020-07-26
[Done] Raise ValueError on to_wkt of multipoint with empty point

### LINKED ISSUE #134: Segfault when creating MultiPoint from sequence including empty point
Triggered by a similar Shapely issue (https://github.com/Toblerity/Shapely/pull/898, https://github.com/Toblerity/Shapely/issues/836), I wondered how this would work in pygeos.

It's not that easy to create an empty point, so using a workaround with intersection:

```
In [1]: import pygeos 

In [8]: p = pygeos.points(1,1) 

In [9]: box = pygeos.box(2,2,4,4)  

In [10]: p_empty = pygeos.intersection(p, box)     

In [11]: p_empty  
Out[11]: <pygeos.Geometry POINT EMPTY>
```

Then trying to create a MultiPoint from a list that includes this empty point segfaults:

```
In [12]: pygeos.multipoints([p, p_empty])  
Out[12]: Segmentation fault (core dumped)
```

Creating one from just an empty point works, resulting in an empty MultiPoint:

```
In [5]: pygeos.multipoints([p_empty]) 
Out[5]: <pygeos.Geometry MULTIPOINT EMPTY>
```

caspervdw:
Closes #134 

It still seems to segfault with GEOS 3.8

### DIFF
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -164,7 +164,7 @@ def test_to_wkt_point_empty():
 
 def test_to_wkt_geometrycollection_with_point_empty():
     collection = pygeos.geometrycollections([empty_point, point])
-    assert pygeos.to_wkt(collection) == "GEOMETRYCOLLECTION Z (POINT EMPTY, POINT (2 3))"
+    assert pygeos.to_wkt(collection) == "GEOMETRYCOLLECTION (POINT EMPTY, POINT (2 3))"
 
 
 def test_to_wkt_multipoint_with_point_empty_errors():
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -158,15 +158,8 @@ def test_to_wkt_exceptions():
         pygeos.to_wkt(point, output_dimension=4)
 
 
-@pytest.mark.skipif(pygeos.geos_version < (3, 7, 0), reason="GEOS < 3.7")
-def test_to_wkt_multipoint_with_point_empty():
-    geom = pygeos.multipoints([point, pygeos.Geometry("POINT EMPTY")])
-    assert pygeos.to_wkt(geom) == "MULTIPOINT Z (1 1, EMPTY)"
-
-
-@pytest.mark.skipif(pygeos.geos_version >= (3, 7, 0), reason="GEOS >= 3.7")
 def test_to_wkt_multipoint_with_point_empty_errors():
-    # Test if segfault is prevented for GEOS < 3.7 
+    # Test if segfault is prevented
     geom = pygeos.multipoints([point, pygeos.Geometry("POINT EMPTY")])
     with pytest.raises(ValueError):
         repr(geom)
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -3,7 +3,7 @@ import pygeos
 import pytest
 from unittest import mock
 
-from .common import all_types, point
+from .common import all_types, point, empty_point
 
 
 POINT11_WKB = (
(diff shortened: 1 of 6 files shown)


caspervdw:
According to https://github.com/libgeos/geos/issues/305, this check is not necessary anymore GEOS 3.7.3, 3.8.2, or 3.9. When these versions are out, we should add version conditionals and test.

src/geos.c
```diff
@@ -16,6 +16,40 @@ int init_geos(PyObject *m)
...
+    geom_type = GEOSGeomTypeId_r(ctx, geom);
+    if (geom_type == -1) { return PGERR_GEOS_EXCEPTION; }
+    if (geom_type != GEOS_MULTIPOINT) { return PGERR_SUCCESS; }
+    
+    n = GEOSGetNumGeometries_r(ctx, geom);
+    if (n == -1) { return PGERR_GEOS_EXCEPTION; }
+    for(i = 0; i < n; i++) {
+        sub_geom = GEOSGetGeometryN_r(ctx, geom, i);
+        if (sub_geom == NULL) { return PGERR_GEOS_EXCEPTION; }
+        is_empty = GEOSisEmpty_r(ctx, sub_geom);
```
brendan-ward, line 45: Not sure if it would be helpful to add as a comment here, but I found the value `2` unexpected until I found this in the GEOS source:
```
Unary predicate - return 2 on exception, 1 on true, 0 on false
```

src/geos.h
```diff
@@ -64,6 +65,8 @@ enum {
         PyErr_SetString(PyExc_MemoryError, "Could not allocate memory"); break;\
       case PGERR_GEOMETRY_TYPE:\
         PyErr_SetString(PyExc_TypeError, "One of the Geometry inputs is of incorrect geometry type."); break;\
+      case PGERR_MULTIPOINT_WITH_POINT_EMPTY:\
+        PyErr_SetString(PyExc_ValueError, "For GEOS < 3.7, WKT output of multipoints with an empty point is unsupported."); break;\
```
brendan-ward, line 69: Error message isn't strictly true, as it fails on 3.8.0, 3.8.1

jorisvandenbossche:
Agreed that we should try to avoid raising an error in the repr (in `to_wkt` that's fine), as that can be very confusing (you easily think that the operation failed, and not the representation of the result).

> Once this no longer segfaults, do we want to allow empty points in multipoints to pass silently? From what I gather, they can't then be read in by GEOS, so if we did, we'd be allowing the user to create something that can't round-trip. It seems like we might want to leave this as a ValueError in to_wkt for all versions?

Personally, I would prefer to have this "fixed" (or a choice made) on the GEOS side, so we don't need to special case this on the long term. I also raised that in the original issue report over there: https://github.com/libgeos/geos/issues/305#issuecomment-616742567 (more comments there might help).

pygeos/test/test_io.py
```diff
@@ -158,6 +158,15 @@ def test_to_wkt_exceptions():
         pygeos.to_wkt(point, output_dimension=4)
 
 
+def test_to_wkt_multipoint_with_point_empty_errors():
+    # Test if segfault is prevented
+    geom = pygeos.multipoints([point, pygeos.Geometry("POINT EMPTY")])
+    with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
caspervdw: That's right, `repr` should not raise. But the geometry is also not invalid, it is just unsupported by GEOS. 

Maybe we can do `<pygeos.Geometry WKT representation unsupported>`

src/pygeom.c
```diff
@@ -79,7 +85,14 @@ static PyObject *GeometryObject_ToWKT(GeometryObject *obj, char *format)
 
 static PyObject *GeometryObject_repr(GeometryObject *self)
 {
-    return GeometryObject_ToWKT(self, "<pygeos.Geometry %s>");
+    PyObject *result = GeometryObject_ToWKT(self, repr_fmt);
+    // we never want a repr() to fail; that can be very confusing
+    if (result == NULL) {
+        PyErr_Clear();
+        return PyUnicode_FromFormat(repr_fmt, "Exception in WKT writer");
+    } else {
```
brendan-ward, line 93: minor suggestion; a bare return would be fine here too, no need for conditions since both branches return:

```
if (result == NULL) { ... }

return result
```

pygeos/test/test_io.py
```diff
@@ -158,6 +158,28 @@ def test_to_wkt_exceptions():
         pygeos.to_wkt(point, output_dimension=4)
 
 
+def test_to_wkt_point_empty():
+    assert pygeos.to_wkt(empty_point) == "POINT EMPTY"
+
+
+def test_to_wkt_geometrycollection_with_point_empty():
+    collection = pygeos.geometrycollections([empty_point, point])
+    assert pygeos.to_wkt(collection) == "GEOMETRYCOLLECTION (POINT EMPTY, POINT (2 3))"
```
brendan-ward, line 167: It looks like this is sensitive to GEOS version (according to the Travis-CI results).

It looks like this is `GEOMETRYCOLLECTION Z (POINT EMPTY, POINT (2 3))` and `GEOMETRYCOLLECTION (POINT EMPTY, POINT (2 3))` depending on the version.  

Perhaps just testing 
```
assert pygeos.to_wkt(collection).endswith('(POINT EMPTY, POINT (2 3))')
```

is sufficient to capture the substantive part of this test, and sidesteps version-specific issues?

Issue #154, opened 2020-06-01
ENH: release the GIL in STRtree query/bulk_query


jorisvandenbossche:
Since we now have the machinery to release the GIL in certain places of the code, introduced in https://github.com/pygeos/pygeos/pull/113, we can use this to release the GIL in the query functions of the STRtree.

(this will be useful for making a parallel sjoin in geopandas-dask)

PR #174, opened 2020-08-05
Release the GIL for STRtree bulk_query

jorisvandenbossche:
Towards https://github.com/pygeos/pygeos/issues/154

Starting with `bulk_query`, as this is the most important for me (for testing a parallel spatial join), and also turned out to be the easiest: I get compilation errors if you do `GEOS_FINISH_THREADS` in a branch, and the `query` function has basically two modes (with predicate checking or without), so would need to be rewritten or split a bit to avoid needing to FINISH_GEOS in two places.

### DIFF
--- a/src/strtree.c
+++ b/src/strtree.c
@@ -412,15 +412,14 @@ static PyObject *STRtree_query_bulk(STRtreeObject *self, PyObject *args) {
     kv_init(src_indexes);
     kv_init(target_indexes);
 
-    GEOS_INIT;
+    GEOS_INIT_THREADS;
 
     for(i = 0; i < n; i++) {
         // get pygeos geometry from input geometry array
         pg_geom = *(GeometryObject **) PyArray_GETPTR1(pg_geoms, i);
         if (!get_geom(pg_geom, &geom)) {
             errstate = PGERR_NOT_A_GEOMETRY;
-            GEOS_FINISH;
-            return NULL;
+            break;
         }
         if (geom == NULL || GEOSisEmpty_r(ctx, geom)) {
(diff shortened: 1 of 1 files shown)


brendan-ward:
@jorisvandenbossche what about using the `goto finish;` pattern used in `ufuncs.c` for multiple return paths, followed by a bit of checking to see if there is anything in the internal vector to copy out to numpy arrays?

the internals of query with / without predicate could be split into two functions in C, or could be split at the top-level C API into two functions (keep Python API as is though).

jorisvandenbossche:
> what about using the goto finish; pattern used in ufuncs.c for multiple return paths

This suggestion is for `STRtree_query` I suppose (and not the current diff)? 
That's certainly an option. We would need to check again in which of both cases we are (as a different vector needs to be copied into the result), so not sure it will in the end be cleaner as splitting the core part in two functions.

In any case, doing that for `query` can also be left for a separate PR, and can just focus on `query_bulk` here.

brendan-ward:
Yes, suggestion was in response to your comments about `query` - it wasn't immediately clear if you were planning on doing that in this PR or a follow up.  Handling that in a follow up PR seems like a good idea, will involve more changes than here.

caspervdw (approved):
To be safe, I ran the valgrind memory leak tester on the strtree tests, and they passed.

So this is good to merge. Just an addition to the changelog?

jorisvandenbossche:
> To be safe, I ran the valgrind memory leak tester on the strtree tests, and they passed.

Thanks for testing!

Added it to the changelog.

PR #176, opened 2020-08-11
Rename get_coordinate_dimension (dimensions -> dimension)

jorisvandenbossche:
I just noticed a small inconsistency with the GEOS API. In GEOS it is `GEOSGeom_getCoordinateDimension` (without trailing "s"), while in pygeos we have `get_coordinate_dimensions` (plural). 
Since our names are mostly based on GEOS' names, I *suppose* this inconsistency was an oversight (unless @caspervdw there is a reason for it?)

BTW, in PostGIS they abbreviate this to [`ST_CoordDim`](https://postgis.net/docs/ST_CoordDim.html) (and have an additional alias [`ST_NDims`](https://postgis.net/docs/ST_NDims.html)), but I think writing it in full is fine for us.


caspervdw:
Yes, singular is indeed better than the plural version. This is OK to merge, could you take note of this in the changelog?

Issue #131, opened 2020-03-26
Support get/set_coordinates/apply with more than 2 dimensions


jorisvandenbossche:
Currently, the `apply` function (based on the `get_coordinates` / `set_coordinates` functions), explicitly states it is only for 2 dimensions and others will be dropped:

https://github.com/pygeos/pygeos/blob/b176e297c673671cd78f82738f1a899310501dd9/pygeos/coordinates.py#L7-L12

However, for using this function for coordinate transformations in GeoPandas, we need to be able to work with more than 2 dimensions (at least right now with Shapely, that is supported). 

@caspervdw do you remember if there was a fundamental reason this could not be done, or rather to keep the initial PR simpler?
 
I suppose the main problem is that there is no guarantee that all geometries in the array will have the same number of dimensions. 
However, I suppose there could be an option to specify how many dimensions should be used, and then in `get_coordinates`, those geometries that have less dimensions as the specified dimension can be filled with 0's or nan's for those missing dimensions?

caspervdw:
There was certainly no fundamental reason that this could not be done, just a lot of work to get it right.

A preferable API would be `ndim=2` or `ndim=3`, determining how many 1D coordinate arrays are returned.

PR #178, opened 2020-08-14
ENH: ability to get z-coordinates in get_coordinates

jorisvandenbossche:
I did this now by adding an `include_z=True/False` keyword. This means that the user needs to choose whether to get z coordinates for the full array or not (regardless whether there are geometries with z or not). 

I think this is the cleanest way to do this at this level (it can then be up to the user to detect if there are geometries with z values or not, to set the keyword as desired).

Related to https://github.com/pygeos/pygeos/issues/131, but only did the "get_coordinates" part (not yet set_coordinates for a full `apply` roundtrip).

### DIFF
--- a/pygeos/coordinates.py
+++ b/pygeos/coordinates.py
@@ -30,7 +30,7 @@ def apply(geometry, transformation):
     [<pygeos.Geometry POINT (0 0)>, None]
     """
     geometry_arr = np.array(geometry, dtype=np.object)  # makes a copy
-    coordinates = lib.get_coordinates(geometry_arr)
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
@caspervdw I notice now that you propose a `ndim=2/3` keyword in the issue (https://github.com/pygeos/pygeos/issues/131), that would be fine for me as well, open to keyword name changes.

caspervdw (approved):
This is a great extension. I think we can go with your `include_z=True/False`. Thanks for doing this.

**edit**: Could you append the changelog in this PR?

PR #179, opened 2020-08-20
[DONE] Fix POINT EMPTY to_wkb

### LINKED ISSUE #60: Fix the ToWKB for empty points

caspervdw:
This closes #60. I am still looking at a segfault, but I wanted to share the code already. If someone can spot the mistake, I would be grateful.


jorisvandenbossche:
Is the segfault already solved?

caspervdw:
Yes, I fixed the segfault. Remaining issues are now:

 - Handle EMPTY POINTS in multipoints and geometrycollections (maybe useful to wait for the merge of #171)
 - Figure out how to create 2D Empty points. `GEOSGeom_createEmptyPoint_r` creates a 3D version.

caspervdw:
This should do it. I sidestepped the above two issues:

 - GEOS only supports POINT Z EMPTY, not POINT EMPTY.
 - WKB serializations of multipoints / geometrycollections with empty points still does not work.

mwtoews:
> * GEOS only supports POINT Z EMPTY, not POINT EMPTY.

I consider this a bug, see [geos trac#1048](https://trac.osgeo.org/geos/ticket/1048). Hopefully this will be addressed before 3.9.0 gets finalised.

pygeos/test/test_io.py
```diff
@@ -248,6 +245,51 @@ def test_to_wkb_srid():
...
 
+def test_to_wkb_point_empty():
+    # empty point converts to POINT (nan, nan, nan) in WKB
+    # this matches PostGIS behaviour
+    assert pygeos.to_wkb(empty_point) == POINT_NAN_WKB
+
+
+def test_from_wkb_point_nan():
+    # WKB representation of POINT (nan, nan, nan) converts to POINT EMPTY
+    geom = pygeos.from_wkb(POINT_NAN_WKB)
```
jorisvandenbossche, line 256: In the reading case, should we test that both variations of the POINT_NAN_WKB (with 2 or 3 coordinates) are read correctly? (because now this test only tests one of both depending on the GEOS version)

pygeos/test/test_io.py
```diff
@@ -248,6 +245,51 @@ def test_to_wkb_srid():
...
+    wkb = pygeos.to_wkb(expected, include_srid=True)
+    actual = pygeos.from_wkb(wkb)
+    assert pygeos.equals_exact(actual, expected)
+
+
+@pytest.mark.parametrize("geom", [
+    pygeos.multipoints([empty_point]),
+    pygeos.geometrycollections([empty_point]),
+    pygeos.geometrycollections([pygeos.multipoints([empty_point])]),
+    pygeos.geometrycollections([pygeos.geometrycollections([empty_point])])
```
jorisvandenbossche, line 285: this is the same parametrization as the test above, but this one checks that conversion to wkb fails, while the above tests that the roundtrip works ?

src/ufuncs.c
```diff
@@ -1711,10 +1720,41 @@ static void to_wkb_func(char **args, npy_intp *dimensions,
             Py_INCREF(Py_None);
             *out = Py_None;
         } else {
+            /* WKB Does not allow empty points.
+             * We check for that and patch the case of a 'simple' POINT EMPTY.
+             * Collections with POINT EMPTY in it will raise an exception
```
jorisvandenbossche, line 1725: is this comment still up to date? Because you now also handle MultiPoint / GeometryCollection with empty points?

jorisvandenbossche (commented):
Thanks a lot for this! Code generally looks good to me.

Added a few small comments. It is adding a lot of special case code, unfortunately (we should really try to get this into GEOS, so we can remove it at some point), but nothing to do about that I suppose if we want to get consistent behaviour. Do you know if PostGIS also handles the multipoint/geometrycollection cases?

I did a quick (small) performance test on WKB roundtrip of normal points, and I didn't notice any problem (I see a small slowdown in `to_wkb`, but it's only minor, a few percent)

caspervdw:
@jorisvandenbossche  Thanks for the review. It turned out to be a lot of work to get it finished, but I think I finally have it now.

Looking back I think we should have waited for an upstream fix. This wasn't worth it ;) But now it is finished, I like to have it in anyway.

jorisvandenbossche:
You reverted the changes in `from_wkb`? Is this because GEOS already makes the conversion from NaN to empty itself?

> Looking back I think we should have waited for an upstream fix. This wasn't worth it ;) But now it is finished, I like to have it in anyway.

That sounds as a good summary .. ;-)

jorisvandenbossche:
Just out of curiosity, can you answer:

> You reverted the changes in from_wkb? Is this because GEOS already makes the conversion from NaN to empty itself?

caspervdw:
> Just out of curiosity, can you answer:
> 
> > You reverted the changes in from_wkb? Is this because GEOS already makes the conversion from NaN to empty itself?

Yes it does! I realized it only when reading https://trac.osgeo.org/geos/ticket/1005

```
>>> pygeos.from_wkb(pygeos.to_wkb(pygeos.Geometry("POINT EMPTY")))                                                                                                                                   
<pygeos.Geometry POINT EMPTY>
>>> pygeos.from_wkb(pygeos.to_wkb(pygeos.Geometry("MULTIPOINT (1 1, nan nan)")))                                                                                         
<pygeos.Geometry Exception in WKT writer>
>>> pygeos.from_wkb(pygeos.to_wkb(pygeos.Geometry("GEOMETRYCOLLECTION (POINT EMPTY, LINESTRING (1 1, 2 2))")))                                                          
<pygeos.Geometry GEOMETRYCOLLECTION Z (POINT EMPTY, LINESTRING (1 1, 2 2))
```

Note that MULTIPOINT (1 1, EMPTY) is not supported by GEOS.
Also there are still odd things going on with the dimensionality, but that should be fixed upstream geos trac#1048

jorisvandenbossche:
Thanks for the explanation!

> Note that MULTIPOINT (1 1, EMPTY) is not supported by GEOS.

Do you know if there is already a trac issue about that?

PR #182, opened 2020-08-21
ENH: Add subclass registry (enable subclassing pygeos.Geometry)

jorisvandenbossche:
@caspervdw I rebased your commits for the subclass registry on latest master, and moved the Point subclass to the tests.

Would it be fine for you (after review of course) to merge this, even without actually using it in PyGEOS? That will make it easier to use in my Shapely refactor (xref https://github.com/Toblerity/Shapely/issues/962)

(I am using this branch right now when working on my shapely branch that makes all shapely's geometry classes subclasses of pygeos.Geometry)


jorisvandenbossche:
> the implementation here adds an additonal GEOS_INIT and GEOS_FINISH to the function.

Good point! 

I repeated the benchmark that I did in https://github.com/pygeos/pygeos/pull/104#issuecomment-609039745, and indeed this actually gives a bit more overhead. For creating 1 million points and with a tuned system (giving lower performance in general), I have ca 600ms for master vs 900ms for this branch, so giving a 50% penalty. 
But applying your suggestion of not initializing GEOS each time in `GeometryObject_FromGEOS`, the overhead goes down again (ca 700ms, less than 20% slowdown, around 100ns per geometry), and then it is similar again to what we saw before.

pygeos/test/test_creation.py
```diff
@@ -208,3 +208,31 @@ def test_box_multiple():
...
+def test_subclasses():
+    orig = pygeos.lib.registry[0]
+    pygeos.lib.registry[0] = Point
+
+    for point in [Point("POINT (1 1)"), pygeos.points(1, 1)]:
+        assert isinstance(point, Point)
+        assert pygeos.get_type_id(point) == pygeos.GeometryType.POINT
+        assert point.x == 1
+
+    pygeos.lib.registry[0] = orig
```
caspervdw, line 238: This teardown should also happen if the test fails. Otherwise we may get confusing situations in the future.

Maybe with a fixture:

```
@pytest.fixture
def with_point_in_registry():
    pygeos.lib.registry[0] = Point
    yield
    pygeos.lib.registry[0] = orig
```

caspervdw:
Thanks for diving into this! I think this is good to merge, only one minor comment in the unittest.

pygeos/test/test_creation.py
jorisvandenbossche: Used that fixture!

jorisvandenbossche:
Do you think a changelog entry is needed? It's not really a user-facing feature that users should start to play with ..

PR #189, opened 2020-09-05
Limit WKT length in repr

### LINKED ISSUE #124: Limit length of WKT in repr
If you have big geometries, checking an array of those quickly completely fills the terminal with its repr. For printing purposes, I think it would be nice to limit the length of the repr (truncate the string after a certain amaount of characters). The `to_wkt` method is still available to see the full representation.

caspervdw:
Closes #124

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -110,7 +110,7 @@ def buffer(
     >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=1)
     <pygeos.Geometry POLYGON ((12 10, 10 8, 8 10, 10 12, 12 10))>
     >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=2))  # doctest: +SKIP
-    POLYGON ((12 10, 11.4 8.59, 10 8, 8.59 8.59, 8 10, 8.59 11.4, 10 12, 11.4 11.4, 12 10))
+    <pygeos.Geometry POLYGON ((12 10, 11.4 8.59, 10 8, 8.59 8.59, 8 10, 8.59 11.4, 10 12, 11.4 11.4, 12 10))>
     >>> buffer(Geometry("POINT (10 10)"), -2, quadsegs=1)
     <pygeos.Geometry POLYGON EMPTY>
     >>> line = Geometry("LINESTRING (10 10, 20 10)")
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -109,8 +109,8 @@ def buffer(
     --------
     >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=1)
     <pygeos.Geometry POLYGON ((12 10, 10 8, 8 10, 10 12, 12 10))>
-    >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=2)
-    <pygeos.Geometry POLYGON ((12 10, 11.4 8.59, 10 8, 8.59 8.59, 8 10, 8.59 11.4, 10 12, 11.4 11.4, 12 10))>
+    >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=2))  # doctest: +SKIP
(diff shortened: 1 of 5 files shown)


brendan-ward (approved):
Other than removing unused comments, this looks good and is a welcome change.  Thanks @caspervdw !

caspervdw:
Is 256 characters not a bit too long? Typical line length is 80 or 120 characters. I can't think of a reason why someone would like to see more. What do you think @brendan-ward @jorisvandenbossche

jorisvandenbossche:
I think limiting it to a single line in a typical terminal would be a good idea. Since we can't infer the terminal width (or at least don't want to ;-)), maybe 80 is the safest?

brendan-ward:
80 chars is fine.  As a point of reference, it looks like pandas default text column width is 50 chars.  However, here it would be nice to give a little bit more space to handle the `<pygeos.Geometry ` prefix.

I wonder if it would be good to give more of a hint that it is truncated, similar to pandas?

E.g., 
```
<pygeos.Geometry LINESTRING (0 0, 1 1, 2 2, 3 3, 4 4, 5 5, 6 6, 7 7, 8 8, 9 9, 10 10, 11 11, 12 1...>
```

jorisvandenbossche:
>  it looks like pandas default text column width is 50 chars. 

Yes, but in a DataFrame you typically have multiple columns, so I don't think we need to look at that.

> I wonder if it would be good to give more of a hint that it is truncated, similar to pandas?

Good point, I think we should indeed include a `...` if it is truncated.

caspervdw:
Adding the ellipsis `...` did improve the output, but the implementation is now less elegant.

jorisvandenbossche:
> but the implementation is now less elegant.

but worth it I think! ;)

Now, I just realize something: in the Shapely subclasses, we might want to customize the repr (like showing `shapely.geometry.Point` or `shapely.Point` instead of `shapely.Geometry` in case of points, etc). *If* we do that, it will be easier to do in Python, and then the repr defined on the base Geometry extension type will basically never be seen ..

caspervdw:
We don't know if and when that will happen. I don't think we can do anything about it here in this PR.

@brendan-ward  Do you want to have another look or is it good to merge?

pygeos/constructive.py
```diff
@@ -109,7 +109,7 @@ def buffer(
     --------
     >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=1)
     <pygeos.Geometry POLYGON ((12 10, 10 8, 8 10, 10 12, 12 10))>
-    >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=2)
+    >>> buffer(Geometry("POINT (10 10)"), 2, quadsegs=2))  # doctest: +SKIP
```
brendan-ward, line 112: It might be OK to leave these either by wrapping with `to_wkt()` to print full string (though it does distract a tiny bit from the example)
or include the truncated version of the output, which may be sufficient to demonstrate that it producted a polygon buffer:
```
<pygeos.Geometry POLYGON ((12 10, 11.4 8.59, 10 8, 8.59 8.59, 8 10, 8.59 11....>
```

brendan-ward (commented):
Thanks for shortening and adding ellipses.

My preference would be include functional doctests instead of skipping those with longer outputs; that way we're not making exceptions when things are truncated (could result in examples that are not valid).  I think we can wrap in `to_wkt` when we want the example output to show the full WKT (if important to demonstrate the example) or include truncated WKT otherwise.

pygeos/constructive.py
jorisvandenbossche: I would also be fine with actually having the truncated one in the docstrings below (it also makes the docstring line length shorter)

jorisvandenbossche:
> We don't know if and when that will happen. 

I am actually doing it in https://github.com/Toblerity/Shapely/pull/983 (for Point only)

jorisvandenbossche:
Thanks @caspervdw !

Issue #165, opened 2020-07-02
ENH: Consolidate repeated snippets used in ufuncs.c


brendan-ward:
There are a few repeated snippets introduced in #156 that can be refactored into functions.

caspervdw:
@brendan-ward I like the suggestion of the refactor, but I myself don't see enough benefit to actually do it.

Do you think you can make time for this?

brendan-ward:
I think this is still a good idea over the long term, so I think we should keep this open.  In the short term, I don't have time to work on this.

PR #190, opened 2020-09-05
Support pickling

### LINKED ISSUE #46: Support to pickle Geometry objects
I think this will involve defining `__reduce__` or `__setstate__` / `__getstate__` or some of the related methods that are involved in pickling.

With @pierreglaser, we did some work on this at EuroScipy (while actually trying to do the dtype), and there is some content to reuse in https://github.com/pierreglaser/pygeos/commit/0e0ea8e0e0ce28b34502cba37382121b1ee064b9

caspervdw:
Closes #46

After a failed approach with `__getstate__` and `__setstate__`, I chose for the `__reduce__` route. `__setstate__` does not play  nice with the fact that our geometry objects are immutable. Further reading: https://docs.python.org/3/library/pickle.html#pickling-class-instances

`__reduce__` should return a tuple (callable, args). A convenient way for us would be the following:
- `(pygeos.lib.Geometry, ("<WKT>", ))`

However, WKT serialization is less efficient than WKB serialization. We could opt for:

- `(pygeos.lib.from_wkb, (b"<WKB>", ))`

When testing these approaches, a noticed that both of them have drawbacks.

- WKT does not support SRID serialization and it has an issue with empty points in multipoints (#171).
- WKB does not support linearrings

I am still in favour of going the WKB route. Do we accept that linearrings gets transformed into linestrings when pickled-unpickled?

```
>>> import pickle, pygeos
>>> g = pygeos.linearrings([[(0, 0), (0, 1), (1, 1), (0, 0)]])
>>> pickle.loads(pickle.dumps(g))
array([<pygeos.Geometry LINESTRING (0 0, 0 1, 1 1, 0 0)>], dtype=object)
```

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -35,7 +35,8 @@ Geometries can be serialized using pickle:
   >>> pickle.loads(point_1)
   <pygeos.Geometry POINT (5.2 52.1)>
 
-.. warning:: Pickling is not supported for linearrings and empty points. See :func:`pygeos.io.to_wkb` for a list of limitations.
+.. warning:: Pickling is not supported for linearrings and empty points.
+             See :func:`pygeos.io.to_wkb` for a complete list of limitations.
 
 Hashing
 ~~~~~~~
@@ -51,7 +52,8 @@ Therefore, geometries are equal if and only if their WKB representations are equ
   >>> {point_1, point_2, point_3}
   {<pygeos.Geometry POINT (5.2 52.1)>, <pygeos.Geometry POINT (1 1)>}
 
-.. warning:: Due to limitations of WKB, linearrings will equal linearrings if they contain the exact same points. Also, different kind of empty points will be considered equal. See :func:`pygeos.io.to_wkb` for a complete list.
+.. warning:: Due to limitations of WKB, linearrings will equal linearrings if they contain the exact same points.
+             Also, different kind of empty points will be considered equal. See :func:`pygeos.io.to_wkb`.
 
 Comparing two geometries directly is also supported.
 This is the same as using :func:`pygeos.predicates.equals_exact` with a ``tolerance`` value of zero.
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -8,7 +8,7 @@ lets the python garbage collector free its memory when it is not
 used anymore.
 
 Geometry objects are immutable. This means that after constructed, they cannot
-be changed inplace. Every PyGEOS operation will result in a new object being returned.
+be changed in place. Every PyGEOS operation will result in a new object being returned.
 
 Construction
 ~~~~~~~~~~~~
@@ -35,15 +35,7 @@ Geometries can be serialized using pickle:
   >>> pickle.loads(point_1)
   <pygeos.Geometry POINT (5.2 52.1)>
 
-.. warning:: Pickling is not supported for linearrings and empty points.
-
-In pickled form, a geometry is represented in its WKB (Well-Known Binary) representation.
-Because of this, the following limitations apply:
-
-- linearrings will be converted to linestrings
-- a point with only NaN coordinates is converted to an empty point
-- empty points are transformed to 3D in GEOS < 3.8
-- empty points are transformed to 2D in GEOS 3.8
+.. warning:: Pickling is not supported for linearrings and empty points. See :func:`pygeos.io.to_wkb` for a list of limitations.
 
 Hashing
 ~~~~~~~
@@ -58,6 +50,8 @@ Therefore, geometries are equal if and only if their WKB representations are equ
   >>> point_3 = Geometry("POINT (5.2 52.1)")
   >>> {point_1, point_2, point_3}
   {<pygeos.Geometry POINT (5.2 52.1)>, <pygeos.Geometry POINT (1 1)>}
+
+.. warning:: Due to limitations of WKB, linearrings will equal linearrings if they contain the exact same points. Also, different kind of empty points will be considered equal. See :func:`pygeos.io.to_wkb` for a complete list.
 
 Comparing two geometries directly is also supported.
 This is the same as using :func:`pygeos.predicates.equals_exact` with a ``tolerance`` value of zero.
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -1,5 +1,78 @@
-Geometry properties
-===================
+Geometry
+========
+
+The `pygeos.Geometry` class is the central datatype in PyGEOS. 
+An instance of `pygeos.Geometry` is a container of the actual GEOSGeometry object.
+The Geometry object keeps track of the underlying GEOSGeometry and
+lets the python garbage collector free its memory when it is not
+used anymore.
+
+Geometry objects are immutable. This means that after constructed, they cannot
+be changed inplace. Every PyGEOS operation will result in a new object being returned.
+
+Construction
+~~~~~~~~~~~~
+
+For convenience, the ``Geometry`` class can be constructed with a WKT (Well-Known Text)
+or WKB (Well-Known Binary) representation of your geometry:
+
+.. code:: python
+
+  >>> from pygeos import Geometry
+  >>> point_1 = Geometry("POINT (5.2 52.1)")
+  >>> point_2 = Geometry(b"\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?")
+
+A more efficient way of constructing geometries is by making use of the (vectorized)
+functions described in :mod:`pygeos.creation`.
+
+Pickling
+~~~~~~~~
+
+Geometries can be serialized using pickle:
+
+  >>> import pickle
+  >>> pickled = pickle.dumps(point_1)
+  >>> pickle.loads(point_1)
+  <pygeos.Geometry POINT (5.2 52.1)>
+
+.. warning:: Pickling is not supported for linearrings and empty points.
+
+In pickled form, a geometry is represented in its WKB (Well-Known Binary) representation.
+Because of this, the following limitations apply:
+
+- linearrings will be converted to linestrings
+- a point with only NaN coordinates is converted to an empty point
+- empty points are transformed to 3D in GEOS < 3.8
+- empty points are transformed to 2D in GEOS 3.8
+
+Hashing
+~~~~~~~
+
+Geometries can be used as elements in sets or as keys in dictionaries.
+Python uses a technique called *hashing* for lookups in these datastructures.
+PyGEOS generates this hash from the WKB representation.
+Therefore, geometries are equal if and only if their WKB representations are equal.
+
+.. code:: python
+
+  >>> point_3 = Geometry("POINT (5.2 52.1)")
+  >>> {point_1, point_2, point_3}
+  {<pygeos.Geometry POINT (5.2 52.1)>, <pygeos.Geometry POINT (1 1)>}
+
+Comparing two geometries directly is also supported.
+This is the same as using :func:`pygeos.predicates.equals_exact` with a ``tolerance`` value of zero.
+
+  >>> point_1 == point_2
+  False
+  >>> point_1 != point_2
+  True
+
+
+Properties
+~~~~~~~~~~
+
+Geometry objects have no attributes or properties.
+Instead, use functions listed below to obtain information about geometry objects.
 
 .. automodule:: pygeos.geometry
    :members:
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -5,7 +5,7 @@ import pickle
 import struct
 from unittest import mock
 
-from .common import all_types, point, empty_point
+from .common import all_types, point, empty_point, point_z
 
 
 POINT11_WKB = b'\x01\x01\x00\x00\x00' + struct.pack("<2d", 1., 1.)
@@ -352,9 +352,18 @@ def test_from_shapely_incompatible_versions():
         pygeos.from_shapely(point)
 
 
-@pytest.mark.parametrize("geom", all_types)
+@pytest.mark.parametrize("geom", all_types + (point_z, empty_point))
 def test_pickle(geom):
     if pygeos.get_type_id(geom) == 2:
-        pytest.xfail("Linearrings are converted to linestrings in pickle")
+        # Linearrings get converted to linestrings
+        expected = pygeos.linestrings(pygeos.get_coordinates(geom))
+    else:
+        expected = geom
     pickled = pickle.dumps(geom)
-    assert pygeos.equals_exact(pickle.loads(pickled), geom)
+    assert pygeos.equals_exact(pickle.loads(pickled), expected)
+
+
+def test_pickle_with_srid():
+    geom = pygeos.set_srid(point, 4326)
+    pickled = pickle.dumps(geom)
+    assert pygeos.get_srid(pickle.loads(pickled)) == 4326
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -354,5 +354,7 @@ def test_from_shapely_incompatible_versions():
 
 @pytest.mark.parametrize("geom", all_types)
 def test_pickle(geom):
+    if pygeos.get_type_id(geom) == 2:
+        pytest.xfail("Linearrings are converted to linestrings in pickle")
     pickled = pickle.dumps(geom)
     assert pygeos.equals_exact(pickle.loads(pickled), geom)
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pygeos
 import pytest
+import pickle
 import struct
 from unittest import mock
 
@@ -351,7 +352,7 @@ def test_from_shapely_incompatible_versions():
         pygeos.from_shapely(point)
 
 
-import pickle
-def test_pickle_geometry_class():
-    pickled = pickle.dumps(pygeos.Geometry)
-    assert pickle.loads(pickled) is pygeos.Geometry
+@pytest.mark.parametrize("geom", all_types)
+def test_pickle(geom):
+    pickled = pickle.dumps(geom)
+    assert pygeos.equals_exact(pickle.loads(pickled), geom)
--- a/pygeos/test/test_io.py
+++ b/pygeos/test/test_io.py
@@ -349,3 +349,9 @@ def test_from_shapely_error(geom):
 def test_from_shapely_incompatible_versions():
     with pytest.raises(ImportError):
         pygeos.from_shapely(point)
+
+
+import pickle
+def test_pickle_geometry_class():
+    pickled = pickle.dumps(pygeos.Geometry)
+    assert pickle.loads(pickled) is pygeos.Geometry
--- a/src/pygeom.c
+++ b/src/pygeom.c
@@ -117,7 +117,7 @@ static PyObject *GeometryObject_ToWKB(GeometryObject *obj)
     // We check for that and patch the POINT EMPTY if necessary
     has_empty = has_point_empty(ctx, obj->ptr);
     if (has_empty == 2) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
-    if (has_empty) {
+    if (has_empty == 1) {
         geom = point_empty_to_nan_all_geoms(ctx, obj->ptr);
     } else {
         geom = obj->ptr;
@@ -148,7 +148,9 @@ static PyObject *GeometryObject_ToWKB(GeometryObject *obj)
     if (wkb != NULL) {
         GEOSFree_r(ctx, wkb);
     }
+
     GEOS_FINISH;
+
     return result;
 }
 
@@ -281,25 +283,24 @@ static PyObject *GeometryObject_FromWKB(PyObject *value)
     unsigned char *wkb;
     Py_ssize_t size;
     GEOSGeometry *geom;
-    GEOSWKBReader *reader;
+    GEOSWKBReader *reader = NULL;
 
     /* Cast the PyObject bytes to char* */
-    if (PyBytes_Check(value)) {
-        size = PyBytes_Size(value);
-        wkb = (unsigned char *)PyBytes_AsString(value);
-        if (wkb == NULL) { return NULL; }
-    } else {
+    if (!PyBytes_Check(value)) {
         PyErr_Format(PyExc_TypeError, "Expected bytes, found %s", value->ob_type->tp_name);
         return NULL;
     }
+    size = PyBytes_Size(value);
+    wkb = (unsigned char *)PyBytes_AsString(value);
+    if (wkb == NULL) { return NULL; }
 
     GEOS_INIT;
 
     reader = GEOSWKBReader_create_r(ctx);
     if (reader == NULL) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
     geom = GEOSWKBReader_read_r(ctx, reader, wkb, size);
-    GEOSWKBReader_destroy_r(ctx, reader);
     if (geom == NULL) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+
     result = GeometryObject_FromGEOS(geom, ctx);
     if (result == NULL) {
         GEOSGeom_destroy_r(ctx, geom);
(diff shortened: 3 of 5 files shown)


caspervdw:
Some further exploration on this: going through WKB works, only there is one issue: LINEARRINGS cannot be represented in WKB. So only the WKB is not a sufficient for pickling. Ref. https://github.com/Toblerity/Shapely/issues/547

brendan-ward:
WKB seems reasonable, given that other similar efforts (e.g., feather / parquet support in geopandas) use WKB as the serializable representation.

linearrings only really make sense in the context of polygons, right?  Thus I think of linearrings as an input toward making polygons or extracted from polygons and used as a linestring, but not something that should be serialized as a discrete entity.  I think noting this as a known limitation (imposed by WKB spec) is sufficient, and perhaps a one-liner for getting back to a linearring if absolutely necessary?

```
ring = pg.get_exterior_ring(pg.Geometry("POLYGON ((1 0, 1 1, 0 1, 0 0, 1 0))"))
pg.linearrings(pg.get_coordinates(pickle.loads(pickle.dumps(ring))))
```

brendan-ward:
At first glance, it looks like this adds a sizeable amount of complexity with some potential overlaps with the from / to WKB functions in ufuncs.  Would there be a benefit to trying to refactor the common bits into functions that are shared between this and the ufuncs?

It looks like the C method gains a significant time advantage when there are multiple geometries:
```
>>> geom = np.repeat(pg.linestrings(np.arange(1000), np.arange(1000)), 100)
>>> timeit('pickle.loads(pickle.dumps(geom))', number=100, globals={'geom': geom, 'pg': pg, 'pickle': pickle})
0.020391266999922664
>>> timeit('pg.from_wkb(pickle.loads(pickle.dumps(pg.to_wkb(geom))))', number=100, globals={'geom': geom, 'pg': pg, 'pickle': pickle})
1.2352546539998457
```

caspervdw:
Thanks for the feedback.

About the linearrings: let's indeed warn the user about this drawback of pickling.

The code duplication bothers me as well. Since #182 , we could implement the `Geometry` features in pure python, this would enable us to remove a lot of complexity. But this should be accompanied by some benchmarks.

The performance benchmark you sent is very surprising to me. I expected that the vectorization in `pygeos.to_wkb` would give you a performance benifit. But I don't know how the pickling of an object-dtype ndarray works precisely.

jorisvandenbossche:
Agreed that WKB should do, the fact that linearrings are not preserved on roundtrip can be a documented limitation.

I am also quite surprised to see that performance characteristic, as I actually did that (converting to WKB vectorized, and then pickling the result of that) in geopandas (https://github.com/geopandas/geopandas/blob/55b99c66081b0281c3ad3ee3ea0fcf14502ba473/geopandas/array.py#L390-L397), assuming this would be faster than letting python pickle each object of the numpy object dtype array individually ...

Now, trying out some other cases, it seems to depend on the lenth of the array.

Original example with a small array (100 elements)  of relatively large linestrings:

```
In [1]: import pygeos

In [2]: import pickle

In [3]: arr = np.repeat(pygeos.linestrings(np.arange(1000), np.arange(1000)), 100)

In [4]: arr.shape
Out[4]: (100,)

In [5]: %timeit pickle.loads(pickle.dumps(arr))
148 µs ± 8.52 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [6]: %timeit pygeos.from_wkb(pickle.loads(pickle.dumps(pygeos.to_wkb(arr))))
13.9 ms ± 1.37 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Large array of tiny linestrings:

```
In [7]: arr2 = pygeos.linestrings(np.random.randn(100000, 3, 2))

In [8]: %timeit pickle.loads(pickle.dumps(arr2))
275 ms ± 6.31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: %timeit pygeos.from_wkb(pickle.loads(pickle.dumps(pygeos.to_wkb(arr2))))
157 ms ± 4.74 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

and now the other method is actually faster. 

Now, for a comparison relevant to this PR, we should actually compare against having it in Python vs in C, so that's still something else

jorisvandenbossche:
Ah, I think I understand what is happening. The array created with `np.repeat` actually contains all the same objects (since it is object dtype array, I suppose it can point to the same python object), and I suppose that pickling has some optimization to detect identical objects being pickled. 
Anyway, when creating the same array but with unique objects (the more typical case in practice), the first pickling method is not significantly faster anymore:

```
In [12]: arr = np.array([pygeos.linestrings(np.arange(1000), np.arange(1000)) for _ in range(100)], dtype=object)

In [13]: %timeit pickle.loads(pickle.dumps(arr))
12.3 ms ± 213 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [14]: %timeit pygeos.from_wkb(pickle.loads(pickle.dumps(pygeos.to_wkb(arr))))
13.1 ms ± 1.26 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

(and I suppose the vectorization benefit will mainly show for larger arrays than 100 elements)

jorisvandenbossche:
Comparing with my shapely branch that subclasses from Geometry in Python, and adding a simple `__reduce__` there:

```
    def __reduce__(self):
        return (pygeos.from_wkb, (self.wkb, ))
```

and repeating the same examples as above:

```
In [9]: arr = np.array([pygeos.linestrings(np.arange(1000), np.arange(1000)) for _ in range(100)], dtype=object)

In [10]: %timeit pickle.loads(pickle.dumps(arr))
15.9 ms ± 747 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [11]: arr2 = pygeos.linestrings(np.random.randn(100000, 3, 2))

In [12]: %timeit pickle.loads(pickle.dumps(arr2))
1.4 s ± 77.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So for the second case (array with many elements), there seems to be quite some overhead from this being implemented in python  (275ms -> 1.4s).

caspervdw:
Thanks for all these benchmarks! And I am happy to read that this PR is already optimal.

We get the extra performance at the expense of some code duplication.

Is this PR good to go?

jorisvandenbossche:
Working on the shapely subclasses branch (https://github.com/Toblerity/Shapely/pull/983) this morning, I was thinking about the linestring/linearring a bit more: in principle we could "relatively easy" pickle the geometry type id as well, together with the WKB? In which case we could ensure type stability?

Now, we would then need to have a custom "deserialize" function, which is not just the class constructor, so not sure how complex this part would be.

brendan-ward:
> pickle the geometry type id as well, together with the WKB

That feels like it is duplicating the geometry type already stored into the WKB for all those types already well-typed there.

Another idea would be to come up with a relatively safe 32 bit identifier for linear rings (as in, unlikely to collide with future geometry type ids) and store that into the WKB instead of the ID for linear string in bytes 2-5 (first serialize the WKB then update these bytes).  Then on decoding, just decode those bytes back to an int, check against this well known code to set a flag that it is a linear ring instead of line string, rewrite the WKB (to flip back to linear string id) and decode as now, and convert to a linear ring afterward.

Since the pickled WKB is not intended to be portable outside this context, perhaps that would be a reasonable compromise?  It *seems* like this would be relatively easy to implement since it mostly involves modifying bytes in a well known location... (not sure about actual complexity or overhead)

caspervdw:
So we're back on supporting linearrings? We also have the GeometryCollections, these may contain LinearRings as well (recursively).

I am no big fan on going down this rabbit hole again.

caspervdw:
Would serializing linearrings to WKT (instead of WKB) be acceptable?

brendan-ward:
It is unfortunate that neither serialization spec (WKT or WKB) perfectly maps to the geometry structures we're able to implement here, and that they are not even consistent with each other.  It seems like the decision not to acknowledge linear rings as a valid top-level geometry type was made in the WKB spec, and not really for us to override here.  My suggestions above add complexity with unclear benefit.

Personally, I like the route of WKB with linear rings => linestrings as a documented issue.  There are mutiple user-side workarounds (convert back to linear ring after unpickling, make into a polygon before pickling) that we can suggest.  We can either warn on pickling linear rings, or raise an exception and force users to convert before pickling.

What are we thinking as a version stability policy for pickled objects?  Perhaps for now, we warn that we make no stability promise, and that objects can only be unpickled with exactly the same version of `pygeos` as they were pickled?

pygeos/test/test_io.py
```diff
@@ -349,3 +350,11 @@ def test_from_shapely_error(geom):
 def test_from_shapely_incompatible_versions():
     with pytest.raises(ImportError):
         pygeos.from_shapely(point)
+
+
+@pytest.mark.parametrize("geom", all_types)
+def test_pickle(geom):
+    if pygeos.get_type_id(geom) == 2:
```
brendan-ward, line 357: Suggestion: instead of expecting to fail on this case, how about expecting the expected output (in this case linear ring => linestring):
```
    unpickled = pickle.loads(pickle.dumps(geom))

    if pygeos.get_type_id(geom) == 2:
        # linear rings are converted to linestrings on unpickling
        expected = pygeos.linestrings(pygeos.get_coordinates(geom))
        assert pygeos.equals_exact(unpickled, expected)

    else:
        assert pygeos.equals_exact(unpickled, geom)
```

That way we know it works correctly in all cases, but that a defined case produces different output.

src/pygeom.c
```diff
@@ -97,6 +97,61 @@ static PyObject *GeometryObject_ToWKT(GeometryObject *obj, char *format)
...
+        return Py_None;
+    }
+
+    GEOS_INIT;
+
+    // WKB Does not allow empty points.
+    // We check for that and patch the POINT EMPTY if necessary
+    has_empty = has_point_empty(ctx, obj->ptr);
+    if (has_empty == 2) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+    if (has_empty) {
```
brendan-ward, line 120:
```suggestion
    if (has_empty == 1) {
```
For clarity, it might be good to make this explict

src/pygeom.c
```diff
@@ -97,6 +97,61 @@ static PyObject *GeometryObject_ToWKT(GeometryObject *obj, char *format)
...
+        geom = point_empty_to_nan_all_geoms(ctx, obj->ptr);
+    } else {
+        geom = obj->ptr;
+    }
+
+    /* Create the WKB writer */
+    writer = GEOSWKBWriter_create_r(ctx);
+    if (writer == NULL) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+    // Allow 3D output and include SRID
+    GEOSWKBWriter_setOutputDimension_r(ctx, writer, 3);
```
brendan-ward, line 130: It doesn't look like the 3rd dimension is tested; geometries with Z value are not included in the `all_types` used for test inputs to pickle tests.

src/pygeom.c
```diff
@@ -97,6 +97,61 @@ static PyObject *GeometryObject_ToWKT(GeometryObject *obj, char *format)
...
+    GEOSWKBWriter_setIncludeSRID_r(ctx, writer, 1);
```
brendan-ward, line 131: It doesn't look like geomentries with SRIDs are tested.

src/pygeom.c
```diff
@@ -97,6 +97,61 @@ static PyObject *GeometryObject_ToWKT(GeometryObject *obj, char *format)
...
+    if (has_empty & (geom != NULL)) {
+        GEOSGeom_destroy_r(ctx, geom);
+    }
+    if (writer != NULL) {
+        GEOSWKBWriter_destroy_r(ctx, writer);
+    }
+    if (wkb != NULL) {
+        GEOSFree_r(ctx, wkb);
+    }
+    GEOS_FINISH;
```
brendan-ward, line 151:
```suggestion

    GEOS_FINISH;
    
```
(suggest adding space above and below, like other usages)

src/pygeom.c
```diff
@@ -113,40 +168,40 @@ static PyObject *GeometryObject_str(GeometryObject *self)
     return GeometryObject_ToWKT(self, "%s");
 }
 
+static PyObject *GeometryObject_reduce(PyObject *self)
```
brendan-ward, line 171: A docstring / comment would be helpful here, as the function name alone does not define its purpose.

src/pygeom.c
```diff
@@ -113,40 +168,40 @@ static PyObject *GeometryObject_str(GeometryObject *self)
...
+            GeometryObject_ToWKB((GeometryObject *)self)
+        )
+    );
+}
+
 static Py_hash_t GeometryObject_hash(GeometryObject *self)
 {
-    unsigned char *wkb;
-    size_t size;
+    PyObject *wkb;
```
brendan-ward, line 186:
```suggestion
    PyObject *wkb = NULL;
```

src/pygeom.c
```diff
@@ -113,40 +168,40 @@ static PyObject *GeometryObject_str(GeometryObject *self)
...
+    wkb = GeometryObject_ToWKB(self);
+    if (wkb == NULL) { return -1; }
 
-    GEOSWKBWriter_setOutputDimension_r(ctx, writer, 3);
-    GEOSWKBWriter_setIncludeSRID_r(ctx, writer, 1);
-    wkb = GEOSWKBWriter_write_r(ctx, writer, self->ptr, &size);
-    GEOSWKBWriter_destroy_r(ctx, writer);
-    if (wkb == NULL) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
-    x = PyHash_GetFuncDef()->hash(wkb, size);
+    x = wkb->ob_type->tp_hash(wkb);
```
brendan-ward, line 196: A comment would be helpful here, what is this for?

src/pygeom.c
```diff
@@ -113,40 +168,40 @@ static PyObject *GeometryObject_str(GeometryObject *self)
...
     if (x == -1) {
         x = -2;
     } else {
         x ^= 374761393UL;  // to make the result distinct from the actual WKB hash //
     }
-    GEOSFree_r(ctx, wkb);
+    Py_DECREF(wkb);
```
brendan-ward, line 202:
```suggestion

    Py_DECREF(wkb);
```
(space above)

src/pygeom.c
```diff
@@ -220,6 +275,42 @@ static PyObject *GeometryObject_FromWKT(PyObject *value)
         }
 }
 
+static PyObject *GeometryObject_FromWKB(PyObject *value)
+{
+    PyObject *result = NULL;
+    unsigned char *wkb;
```
brendan-ward, line 281:
```suggestion
    unsigned char *wkb = NULL;
```

src/pygeom.c
```diff
@@ -220,6 +275,42 @@ static PyObject *GeometryObject_FromWKT(PyObject *value)
...
+    Py_ssize_t size;
+    GEOSGeometry *geom;
```
brendan-ward, line 283:
```suggestion
    GEOSGeometry *geom = NULL;
```

src/pygeom.c
```diff
@@ -220,6 +275,42 @@ static PyObject *GeometryObject_FromWKT(PyObject *value)
...
+    GEOSWKBReader *reader;
```
brendan-ward, line 284:
```suggestion
    GEOSWKBReader *reader = NULL;
```

src/pygeom.c
```diff
@@ -220,6 +275,42 @@ static PyObject *GeometryObject_FromWKT(PyObject *value)
...
+
+    /* Cast the PyObject bytes to char* */
+    if (PyBytes_Check(value)) {
```
brendan-ward, line 287: Recommend flipping the conditions to flatten the flow a bit:

```
if (!PyBytes_Check(value)) {
    PyErr_Format(PyExc_TypeError, "Expected bytes, found %s", value->ob_type->tp_name);
    return NULL;
}

size = PyBytes_Size(value);
wkb = (unsigned char *)PyBytes_AsString(value);
if (wkb == NULL) { return NULL; }
```

src/pygeom.c
```diff
@@ -220,6 +275,42 @@ static PyObject *GeometryObject_FromWKT(PyObject *value)
...
+        PyErr_Format(PyExc_TypeError, "Expected bytes, found %s", value->ob_type->tp_name);
+        return NULL;
+    }
+
+    GEOS_INIT;
+
+    reader = GEOSWKBReader_create_r(ctx);
+    if (reader == NULL) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+    geom = GEOSWKBReader_read_r(ctx, reader, wkb, size);
+    GEOSWKBReader_destroy_r(ctx, reader);
```
brendan-ward, line 301: suggest moving to `finish` similar to flow in `GeometryObject_ToWKB`

src/pygeom.c
```diff
@@ -220,6 +275,42 @@ static PyObject *GeometryObject_FromWKT(PyObject *value)
...
+    if (geom == NULL) { errstate = PGERR_GEOS_EXCEPTION; goto finish; }
+    result = GeometryObject_FromGEOS(geom, ctx);
```
brendan-ward, line 303:
```suggestion

    result = GeometryObject_FromGEOS(geom, ctx);
```
(space above)

src/pygeom.c
```diff
@@ -230,20 +321,22 @@ static PyObject *GeometryObject_new(PyTypeObject *type, PyObject *args,
...
+    } else if (PyBytes_Check(value)) {
+        return GeometryObject_FromWKB(value);
+    } else {
         PyErr_Format(PyExc_TypeError, "Expected string, got %s", value->ob_type->tp_name);
         return NULL;
     }
 }
 
 static PyMethodDef GeometryObject_methods[] = {
+    {"__reduce__", GeometryObject_reduce, METH_NOARGS, "For pickling."},
```
brendan-ward, line 333: I haven't had a chance to figure out why, but this is raising a compiler warning about incompatible pointer types:

```
src/pygeom.c:333:20: warning: incompatible pointer types initializing 'PyCFunction' (aka 'struct _object *(*)(struct _object *, struct _object *)') with an expression of type 'PyObject *(PyObject *)' (aka 'struct _object *(struct _object *)')
      [-Wincompatible-pointer-types]
    {"__reduce__", GeometryObject_reduce, METH_NOARGS, "For pickling."},
```

brendan-ward (changes requested):
Sorry for not reviewing sooner!  Thank you for your work on this.  We can revisit consolidating code in #165 at a later time.

In addition to the work around pickling, supporting WKB in `GeometryObject_new` may open up future opportunities (not sure what yet).

A few suggested changes here, the biggest being around tests:
- we should expect an output of linearstrings from linear rings
- it would be good to add tests for 3D geometries and geometries with SRIDs, or create an issue to come back to those later.

Otherwise things looked pretty solid to me.  Note: I'm not familiar with pickling, so I may well have missed something important here.

jorisvandenbossche:
> So we're back on supporting linearrings?

Not necessarily, we can still decide not to. The main reason I raised it again is because I noticed in my shapely branch that in *current* shapely LinearRing is preserved in pickle roundtrip. So it would be a regression for shapely to no longer do that. But so maybe something to discuss with the other shapely devs if we are fine with such a regression.

> We also have the GeometryCollections, these may contain LinearRings as well 

Indeed .. and here shapely is also returning a LineString after roundtrip:

```
In [1]: from shapely.geometry import LinearRing, GeometryCollection  

In [4]: g = GeometryCollection([LinearRing([(0,0), (1,1), (0,1), (0,0)])])   

In [5]: print(g)  
GEOMETRYCOLLECTION (LINEARRING (0 0, 1 1, 0 1, 0 0))

In [6]: import pickle 

In [8]: print(pickle.loads(pickle.dumps(g))) 
GEOMETRYCOLLECTION (LINESTRING (0 0, 1 1, 0 1, 0 0))
```

jorisvandenbossche:
>  But so maybe something to discuss with the other shapely devs if we are fine with such a regression.

I opened https://github.com/Toblerity/Shapely/issues/988 to discuss it there as well. 

Now, while writing it, I realized we could perfectly override the default pickling behaviour in Python specifically for the LinearRing subclass. So that way we keep it fast / simple in C for all geometry types except LinearRings, but since that is a special case anyway, it's probably also not a problem that pickling many linearrings is a bit slower.

Now, for eg GeoPandas, that would still give an inconsistency, if it uses the vectorized `to_wkb` on the array first before pickling .. (but that is then something for GeoPandas to decide)

jorisvandenbossche:
> In addition to the work around pickling, supporting WKB in GeometryObject_new may open up future opportunities (not sure what yet).

Personally, I am not a big fan of overloading the `Geometry(..)` constructor like that, certainly if we don't have a specific use case in mind. I personally find using `from_wkb` much more explicit.

mwtoews:
Another limitation of WKB to note here is that it cannot differentiate between `POINT(nan nan)` and `POINT EMPTY` (https://github.com/pygeos/pygeos/issues/196#issuecomment-697214220).

caspervdw:
Thanks for the review @brendan-ward . I implemented your suggestions. Also I wrote a documentation on `Geometry` objects, it was missing. Good to merge now?

docs/geometry.rst
```diff
@@ -1,5 +1,78 @@
...
+========
+
+The `pygeos.Geometry` class is the central datatype in PyGEOS. 
+An instance of `pygeos.Geometry` is a container of the actual GEOSGeometry object.
+The Geometry object keeps track of the underlying GEOSGeometry and
+lets the python garbage collector free its memory when it is not
+used anymore.
+
+Geometry objects are immutable. This means that after constructed, they cannot
+be changed inplace. Every PyGEOS operation will result in a new object being returned.
```
brendan-ward, line 11:
```suggestion
be changed in place. Every PyGEOS operation will result in a new object being returned.
```

docs/geometry.rst
```diff
@@ -1,5 +1,78 @@
...
+- empty points are transformed to 3D in GEOS < 3.8
+- empty points are transformed to 2D in GEOS 3.8
+
+Hashing
+~~~~~~~
+
+Geometries can be used as elements in sets or as keys in dictionaries.
+Python uses a technique called *hashing* for lookups in these datastructures.
+PyGEOS generates this hash from the WKB representation.
+Therefore, geometries are equal if and only if their WKB representations are equal.
```
brendan-ward, line 54: Do you think it would be a good idea to add the caveat about WKB limitations here as well?  Or have a separate section for WKB limits (since it applies to pickle, hash, to_wkb / from_wkb), and crosslink to it from here?

I haven't tested, but it seems like the linear ring => linestring or issues around empty / nan points applies here as well.

brendan-ward (commented):
Thanks for updating this @caspervdw .

The docs are a really nice addtion!

Please see a few outstanding prior suggestions in the PR (comments for a few things in C, initializing a couple variables to `NULL`); I've resolved all of those that you addressed in your most recent changes.

caspervdw:
> The docs are a really nice addtion!

Thanks!

> Please see a few outstanding prior suggestions in the PR (comments for a few things in C, initializing a couple variables to `NULL`); I've resolved all of those that you addressed in your most recent changes.

I missed them because github folded them for me. I think I addressed them all now.

brendan-ward (approved):
Thanks @caspervdw !

docs/geometry.rst
```diff
@@ -1,5 +1,74 @@
...
+~~~~~~~~
+
+Geometries can be serialized using pickle:
+
+  >>> import pickle
+  >>> pickled = pickle.dumps(point_1)
+  >>> pickle.loads(point_1)
+  <pygeos.Geometry POINT (5.2 52.1)>
+
+.. warning:: Pickling is not supported for linearrings and empty points.
```
jorisvandenbossche, line 38: Why not for empty points? (or do you mean it might not preserve the dimension?)

docs/geometry.rst
```diff
@@ -1,5 +1,74 @@
...
+~~~~~~~~~~~~
+
+For convenience, the ``Geometry`` class can be constructed with a WKT (Well-Known Text)
+or WKB (Well-Known Binary) representation of your geometry:
+
+.. code:: python
+
+  >>> from pygeos import Geometry
+  >>> point_1 = Geometry("POINT (5.2 52.1)")
+  >>> point_2 = Geometry(b"\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?")
```
jorisvandenbossche, line 23: What is the added value of supporting WKB here? IMO that only duplicates `from_wkb`, while being less explicit.

jorisvandenbossche (commented):
Can you add a test for hashing of empty point (then this closes also https://github.com/pygeos/pygeos/issues/196, I suppose)

Issue #196, opened 2020-09-23
GEOSException with hash on POINT EMPTY


mwtoews:
PyGEOS now has two good features:
 * #102 hash support
 * #179 `POINT EMPTY` supported by `to_wkb`

It would be expected that `POINT EMPTY` could be hashed, however GEOSException is raised:
```python
from pygeos import Geometry, to_wkb

g = Geometry('POINT EMPTY')

to_wkb(g)  # now works!

hash(g)
---------------------------------------------------------------------------
GEOSException                             Traceback (most recent call last)
<ipython-input-34-ace70842997c> in <module>
----> 1 hash(g)

GEOSException: IllegalArgumentException: Empty Points cannot be represented in WKB
```

mwtoews:
Looking at `point_nan` from 8d1b66d47083054e9c9ff23a0c9787a9b54a44a3, this is not an empty point:
```python
import pygeos
import numpy as np

point_nan = pygeos.points(np.nan, np.nan)
point_empty = pygeos.Geometry('point empty')

pygeos.is_empty(point_nan)  # False
pygeos.is_empty(point_empty)  # True

repr(point_nan)  # <pygeos.Geometry POINT (nan nan)>
repr(point_empty)  # <pygeos.Geometry POINT EMPTY>

hash(point_nan)  # 8484330208244771434
# hash(point_empty)  # crash

pygeos.to_wkb(point_nan) == pygeos.to_wkb(point_empty) # True
```
This behaviour (apart from the hash crash) isn't wrong, and is similar to PostGIS (via lwgeom):
```sql
SELECT ST_AsEWKT(point_nan), ST_AsEWKT(point_empty),
  ST_IsEmpty(point_nan), ST_IsEmpty(point_empty),
  ST_AsEWKT(point_nan) = ST_AsEWKT(point_empty) AS wkt_eq,
  ST_AsEWKB(point_nan) = ST_AsEWKB(point_empty) AS wkb_eq
FROM (
  SELECT ST_Point('nan'::float, 'nan'::float) AS point_nan,
                      'POINT EMPTY'::geometry AS point_empty
) AS f;
-- -[ RECORD 1 ]--------------
-- st_asewkt  | POINT(nan nan)
-- st_asewkt  | POINT EMPTY
-- st_isempty | f
-- st_isempty | t
-- wkt_eq     | f
-- wkb_eq     | t
```
So these two points have different WKT but same WKB.

mwtoews:
[...]

PR #199, opened 2020-09-26
Fix bug when setting coordinates on an empty point

caspervdw:
This confirms that #196 was indeed solved in master by #190 . The test however uncovered another bug in `set_coordinates`, it raised a GEOSException when setting coordinates on an empty point.

Also, I cleaned up the old `NAG` geometrytype (we use `None` for that since a long time)


jorisvandenbossche (approved):
Thanks, looks good to me!

PR #202, opened 2020-09-28
Minor doc changes

caspervdw:


docs/geometry.rst
```diff
@@ -67,8 +67,8 @@ This is the same as using :func:`pygeos.predicates.equals_exact` with a ``tolera
 Properties
 ~~~~~~~~~~
 
-Geometry objects have no attributes or properties.
-Instead, use functions listed below to obtain information about geometry objects.
+Geometry objects have neither attributes nor properties.
```
brendan-ward, line 70: Should this be `Geometry objects have neither attributes nor methods`?  It seems like attributes are synonymous with properties in this context.

brendan-ward (approved):
Thanks @caspervdw !

Issue #129, opened 2020-03-25
C cleanup / optimization


brendan-ward:
Running list of things spotted in C code that can be cleaned up, refactored, or otherwise optimized:

* [ ] `count_coordinates` and `get_coordinates` can use `METH_O` for args in instead of `METH_VARARGS`, and skip arg parsing
* [ ] STRtree `_geoms` should allocate an array since size is known at construction time, instead of using resizeable vector, which results in lots of reallocations.

jorisvandenbossche:
Related to "C cleanup", I have also been wondering if using some style formatter would be useful (certainly with C noobs like me (us?) as collaborators), in the line of using black in python. At work, we use clang-format for C++, and that also works for C. But that might be a bit heavy dependency for just formatting. Or trying to abide to some a specific formatting guideline could also already a start.

PR #203, opened 2020-09-28
C code formatting

caspervdw:
I ran [clang-format](http://clang.llvm.org/docs/ClangFormat.html) on the code (whithout having a look at the settings).

It looks nicer now.


jorisvandenbossche:
Cool, I am very much in favor of using some formatter, like we also use black for python (xref https://github.com/pygeos/pygeos/issues/129#issuecomment-603996645), so +1.

For Apache Arrow, we specify to use Google style, and a slightly larger line length: https://github.com/apache/arrow/blob/master/.clang-format

brendan-ward (approved):
Agreed - this looks much better!

PR #204, opened 2020-09-28
Re-add GeometryType -1 (but name it MISSING)

caspervdw:
Ref. https://github.com/geopandas/geopandas/issues/1640


jorisvandenbossche (approved):
Yes, I think it is good to add this back

PR #207, opened 2020-10-01
Release GIL for is_geometry, is_missing and is_valid_input

caspervdw:


jorisvandenbossche:
Is this failing now? 

With pygeos master and my shapely branch, this seems to work fine:

```
In [10]: p = shapely.geometry.Point(0,0)

In [11]: pygeos.is_geometry([p, None, 1])
Out[11]: array([ True, False, False])

In [12]: pygeos.is_missing([p, None, 1])
Out[12]: array([False,  True, False])

In [13]: pygeos.is_valid_input([p, None, 1])
Out[13]: array([ True,  True, False])

```

caspervdw:
You're right. I thought it wouldn't work, but it actually does work.

Still nothing wasted, because this change allows for releasing the GIL :)

src/ufuncs.c
```diff
@@ -165,24 +165,33 @@ static PyUFuncGenericFunction Y_b_funcs[1] = {&Y_b_func};
 
 /* Define the object -> bool functions (O_b) which do not raise on non-geom objects*/
 static char IsMissing(void* context, PyObject* obj) {
-  return ((PyObject*)obj == Py_None);
+  GEOSGeometry* g = NULL;
+  if (!get_geom((GeometryObject *) obj, &g)) {
+    return 0;
+  };
+  return g == NULL;
```
jorisvandenbossche, line 172: Can you add a comment that `g == NULL` is basically a check for "is None"?

src/ufuncs.c
```diff
@@ -165,24 +165,33 @@ static PyUFuncGenericFunction Y_b_funcs[1] = {&Y_b_func};
...
 }
 static void* is_missing_data[1] = {IsMissing};
 static char IsGeometry(void* context, PyObject* obj) {
-  return (PyObject_IsInstance(obj, (PyObject*)&GeometryType));
```
jorisvandenbossche, line 172: I suppose the reason it worked before as well is that IsInstance should work for subclasses?

jorisvandenbossche (commented):
> Still nothing wasted, because this change allows for releasing the GIL :)

Ah, yes, indeed ;)

src/ufuncs.c
caspervdw: Yes, indeed, this PyObject_IsInstance recurses through the MRO.

PR #208, opened 2020-10-02
Fix line_interpolate_point for multilinestrings

### LINKED ISSUE #206: line_interpolate_point with MultiLineString returns POINT EMPTY for GEOS < 3.8.0
Using pygeos with GEOS 3.7.1, `line_interpolate_point` behaves differently, see implementation in [src/ufuncs.c](https://github.com/pygeos/pygeos/blame/7a2798ab8ef43acf59e2023b8cfcb49e6d9384d3/src/ufuncs.c#L341-L365).

```python
import pygeos
multiline = pygeos.Geometry('MULTILINESTRING ((0 0, 2 0), (3 0, 3 6))')
pygeos.line_interpolate_point(multiline, 0)
```
this returns `POINT EMPTY` with GEOS 3.7.1. The expected result is `POINT(0 0)`, as correctly returned with GEOS >= 3.8.0.

caspervdw:
Closes #206

### DIFF
--- a/src/geos.c
+++ b/src/geos.c
@@ -283,26 +283,23 @@ char check_to_wkt_compatible(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
 }
 
 /* GEOSInterpolate_r and GEOSInterpolateNormalized_r segfault on empty
- * linestrings, linearrings, multilinestrings, geometrycollections,
- * or collections with first geometry empty.
+ * geometries and also on collections with the first geometry empty.
  * Returns 1 in those cases, 2 on error, 0 otherwise. */
 char has_empty_line(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
   char type;
   char is_empty;
   const GEOSGeometry* sub_geom;
 
+  is_empty = GEOSisEmpty_r(ctx, geom);
+  if (is_empty != 0) {
+    return is_empty;  // empty (1) or GEOSException (2)
+  }
+
   type = GEOSGeomTypeId_r(ctx, geom);
   if (type == -1) {
     return 2;
-  } else if ((type == GEOS_LINESTRING) | (type == GEOS_LINEARRING)) {
-    // linestring: check if it is empty
-    return GEOSisEmpty_r(ctx, geom);
-  } else if ((type == GEOS_MULTILINESTRING) | (type = GEOS_GEOMETRYCOLLECTION)) {
-    // collection: check if it is empty
-    is_empty = GEOSisEmpty_r(ctx, geom);
-    if (is_empty != 0) {
-      return is_empty;  // empty (1) or GEOSException (2)
-    }
+  } else if ((type == GEOS_MULTIPOINT) | (type == GEOS_MULTILINESTRING) |
+             (type == GEOS_MULTIPOLYGON) | (type = GEOS_GEOMETRYCOLLECTION)) {
     // also check if the first geometry is empty
     sub_geom = GEOSGetGeometryN_r(ctx, geom, 0);
     if (sub_geom == NULL) {
--- a/src/geos.c
+++ b/src/geos.c
@@ -282,6 +282,40 @@ char check_to_wkt_compatible(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
   }
 }
 
+
+/* GEOSInterpolate_r and GEOSInterpolateNormalized_r segfault on empty
+ * linestrings, empty multilinestrings or multilinestrings with the first geometry empty.
+ * Returns 1 in those cases, 2 on error, 0 otherwise. */
+#if !GEOS_SINCE_3_8_0
+char has_empty_line(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
+    char type;
+    char is_empty;
+    const GEOSGeometry* sub_geom;
+
+    type = GEOSGeomTypeId_r(ctx, geom);
+    if (type == -1) {
+      return 2;
+    } else if (type == GEOS_LINESTRING) {
+      // linestring: check if it is empty
+      return GEOSisEmpty_r(ctx, geom);
+    } else if (type == GEOS_MULTILINESTRING) {
+      // multilinestring: check if it is empty
+      is_empty = GEOSisEmpty_r(ctx, geom);
+      if (is_empty != 0) {
+        return is_empty;  // empty (1) or GEOSException (2)
+      }
+      // also check if the first linestring is empty
+      sub_geom = GEOSGetGeometryN_r(ctx, geom, 0);
+      if (sub_geom == NULL) {
+        return 2;  // GEOSException
+      }
+      return GEOSisEmpty_r(ctx, sub_geom);
+    } else {
+      return 0;
+    }
+}
+#endif
+
 /* Define GEOS error handlers. See GEOS_INIT / GEOS_FINISH macros in geos.h*/
 void geos_error_handler(const char* message, void* userdata) {
   snprintf(userdata, 1024, "%s", message);
--- a/src/geos.c
+++ b/src/geos.c
@@ -283,7 +283,8 @@ char check_to_wkt_compatible(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
 }
 
 /* GEOSInterpolate_r and GEOSInterpolateNormalized_r segfault on empty
- * linestrings, empty multilinestrings or multilinestrings with the first geometry empty.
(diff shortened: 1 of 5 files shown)


caspervdw:
Referencing https://github.com/Toblerity/Shapely/pull/983

caspervdw:
One of the new tests (I guess, the `MULTILINESTRING EMPTY` or `MULTILINESTRING (EMPTY, (0 0, 1 1))` still segfaults on GEOS 3.8. I just dropped the version conditional, else it would become very tedious. Checks are now done also for GEOS 3.8.

jorisvandenbossche (changes requested):
Shouldn't we only do this empty check for linestrings / multilinestrings / linearrings? 

Because now, you actually get a proper error for other types of geometries:

```
In [4]: pygeos.line_interpolate_point(pygeos.Geometry("POINT EMPTY"), 0.2)
...
GEOSException: IllegalArgumentException: LinearIterator only supports lineal geometry components

In [5]: pygeos.line_interpolate_point(pygeos.box(0, 0, 1, 1), 0.2)
...
GEOSException: IllegalArgumentException: LinearIterator only supports lineal geometry components
```

while with this change that would now return an empty point *if* the geometry is empty?

caspervdw:
I did some research on what happens for different input values:

GEOS 3.8
```
POINT EMPTY                                                  --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
LINESTRING EMPTY                                             --> POINT EMPTY
LINEARRING EMPTY                                             --> POINT EMPTY
POLYGON EMPTY                                                --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
MULTIPOINT EMPTY                                             --> Segfault
MULTILINESTRING EMPTY                                        --> Segfault
MULTILINESTRING Z (EMPTY, (0 0 0, 1 1 0))                    --> Segfault
MULTIPOLYGON EMPTY                                           --> Segfault
MULTIPOLYGON Z (EMPTY, ((0 0 0, 1 1 0, 1 0 0, 0 0 0)))       --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION EMPTY                                     --> Segfault
GEOMETRYCOLLECTION Z (POINT EMPTY, POINT (1 1))              --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION Z (LINESTRING EMPTY, POINT (1 1))         --> Segfault
POINT (0 0)                                                  --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
LINESTRING (0 0, 1 1)                                        --> POINT (0 0)
LINEARRING (0 0, 1 1, 1 0, 0 0)                              --> POINT (0 0)
POLYGON ((0 0, 1 1, 1 0, 0 0))                               --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
MULTIPOINT (0 0, 1 1)                                        --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
MULTILINESTRING ((0 0, 1 1))                                 --> POINT (0 0)
MULTIPOLYGON (((0 0, 1 1, 1 0, 0 0)))                        --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION (POINT (1 1))                             --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION (LINESTRING (0 0, 1 1))                   --> POINT (0 0)
```

GEOS 3.5
```
POINT EMPTY                                                  --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
LINESTRING EMPTY                                             --> Segfault
LINEARRING EMPTY                                             --> Segfault
POLYGON EMPTY                                                --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
MULTIPOINT EMPTY                                             --> Segfault
MULTILINESTRING EMPTY                                        --> Segfault
MULTILINESTRING Z (EMPTY, (0 0 0, 1 1 0))                    --> Segfault
MULTIPOLYGON EMPTY                                           --> Segfault
MULTIPOLYGON Z (EMPTY, ((0 0 0, 1 1 0, 1 0 0, 0 0 0)))       --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION EMPTY                                     --> Segfault
GEOMETRYCOLLECTION Z (POINT EMPTY, POINT (1 1))              --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION Z (LINESTRING EMPTY, POINT (1 1))         --> Segfault
POINT (0 0)                                                  --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
LINESTRING (0 0, 1 1)                                        --> POINT (0 0)
LINEARRING (0 0, 1 1, 1 0, 0 0)                              --> POINT (0 0)
POLYGON ((0 0, 1 1, 1 0, 0 0))                               --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
MULTIPOINT (0 0, 1 1)                                        --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
MULTILINESTRING ((0 0, 1 1))                                 --> POINT (0 0)
MULTIPOLYGON (((0 0, 1 1, 1 0, 0 0)))                        --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION (POINT (1 1))                             --> IllegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries
GEOMETRYCOLLECTION (LINESTRING (0 0, 1 1))                   --> POINT (0 0)
```

Based on this I think we should replace the GEOS exception with a PyGEOS exception as follows:
- Raise a custom pygeos exception for POINT, POLYGON, MULTIPOINT, MULTIPOLYGON (prevent segfault on e.g. MULTIPOINT EMPTY)
- Return POINT EMPTY for empty geometries
- Raise a custom pygeos exception if the first element of a GEOMETRYCOLLECTION is not LINESTRING / LINEARRING
- Return POINT EMPTY for collections with the first element empty
- Return GEOSInterpolate(geom)

brendan-ward:
Thanks for the exhaustive checking and writeup!

Validating and raising exceptions on invalid input geometry types seems reasonable.

Of the rules, this one feels odd to me, but I understand that it comes from GEOS:

> Return POINT EMPTY for empty geometries

While it is a bit of an existential problem (what is a point on a line that doesn't exist), it seems like it should evaluate to `None` here, which I think would help prevent it from being treated as non-`None` / `NULL` input for later operations.  But returning empty point retains geometry type information, so that we can say that other than exceptions, this always returns point geometries, which is nice from an API clarity / consistency standpoint.

In either case, it is up to the user to check `is_empty()` on inputs to avoid those, or after to filter empty points out.  The updated docstring makes this clear.

src/geos.c
```diff
@@ -284,31 +284,58 @@ char check_to_wkt_compatible(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
...
+  if ((type == GEOS_MULTILINESTRING) | (type == GEOS_GEOMETRYCOLLECTION)) {
     sub_geom = GEOSGetGeometryN_r(ctx, geom, 0);
     if (sub_geom == NULL) {
-      return 2;  // GEOSException
+      return PGERR_GEOS_EXCEPTION;  // GEOSException
+    }
+    type = GEOSGeomTypeId_r(ctx, sub_geom);
+    if (type == -1) {
+      return PGERR_GEOS_EXCEPTION;
+    } else if ((type != GEOS_LINESTRING) & (type != GEOS_LINEARRING)) {
```
brendan-ward, line 328: Should there be an additional check for sub geometries in a geometry collection that are multlinestrings?

caspervdw: `GEOS_Intepolate` luckily does not go into nested geometrycollections. It just raises the `llegalArgumentException: LinearLocation::getCoordinate only works with LineString geometries`, for instance with `GEOMETRYCOLLECTION (MULTILINESTRING ((0 0, 1 1)))` as input.

jorisvandenbossche:
@caspervdw thanks a lot for the thorough analysis and fix. This seems like a nice, consistent solution now.

Should we report some of those inconsistencies or segfaults that still happen on GEOS 3.8 ? Or is there already a trac ticket for it?

caspervdw:
I finally got myself an OSGeo id, so I created a ticket: https://trac.osgeo.org/geos/ticket/1055

PR #209, opened 2020-10-05
Rename normalize -> normalized in linear referencing functions

mwtoews:
This PR renames the argument `normalize` to `normalized` for `line_locate_point` and `line_locate_point` linear referencing functions. This brings the argument name in-line with Shapely (see [object.interpolate](https://shapely.readthedocs.io/en/stable/manual.html#object.interpolate) and [object.project](shapely.readthedocs.io/en/stable/manual.html#object.project)) and the underlying GEOS functions `GEOSProjectNormalized_r` and `GEOSInterpolateNormalized_r`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Version 0.9 (unreleased)
 * Addition of a ``is_ccw()`` function for GEOS >= 3.7 (#201)
 * Added support for pickling to ``Geometry`` objects (#190)
 * Limited the length of geometry repr to 80 characters (#189)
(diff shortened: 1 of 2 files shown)


mwtoews:
One question is if the former "normalize" keyword argument should be supported/deprecated (as is currently implemented via `**kwargs`), or should the argument be simply renamed?

caspervdw (approved):

caspervdw:
Thanks @mwtoews . I think it is good that this DeprecationWarning is in place, we can remove it on a major version bump (or when pygeos is merged into shapely)

Issue #125, opened 2020-03-24
Dropping python 3.5 support


caspervdw:
Following up on https://github.com/pygeos/pygeos/pull/116

We decided that we want to start following numpy's deprecation protocol at https://numpy.org/neps/nep-0029-deprecation_policy.html .

For the next release, we want to limit Python to >= 3.6 and numpy to >= 1.15
GEOS we leave to >= 3.5 until Ubuntu 16 is out of support (April 2021).

@jorisvandenbossche will check with shapely what their policy is

jorisvandenbossche:
All good on using Python 3.6+ here -> https://github.com/Toblerity/Shapely/issues/886

jorisvandenbossche:
In the meantime Python 3.5 support has been removed (#211), and we have

https://github.com/pygeos/pygeos/blob/b3ae3004ab7081f7927d33a0df1abfaf5dc57136/setup.py#L201

So closing this issue.

PR #211, opened 2020-10-14
Drop Python 3.5 from Travis-CI matrix

brendan-ward:
Python 3.5 is now no longer supported.  Pulled this change from #210 plus drop special handling for `pytest`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: minimal
 
 matrix:
   include:
-  - os: linux  # 2015
-    env: PYTHON_VERSION="3.5" RESTORE_FREE_CHANNEL=1 DEPS="numpy=1.10 geos=3.5"
   - os: linux  # 2017
     env: PYTHON_VERSION="3.6" DEPS="numpy=1.13 geos=3.6"
   - os: linux  # 2018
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (commented):
Certainly fine to drop python 3.5, but this change also dropped testing for GEOS 3.5, which is something we still want to keep I think

jorisvandenbossche:
And this also moves the tested numpy version from 1.10 to 1.13 (which I am personally fine with, 1.10 is really old, but I think this should also be a conscious decision)

brendan-ward:
It looks like numpy only [currently supports](https://numpy.org/neps/nep-0029-deprecation_policy.html) 1.13+?

Related discussion in #125; we indeed should keep GEOS 3.5 testing intact until that April 2021 (sorry I missed that here).  This discussion also suggests upgrading minimum numpy to 1.15, though we don't have to do that just yet.

jorisvandenbossche:
> It looks like numpy only currently supports 1.13+?

The table with dates on that page is not what numpy itself supports (they typically only do bug fix releases for the last release anyway), but what they recommend to *other* libraries having numpy as a dependency (so, like pyeos). 
Now, the recommendation on that page is quite focused on a small support range of versions, and many libraries depending on numpy support more versions (also certainly if using Ubuntu as a reference, you'll need older versions ..). I am personally happy to follow that, though (but we should maybe also check with other Shapely devs what they think about the minimum numpy version)

jorisvandenbossche:
BTW, @brendan-ward see also https://github.com/Toblerity/Shapely/issues/1000

PR #212, opened 2020-10-14
Move testing of GEOS 3.5 to Python 3.6

brendan-ward:
Adds back in testing of GEOS 3.5 until April 2021.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: minimal
 
 matrix:
   include:
+  - os: linux  # GEOS 3.5 testing until April 2021
+    env: PYTHON_VERSION="3.6" DEPS="numpy=1.13 geos=3.5"
   - os: linux  # 2017
     env: PYTHON_VERSION="3.6" DEPS="numpy=1.13 geos=3.6"
   - os: linux  # 2018


jorisvandenbossche (approved):
Thanks, looks good to me!

brendan-ward:
Unfortunately GEOS 3.5 is not available with [Python 3.6 in conda](https://travis-ci.org/github/pygeos/pygeos/jobs/735771335), so this is not quite as easy as I hoped.  Don't merge yet...

brendan-ward:
It looks like Python 3.5 is the [default version](https://packages.ubuntu.com/xenial/python/python3) on Ubuntu 16.04.  If we're holding back GEOS version based on 16.04, then it seems like we have to hold back Python to 3.5 as well?

caspervdw:
In that case I think we should build / cache GEOS ourselves (using https://github.com/Toblerity/Shapely/blob/master/ci/travis/install_geos.sh )

mwtoews:
> In that case I think we should build / cache GEOS ourselves (using https://github.com/Toblerity/Shapely/blob/master/ci/travis/install_geos.sh )

I can investigate this for Travis CI.

jorisvandenbossche:
This is now covered by https://github.com/pygeos/pygeos/pull/219

PR #216, opened 2020-10-16
BUG: Fix error handling for line_locate_point (GEOSProject(Normalized))

### LINKED ISSUE #215: BUG: errors in GEOSProject (line_locate_point) are not properly catched
I discovered that GEOS for some cases also raises `std::runtime_error` (instead of using the "error" message handler mechanism, which we capture and raise as Python  exceptions. Not sure why this mechanism is not used for those cases ..)
 
For example, in the `GEOSProject` function, if the second geometry is not a point a runtime error is raised: https://github.com/libgeos/geos/blob/27516e0bdb7526699c00fef328c61561ab7a4b47/capi/geos_ts_c.cpp#L3071 

When running an example that should trigger that runtime error, with PyGEOS we don't see anything special in Python:
```
In [5]: import pygeos

In [6]: pygeos.line_locate_point(pygeos.linestrings([(0, 0), (2, 2)]), pygeos.linestrings([(1, 1), (1.5, 1.5)]))
Out[6]: -1.0
```

As comparison, Shapely is actually somehow "catching" this somewhat (int the logger, but also printing to stdout, not sure where this is done in the Shapely code):

```
In [3]: from shapely.geometry import LineString

In [4]: LineString([(0, 0), (2, 2)]).project(LineString([(1, 1), (1.5, 1.5)]))
third argument of GEOSProject_r must be Point*
Out[4]: -1.0
```

But so general question: do we need to do something specific with those runtime errors? I am also not sure how that works in C since the exceptions is a C++ thing (https://en.cppreference.com/w/cpp/error/runtime_error)

jorisvandenbossche:
Fixes #215

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -887,23 +887,39 @@ static void* frechet_distance_data[1] = {GEOSFrechetDistanceWrapped_r};
 #endif
 /* Project and ProjectNormalize don't return error codes. wrap them. */
 static int GEOSProjectWrapped_r(void* context, void* a, void* b, double* c) {
-  /* Handle empty points (they give segfaults) */
-  if (GEOSisEmpty_r(context, b)) {
+  /* Handle empty points (they give segfaults (for b) or give exception (for a)) */
+  if (GEOSisEmpty_r(context, a) | GEOSisEmpty_r(context, b)) {
     *c = NPY_NAN;
   } else {
     *c = GEOSProject_r(context, a, b);
   }
-  return 1;
+  if (*c == -1.0) {
+    return 0;
+  } else {
+    return 1;
+  }
 }
 static void* line_locate_point_data[1] = {GEOSProjectWrapped_r};
 static int GEOSProjectNormalizedWrapped_r(void* context, void* a, void* b, double* c) {
-  /* Handle empty points (they give segfaults) */
-  if (GEOSisEmpty_r(context, b)) {
+  double length;
+  double distance;
+
+  /* Handle empty points (they give segfaults (for b) or give exception (for a)) */
+  if (GEOSisEmpty_r(context, a) | GEOSisEmpty_r(context, b)) {
     *c = NPY_NAN;
   } else {
-    *c = GEOSProject_r(context, a, b);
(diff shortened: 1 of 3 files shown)


src/ufuncs.c
```diff
@@ -887,23 +887,39 @@ static void* frechet_distance_data[1] = {GEOSFrechetDistanceWrapped_r};
...
+      return 0;
+    };
+    distance = GEOSProject_r(context, a, b);
+    if (distance == -1.0) {
+      return 0;
+    } else {
+      *c = distance / length;
+    }
   }
-  *c = GEOSProjectNormalized_r(context, a, b);
```
jorisvandenbossche, line 906: Here I replaced the call to `GEOSProjectNormalized` with the actual implementation (which is only a very small wrapper around GEOSProject, to divide by the distance). And then directly integrated that code snippet in the wrapper function to have the proper return value.

Reason for this is that there is currently a bug in GEOSProjectNormalized regarding the -1.0 exception value, see my PR to fix this in GEOS: https://github.com/libgeos/geos/pull/341

src/ufuncs.c
```diff
@@ -887,23 +887,39 @@ static void* frechet_distance_data[1] = {GEOSFrechetDistanceWrapped_r};
 #endif
 /* Project and ProjectNormalize don't return error codes. wrap them. */
 static int GEOSProjectWrapped_r(void* context, void* a, void* b, double* c) {
-  /* Handle empty points (they give segfaults) */
-  if (GEOSisEmpty_r(context, b)) {
+  /* Handle empty points (they give segfaults (for b) or give exception (for a)) */
+  if (GEOSisEmpty_r(context, a) | GEOSisEmpty_r(context, b)) {
```
jorisvandenbossche, line 891: Because we now properly catch and raise the GEOS exception, I needed to add this check for `a` being empty to keep the current behaviour, as GEOS itself raises an error (without any message) for empty linestrings.

caspervdw (approved):
Thanks for the fix! Looks good. Can you add a note to the changelog?

Issue #198, opened 2020-09-24
Migrate from Travis / Appyeyor to Github Actions


brendan-ward:
GeoPandas recently [migrated from Travis / Appveyor to Github Actions](https://github.com/geopandas/geopandas/pull/1576) thanks to the hard work of @adriangb.

This simplifies the platform-specific maintenance of configuration files; particularly applicable for integrating Cython as per #51.

As part of this, I think we should create conda environment yml files for each environment we're testing (similar to geopandas).  Will be easier to maintain than deps specified via testing matrix.

jorisvandenbossche:
I think the GitHub Actions certainly give a nice, unified interface. 
Regarding dependencies, pygeos has much less dependencies compared to geopandas (only numpy and GEOS versions), so it might be less worth the overhead of having environment files.

That said, I *personally* think it is more useful to focus "infrastructural" effort towards Shapely. I am actively working on integrating pygeos into Shapely, and so soon we will have to deal with Shapely's CI, and improving it here won't make that easier .. 
(to be clear, I don't want to say that you shouldn't do this (it might also be easier to experiment with it here as in shapely), just that it's not a priority *for me*. And actual code improvements are of course still focused here)

PR #219, opened 2020-11-02
TST: Use native CI build tools to allow flexible GEOS versions

mwtoews:
* Move away from conda, which does not offer many versions of GEOS
* Build and cache GEOS for Travis CI and AppVeyor
* Test Python 3.5 and GEOS 3.5.2
* Test Python 3.6, GEOS 3.6.4, and NumPy 1.13.3
* Test Python 3.7, GEOS 3.7.3, and NumPy 1.15.4
* Test Python 3.8, GEOS 3.8.1, and NumPy 1.17.5
* Test Python 3.9, GEOS 3.8.1, and NumPy 1.19.4
* Test Python 3.9-dev, GEOS master, and NumPY pre-release (Travis CI only)
* Skip `test_geos_capi_version` for GEOS 3.7.x on Windows due to GEOS issue

Supersedes #212

<s>Future considerations could also mix a few versions of NumPy too.</s> Updated to mix NumPy versions.

Also, this PR currently drops osx from the Travis CI testing. Is it worth keeping? It would require a bit more work if this is desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,56 @@
-language: minimal
+---
+language: python
+dist: xenial
+os: linux
 
-matrix:
+cache:
+  directories:
+    - $HOME/geosinstall
+    - ~/.cache/pip
+
+jobs:
   include:
-    - os: linux # 2017
-      env: PYTHON_VERSION="3.6" DEPS="numpy=1.13 geos=3.6"
-    - os: linux # 2018
-      env: PYTHON_VERSION="3.7" DEPS="numpy=1.15 geos=3.7"
-    - os: linux # 2019
-      env: PYTHON_VERSION="3.8" DEPS="numpy=1.17 geos=3.8"
-    - os: linux # now
-      env: PYTHON_VERSION="3.8" CONDA_FORGE=1 DEPS="numpy geos"
-    - os: osx
-      env: PYTHON_VERSION="3.8" DEPS="numpy geos"
+    - python: "3.5"
+      env:  # 2015
+        GEOS_VERSION="3.5.2"
+        NUMPY_VERSION="1.10.4"
+    - python: "3.6"
+      env:  # 2017
+        GEOS_VERSION="3.6.4"
+        NUMPY_VERSION="1.13.3"
+    - python: "3.7"
+      env:  # 2018
+        GEOS_VERSION="3.7.3"
+        NUMPY_VERSION="1.15.4"
+    - python: "3.8"
+      env:  # 2019
+        GEOS_VERSION="3.8.1"
+        NUMPY_VERSION="1.17.5"
+    - python: "3.9"
+      env:  # 2020
+        GEOS_VERSION="3.8.1"
+        NUMPY_VERSION="1.19.4"
+    - python: "3.9-dev"
+      env: GEOS_VERSION="master"
+  allow_failures:
+    - python: "3.9-dev"
+      env: GEOS_VERSION="master"
 
 install:
-  - wget https://repo.anaconda.com/pkgs/misc/conda-execs/conda-latest-$TRAVIS_OS_NAME-64.exe -O conda.exe
-  - chmod +x conda.exe
-  - export CONDA_ALWAYS_YES=1
-  - if [ $RESTORE_FREE_CHANNEL ]; then ./conda.exe config --set restore_free_channel true; fi
-  - if [ $CONDA_FORGE ]; then ./conda.exe config --add channels conda-forge; fi
-  - ./conda.exe create --prefix $HOME/miniconda python=$PYTHON_VERSION conda pytest $DEPS
-  - export PATH="$HOME/miniconda/bin:$PATH"
-  - hash -r
-  - pip install -e . --no-deps -v
+  - export GEOS_INSTALL=$HOME/geosinstall/geos-$GEOS_VERSION
+  - ./ci/install_geos.sh
+  - cd $TRAVIS_BUILD_DIR
+  - pip install --disable-pip-version-check --upgrade pip
+  - pip install --upgrade wheel
+  - if [ "$GEOS_VERSION" = "master" ]; then
+      pip install --upgrade --pre Cython numpy pytest;
+    else
+      pip install --upgrade Cython "numpy==$NUMPY_VERSION" pytest;
+    fi
+  - export LD_LIBRARY_PATH=$GEOS_INSTALL/lib
+  - export PATH=$GEOS_INSTALL/bin:$PATH
+  - python setup.py build_ext --inplace
+  - pip install --no-deps -e .
 
 script:
-  - pytest --doctest-modules;
+  - pytest --doctest-modules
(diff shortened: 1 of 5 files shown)


brendan-ward:
@mwtoews thank you for working on this!

Also see related #198.  I don't know that there is any rush to migrate to Github Actions, but that might enable us to pick back up macos tests.  My hunch is that it is OK not to test on MacOS specifically, at least for a bit.

Since we often recommend conda for installation of this or downstream packages (e.g., geopandas), it seems like it would be a good idea to include at least some testing with conda.  I don't have a good sense of how hard it would be from looking at the changes here, but do you think it would be possible include some core tests via conda for versions of GEOS / python / numpy that conda has available, and then use the custom-compiled changes here for versions not available via conda (or in addition to those available via conda)?

I haven't yet found examples of similar projects that use both.  `fiona` and `shapely` use a similar approach as what you used here (I see you adapted from `fiona`), and `pandas` and `geopandas` appear to only use conda. 

That said, I defer to @jorisvandenbossche and @caspervdw about whether or not we should include testing via conda here, as it does seem like the goals might be a bit different.  The tests (including your changes here) validate that the code is correct for various combinations of versions.  Aside from that, testing with conda specifically seems more about validating that we can resolve an environment that users can also easily install.

mwtoews:
Thanks for the feedback @brendan-ward 

While conda is well suited for managing environments with complex interdependencies, I see pygeos as a relatively low-level component with limited dependencies (in contrast to [geo]pandas). That said, I could now see value in keeping some form of testing within a conda environment to directly benefit primary users. (And yes, I normally use and recommend using Python from a conda environment too.)

Incidentally, Github Actions is something that I'd like to exploring further. From what I've seen, it is better suited to create workflows for (e.g.) macOS, conda, or other tests.

So to get conda tests back in could be possible with Travis CI, but would require a few conditional blocks (perhaps calling external scripts). It could look weird. The other option is to migrate to Github Actions, but would require much more time and effort to achieve.

jorisvandenbossche:
Thanks a lot for working on this @mwtoews !

In general I am a fan of using conda (I refactored the geopandas tests to use conda), but since 1) conda-forge doesn't easily support the range of GEOS versions we want to test, 2) we indeed have a very simply test environment (basically only Python+numpy+GEOS versions), and 3) we need GEOS building support anyway to include GEOS master (which would be nice), I am personally fine with using "native" (non conda) tools in our CI. 

I think we could indeed easily have one build with eg "latest from conda-forge", if we would like to cover this case. But that's probably easier to do with Github Actions, so fine to leave that for a follow-up IMO.

Long term we should certainly keep MacOS, I think. But that migth also be easier with Github Actions. 

> Future considerations could also mix a few versions of NumPy too.

Ideally that happens in this PR, since we already test multiple versions right now. Now, I don't think it should be very complicated to add this back (it's basically the `pip install --upgrade numpy` that needs to get a version from a ENV variable)


BTW, another reason to long term switch from Travis to Github Actions is that Travis moved to a different pricing model, which would mean we need to specifically request for additional credits as open source project (because the free credits would be quickly consumed)

.travis.yml
```diff
@@ -1,29 +1,41 @@
...
-  - export CONDA_ALWAYS_YES=1
-  - if [ $RESTORE_FREE_CHANNEL ]; then ./conda.exe config --set restore_free_channel true; fi
-  - if [ $CONDA_FORGE ]; then ./conda.exe config --add channels conda-forge; fi
-  - ./conda.exe create --prefix $HOME/miniconda python=$PYTHON_VERSION conda pytest $DEPS
-  - export PATH="$HOME/miniconda/bin:$PATH"
-  - hash -r
+  - ./ci/travis/install_geos.sh
+  - cd $TRAVIS_BUILD_DIR
+  - pip install --disable-pip-version-check --upgrade pip
+  - pip install --upgrade wheel
```
jorisvandenbossche, line 33: do we need wheel?

mwtoews: This provides wheel in the case when pip needs to create whl files for dependencies if they are not on PyPI (I think). It's a relatively uncommon thing, but this happened a few days ago while rolling out Python 3.9 support [here](https://ci.appveyor.com/project/frsci/shapely/builds/36080775/job/18hc5h93tjxysw4b#L1636), which was fixed by https://github.com/Toblerity/Shapely/pull/1014

mwtoews:
I've updated this PR with a mixture of NumPy versions. Also, I've unintentionally re-added Python 3.5 (removed with #211) because it is of the similar era as GEOS 3.5. Should Python 3.5 be scrubbed off this PR too? Should macOS be re-added too? ([note](https://docs.travis-ci.com/user/languages/python/#running-python-tests-on-multiple-operating-systems)) 

The two geos_install.* scripts added here have been re-written slightly from their initial version to make them more generic, so they are not specific to either Travis CI or AppVeyor, so should be useful when migrating to (e.g.) Github Actions.

brendan-ward:
@mwtoews thanks for the continued work on this!

Since it looks like your Python 3.5 tests are passing with this setup, it doesn't seem a problem to me to include them for now.  I think once we start introducing Python 3.6+ specific features, we can remove it at that time.  I think mostly we shouldn't spend much effort making sure that Python 3.5 tests always pass, but we _do_ want GEOS 3.5 for a bit longer.

I've been making progress on a companion Github Action in #222  - this will pick up MacOS testing based on Conda, so I don't think you need to worry about MacOS here.

jorisvandenbossche:
This looks good to me!

mwtoews:
Apologies for all the `push -f`'s here, but there were too many trivial edits to care about keeping track of. Pending the last Travis CI check, this is ready for any final review and merge.

The biggest value here is seeing how PyGEOS works with GEOS master, which has been active over the past few days with a [big OverlayNG switch](https://github.com/libgeos/geos/commit/be81a1e139d37ff46b854bae31574e3ae18d0068), e.g. tests from this PR:
 * [before  - with 2 failed, 1534 passed, 9 skipped, 2 xpassed, 41 warnings](https://travis-ci.org/github/pygeos/pygeos/jobs/741016475)
* [after - with 5 failed, 1559 passed, 9 skipped, 2 xpassed, 40 warnings](https://travis-ci.org/github/pygeos/pygeos/jobs/741352056)

I haven't fully assessed these failures yet, but I anticipate a few downstream changes may be needed here and in Shapely.

caspervdw:
Thanks @mwtoews !

@jorisvandenbossche  I don't have the time currently to go through all this. If you feel it is good to merge, please go ahead.

jorisvandenbossche:
Thanks @mwtoews !

As a next step, I think we can look into moving this setup from Travis/Appveyor to Github Actions as well

mwtoews:
@caspervdw a follow-up request that I highly recommend to speed up PR tests is to go to your AppVeyor account settings [here](https://ci.appveyor.com/project/caspervdw/pygeos-3e5cu/settings) and enable "Save build cache in Pull Requests" (the default is not enabled).

caspervdw:
@mwtoews Done!

Op ma 9 nov. 2020 om 09:40 schreef Mike Taves <notifications@github.com>

PR #220, opened 2020-11-03
API change: return 0 for None values in geometry counting functions

### LINKED ISSUE #218: API: get_num_geometries returning -1 for missing values?
Copying @brendan-ward's comment from https://github.com/pygeos/pygeos/pull/197#issuecomment-716961714:

> I just added a fix here for an unexpected result: `get_num_geometries` returns -1 for `None` elements in the array.  Other than where this is returned in `ufuncs.c`, I don't see that we are specifically testing for that result nor declaring it to the user via docstrings, etc.  Is this a documentation / testing gap, or is it a bug?   (related functions declare -1 as expected result for non-geometries)

So as a practical example:

```
In [1]: import pygeos

In [2]: arr = [pygeos.Geometry("POINT (0 0)"), pygeos.Geometry("POINT EMPTY"), None]

In [3]: pygeos.get_num_geometries(arr)
Out[3]: array([ 1,  1, -1], dtype=int32)
```

Personally, I would say that returning 0 makes more sense, certainly because a "count" is something you typically sum (like you were doing in this PR), and then such -1 values will be very confusing / leading to subtle bugs.

### LINKED ISSUE #234: BUG: Windows: testing GEOS operation function signature does not work properly on Windows
Detected in #220 

There are a few places in `ufuncs.c` where we are testing the function signature of an incoming operation to determine appropriate error code, etc.

Example, for `Y_i_func`, given `static void* get_dimensions_data[1] = {GEOSGeom_getDimensions_r};`:

```
  FuncGEOS_Y_i* func = (FuncGEOS_Y_i*)data;

...

  if (((void*)func == GEOSGeom_getDimensions_r) || ((void*)func == GEOSGetSRID_r)) {
    errcode = 0;
  } else {
    errcode = -1;
  }
```

On Windows, `((void*)func == GEOSGeom_getDimensions_r)` does not correctly detect that the operation resolved from `data` matches that function.

This appears to work fine for functions defined directly in the C Extension, like `GetNumPoints`.

This is most pronounced where we are using this type of test to determine return values, like in #220.

brendan-ward:
Resolves #218  
Resolves #234

Previously, `get_num_geometries`, `get_num_coordinates`, `get_num_points`, `get_num_interior_rings` returned -1 for non-geometry values (`None`).

This changes the API to return 0 instead, which is more appropriate for counting type functions, since users may sum these values to perform subsequent operations as noted in #218, #197.

The only previously-stated return value that is changed here is for `get_num_coordinates`; the return value was not previously declared in docstrings for the other functions and -1 is likely unexpected for them.

### DIFF
--- a/pygeos/geometry.py
+++ b/pygeos/geometry.py
@@ -316,6 +316,8 @@ def get_point(geometry, index):
 def get_num_points(geometry):
     """Returns number of points in a linestring or linearring.
 
+    Returns 0 for not-a-geometry values.
+
     Parameters
     ----------
     geometry : Geometry or array_like
@@ -333,6 +335,8 @@ def get_num_points(geometry):
     >>> get_num_points(line)
     4
     >>> get_num_points(Geometry("MULTIPOINT (0 0, 1 1, 2 2, 3 3)"))
+    0
+    >>> get_num_points(None)
     0
     """
     return lib.get_num_points(geometry)
--- a/pygeos/geometry.py
+++ b/pygeos/geometry.py
@@ -128,7 +128,7 @@ def get_coordinate_dimension(geometry):
 def get_num_coordinates(geometry):
     """Returns the total number of coordinates in a geometry.
 
-    Returns -1 for not-a-geometry values.
+    Returns 0 for not-a-geometry values.
 
     Parameters
     ----------
@@ -143,7 +143,7 @@ def get_num_coordinates(geometry):
     >>> get_num_coordinates(Geometry("GEOMETRYCOLLECTION (POINT(0 0), LINESTRING(0 0, 1 1))"))
     3
     >>> get_num_coordinates(None)
-    -1
+    0
     """
     return lib.get_num_coordinates(geometry)
 
@@ -391,6 +391,8 @@ def get_interior_ring(geometry, index):
 def get_num_interior_rings(geometry):
     """Returns number of internal rings in a polygon
 
+    Returns 0 for not-a-geometry values.
+
     Parameters
     ----------
     geometry : Geometry or array_like
@@ -410,6 +412,8 @@ def get_num_interior_rings(geometry):
     >>> get_num_interior_rings(polygon_with_hole)
     1
     >>> get_num_interior_rings(Geometry("POINT (1 1)"))
+    0
+    >>> get_num_interior_rings(None)
     0
     """
     return lib.get_num_interior_rings(geometry)
@@ -457,6 +461,8 @@ def get_geometry(geometry, index):
 def get_num_geometries(geometry):
     """Returns number of geometries in a collection.
 
+    Returns 0 for not-a-geometry values.
+
     Parameters
     ----------
     geometry : Geometry or array_like
@@ -474,5 +480,7 @@ def get_num_geometries(geometry):
     4
     >>> get_num_geometries(Geometry("POINT (1 1)"))
     1
+    >>> get_num_geometries(None)
+    0
     """
     return lib.get_num_geometries(geometry)
--- a/pygeos/geometry.py
+++ b/pygeos/geometry.py
@@ -146,8 +146,6 @@ def get_num_coordinates(geometry):
     1
     >>> get_num_coordinates(Geometry("GEOMETRYCOLLECTION (POINT(0 0), LINESTRING(0 0, 1 1))"))
     3
-    >>> get_num_coordinates(None)
-    0
     """
     return lib.get_num_coordinates(geometry)
 
@@ -533,7 +531,5 @@ def get_num_geometries(geometry):
     4
     >>> get_num_geometries(Geometry("POINT (1 1)"))
     1
-    >>> get_num_geometries(None)
-    0
     """
     return lib.get_num_geometries(geometry)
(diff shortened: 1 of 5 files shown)


jorisvandenbossche (commented):
To be repeat from the issue, I am in favor of changing to return 0. 
@caspervdw thoughts?

@brendan-ward it seems it is actually failing on windows

pygeos/geometry.py
```diff
@@ -132,7 +132,7 @@ def get_coordinate_dimension(geometry):
 def get_num_coordinates(geometry):
     """Returns the total number of coordinates in a geometry.
 
-    Returns -1 for not-a-geometry values.
+    Returns 0 for not-a-geometry values.
```
caspervdw, line 135: "not-a-geometry" is not a term we are using in pygeos, mostly we talk about 'missing' or just "None"  values.

pygeos/test/test_geometry.py
```diff
@@ -135,10 +140,17 @@ def test_get_coordinate_dimension():
     actual = pygeos.get_coordinate_dimension([point, point_z, None]).tolist()
     assert actual == [2, 3, -1]
 
-
+# FIXME:
+@pytest.mark.xfail(sys.platform == "win32", reason="Not correctly detecting function signature")
```
caspervdw, line 144: Is this now solved @brendan-ward ? Otherwise, an approach like in #92 could be taken: supply the function pointer along with the none_value in the ufuncs `data` parameter.

caspervdw (approved):
I agree on this API change. It makes sense to return `0` for these counting functions.

The issue on Windows is a mystery to me though. But it appears @brendan-ward already solved it?

Approving this, of course the debug stuff should be cleaned still and the xfail solved.

pygeos/geometry.py
brendan-ward: The only change here was the return value; `not-a-geometry` was there originally.  For clarity, perhaps we should address changing that in a different PR?

brendan-ward:
Resolves #234 , thanks for the tip about passing extra things to `data` @caspervdw .

pygeos/test/test_geometry.py
```diff
@@ -21,18 +22,20 @@
...
+    assert actual == [0, 3, 5, 0, 0, 0, 0, 0, 0, 0]
 
 
 def test_get_num_interior_rings():
-    actual = pygeos.get_num_interior_rings(all_types + (polygon_with_hole,)).tolist()
-    assert actual == [0, 0, 0, 0, 0, 0, 0, 0, 0, 1]
+    actual = pygeos.get_num_interior_rings(
+        all_types + (polygon_with_hole, None)
+    ).tolist()
+    assert actual == [0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0]
```
jorisvandenbossche, line 33:
```suggestion
    actual = pygeos.get_num_interior_rings(all_types + (polygon_with_hole, None))
    assert actual.tolist() == [0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0]
```

(slightly more readable IMO)

src/ufuncs.c
```diff
@@ -818,10 +818,21 @@ static void Y_d_func(char** args, npy_intp* dimensions, npy_intp* steps, void* d
 static PyUFuncGenericFunction Y_d_funcs[1] = {&Y_d_func};
 
 /* Define the geom -> int functions (Y_i) */
-static void* get_type_id_data[1] = {GEOSGeomTypeId_r};
-static void* get_dimensions_data[1] = {GEOSGeom_getDimensions_r};
-static void* get_coordinate_dimension_data[1] = {GEOSGeom_getCoordinateDimension_r};
-static void* get_srid_data[1] = {GEOSGetSRID_r};
+/* data values are GEOS func, GEOS error code, return value when input is None */
+static void* get_type_id_func_tuple[3] = {GEOSGeomTypeId_r, (void*)-1, (void*)-1};
+static void* get_type_id_data[1] = {get_type_id_func_tuple};
```
jorisvandenbossche, line 823: Defining it in two steps like this is needed? 
(the `_data` object needs to have length 1?)

jorisvandenbossche (commented):
Thanks for figuring this out! 
Maybe can add something to the changelog?

src/ufuncs.c
brendan-ward: Yes, it appeared necessary to do this in 2 parts, like in #92.  I got segfaults when doing it in one step and didn't investigate further once the 2 part approach solved it.

The [numpy ufuncs custom example](https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html?highlight=pyufuncgenericfunction) includes an example of a 4 element `data` value - which implies it should have worked originally, but the example doesn't include mixed types like here (shouldn't matter though?).

@caspervdw any insight into why #92 and here needed to pass these parameters in 2 parts?

caspervdw: Yes I investigated this indeed.

This is actually explained https://numpy.org/doc/stable/reference/c-api/ufunc.html#functions .

> data – Should be NULL or a pointer to an array of size ntypes . 

Where ntypes is:

> ntypes – How many different data-type-specific functions the ufunc has implemented.

In other words, numpy allows for providing multiple functions to let a function accept for example double and single precision floats. The example that @brendan-ward illustrates that nicely - the function there has `data` of length four, but also:

```
PyUFuncGenericFunction funcs[4] = {&half_float_logit,
                                   &float_logit,
                                   &double_logit,
                                   &long_double_logit};
```

PR #221, opened 2020-11-04
ENH: support z-coordinates in apply (coordinate transformation)

### LINKED ISSUE #131: Support get/set_coordinates/apply with more than 2 dimensions

jorisvandenbossche:
Closes #131

Building further upon https://github.com/pygeos/pygeos/pull/178, which already added this to `get_coordinates`, expanding to `set_coordinates`/`apply`. 
For `set_coordinates`, I didn't add a keyword, because there you can know it from the shape of the supplied coordinates array.

### DIFF
--- a/pygeos/coordinates.py
+++ b/pygeos/coordinates.py
@@ -147,9 +147,9 @@ def set_coordinates(geometry, coordinates):
     not include one:
 
     >>> set_coordinates(Geometry("POINT Z (0 0 0)"), [[1, 1]])
-    <pygeos.Geometry POINT (1 1) >
+    <pygeos.Geometry POINT (1 1)>
     >>> set_coordinates(Geometry("POINT Z (0 0 0)"), [[1, 1, 1]])
-    <pygeos.Geometry POINT Z (1 1 1) >
+    <pygeos.Geometry POINT Z (1 1 1)>
     """
     geometry_arr = np.asarray(geometry, dtype=np.object)
     coordinates = np.atleast_2d(np.asarray(coordinates)).astype(np.float64)
--- a/pygeos/coordinates.py
+++ b/pygeos/coordinates.py
@@ -11,7 +11,7 @@ def apply(geometry, transformation, include_z=False):
     With the default of ``include_z=False``, all returned geometries will be
     two-dimensional; the third dimension will be discarded, if present.
     When specifying ``include_z=True``, the returned geometries preserve
-    the dimensionality of the input.
+    the dimensionality of the respective input geometries.
 
     Parameters
     ----------
--- a/pygeos/coordinates.py
+++ b/pygeos/coordinates.py
@@ -4,19 +4,26 @@ import numpy as np
 __all__ = ["apply", "count_coordinates", "get_coordinates", "set_coordinates"]
 
 
-def apply(geometry, transformation):
+def apply(geometry, transformation, include_z=False):
     """Returns a copy of a geometry array with a function applied to its
     coordinates.
 
-    All returned geometries will be two-dimensional; the third dimension will
-    be discarded, if present.
+    With the default of ``include_z=False``, all returned geometries will be
+    two-dimensional; the third dimension will be discarded, if present.
+    When specifying ``include_z=True``, the returned geometries preserve
+    the dimensionality of the input.
 
     Parameters
     ----------
     geometry : Geometry or array_like
     transformation : function
-        A function that transforms a (N, 2) ndarray of float64 to another
-        (N, 2) ndarray of float64.
+        A function that transforms a (N, 2) or (N, 3) ndarray of float64 to
+        another (N, 2) or (N, 3) ndarray of float64.
+    include_z : bool, default False
+        Whether to include the third dimension in the coordinates array
+        that is passed to the `transformation` function. If True, and a
+        geometry has no third dimension, the z-coordinates passed to the
+        function will be NaN.
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


pygeos/coordinates.py
```diff
@@ -4,19 +4,26 @@
...
 
 
-def apply(geometry, transformation):
+def apply(geometry, transformation, include_z=False):
     """Returns a copy of a geometry array with a function applied to its
     coordinates.
 
-    All returned geometries will be two-dimensional; the third dimension will
-    be discarded, if present.
+    With the default of ``include_z=False``, all returned geometries will be
```
caspervdw, line 11: Maybe `include_z` should better be `True`? I don't care too much about breaking the API at this stage of pygeos.

src/coords.c
```diff
@@ -442,6 +452,13 @@ PyObject* SetCoords(PyArrayObject* geoms, PyArrayObject* coords) {
     return (PyObject*)geoms;
   }
 
+  coords_shape = PyArray_SHAPE(coords);
+  if (coords_shape[1] == 3) {
```
caspervdw, line 456: I think this should work as well:

```
include_z = (coord_shape[1] == 3);
```

caspervdw (commented):
Great that you tackled this @jorisvandenbossche ! Two comments.

pygeos/coordinates.py
jorisvandenbossche: I should maybe check the performance overhead in case all your geometries are actually 2D (which I think is by far the more common case) and when using ``include_z=True`` 
And note that this *does* have an influence on the function being passed, because this function will now by default have to deal with z coordinates

jorisvandenbossche: Hmm, doesn't seem to matter much in case of an array of points:

```
In [3]: arr = pygeos.points(np.random.randn(1_000_000, 2))

In [4]: 

In [4]: arr
Out[4]: 
array([<shapely.geometry.Point POINT (1.459720029804123 -1.660916227958426) >,
       <shapely.geometry.Point POINT (-2.336782731229145 2.129473193643444) >,
       <shapely.geometry.Point POINT (0.1039792330580081 -2.048775667056068) >,
       ...,
       <shapely.geometry.Point POINT (0.2416382706879835 -1.697157520118229) >,
       <shapely.geometry.Point POINT (0.2381751392103192 -0.05537962788130606) >,
       <shapely.geometry.Point POINT (-0.3401224146947915 0.2503277994539527) >],
      dtype=object)

In [5]: pygeos.apply(arr, lambda coords: coords*2)
Out[5]: 
array([<shapely.geometry.Point POINT (2.919440059608247 -3.321832455916851) >,
       <shapely.geometry.Point POINT (-4.673565462458289 4.258946387286889) >,
       <shapely.geometry.Point POINT (0.2079584661160162 -4.097551334112135) >,
       ...,
       <shapely.geometry.Point POINT (0.483276541375967 -3.394315040236458) >,
       <shapely.geometry.Point POINT (0.4763502784206384 -0.1107592557626121) >,
       <shapely.geometry.Point POINT (-0.680244829389583 0.5006555989079053) >],
      dtype=object)

In [6]: %timeit pygeos.apply(arr, lambda coords: coords*2)
472 ms ± 4.88 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit pygeos.apply(arr, lambda coords: coords*2, include_z=True)
472 ms ± 2.62 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

jorisvandenbossche: Another default could also be to "infer" the dimensionality, either by checking all geometries (`pygeos.has_z` is quite cheap compared to `pygeos.apply`, it's 16ms for the example above, where the dummy apply takes 470ms) or eg the first geometry. And then the user can avoid this inference by either passing include_z=True/False

But given that it doesn't seem to give much overhead (and actually less that doing a `pygeos.has_z` ...), I am also fine with simply having `include_z=True` as the default. We only need to be aware that the applied function will then always receive a (N, 3) array as argument by default, and not a (N, 2)

brendan-ward: Personally, I prefer the default of `include_z=False`.  Since 2D is much more common, it makes sense to make 3D opt-in, whereas it would be unexpected to have to suppress 3D by always passing `include_z=False`.

caspervdw: Yeah I agree, the argument of having to provide a `(N, 3)` -> `(N, 3) function is convincing

caspervdw (approved):

jorisvandenbossche:
Fixed a small bug, I was still creating a 3D CoordinateSequence for 2D output geometries if the input geometry was 3D. And this was only catched in the *doctest* on older GEOS versions, so also added a proper test for it.

PR #222, opened 2020-11-04
TST: Add Github Action for CI testing using conda

### LINKED ISSUE #198: Migrate from Travis / Appyeyor to Github Actions

brendan-ward:
Resolves #198 

This ports the Travis-CI testing using conda to Github actions, roughly adapted from the geopandas Github CI action.

Appears to be working properly for Ubuntu / MacOS environments.  Still working on getting this to work properly for windows (likely need fixes to GEOS library / include paths).

This is intended to be complementary to a workflow for building from source based on #219


brendan-ward:
I don't think Github will enable this action in this repo until it is on `master`.  See [testing results for my fork](https://github.com/brendan-ward/pygeos/actions) in the meantime

brendan-ward:
All tests now passing all environments and operating systems ([see here](https://github.com/brendan-ward/pygeos/actions)).

.github/workflows/test-conda.yml
```diff
@@ -0,0 +1,55 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        env:
+          - ci/envs/py36-geos36.yml
+          - ci/envs/py37-geos37.yml
+          - ci/envs/py38-geos38.yml
+          - ci/envs/py38-geos-latest.yml
+          - ci/envs/py38-geos-latest-conda-forge.yml
```
jorisvandenbossche, line 18: For such a small env, I have the feeling that the environment.yaml files are a bit overkill, and having here matrix variables for python/numpy/geos version is more flexible to have certain combinations.

jorisvandenbossche (commented):
Thanks for looking into this!

In general, if we also have the builds as now in https://github.com/pygeos/pygeos/pull/219, I think the full matrix here is a bit too much. Maybe only "latest" is enough for the platforms covered on in #219 as well?

brendan-ward:
Closing temporarily to cut down on the noise here; having little success migrating from environment file approach so far, but likely user error interacting with poor local dev tools for validating action workflows...  And the only way to really test is by pushing to Github...

jorisvandenbossche:
> having little success migrating from environment file approach so far

To be clear, it was only a suggestion, without really knowing what it would entail in practice. So if it turns out to be difficult, that's certainly a reason to keep the env files

.github/workflows/test-conda.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: s-weigand/setup-conda@v1.0.5
+        with:
+          activate-conda: false
+          conda-channels: ${{ matrix.channel }}
+          python-version: ${{ matrix.python }}
+
+      - name: Setup Environment
+        shell: bash
+        run: |
+          conda create --name test python=${{ matrix.python}} conda ${{ matrix.geos }} ${{ matrix.numpy }}
```
jorisvandenbossche, line 31: the `conda` as python package to install shouldn't be there I think?

brendan-ward: Right, had added that back in to match the Travis config while trying to get things working, but it was a different issue.  Taking back out now.

brendan-ward:
It looks like the best path forward is to use environment files.  I could not make it work due to undocumented issues with the build matrix / [`include`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations) statement: either we got inexplicable syntax errors or it only ran the last combination of parameters instead of each set.

I don't think it is necessarily a problem to run conda tests on Linux even though it is well-covered by #219.  The full suite goes quickly, and we'll likely bump into issues trying to get the `exclude` statement to work properly for excluding entries from the matrix based on operating system.

Definitely squash this on merge!

jorisvandenbossche:
Sorry for all the troubles with github actions syntax ..  
I did some testing myself, going back to one of your earlier versions with just the latest for each platform, and from there added additional versions *only* for MacOS, which seems to have worked: https://github.com/jorisvandenbossche/pygeos/commit/23fe89ff8bfc928b76c45590f44d0edf792aa6d0 (the actions are passing on my fork). 
What do you think of this?

Although it indeed runs fast, I still think there is value in keeping it sometwhat limited (certainly where it is duplicating the non-conda based tests).

brendan-ward:
@jorisvandenbossche thanks for helping here!  

I think we fell through undocumented aspects of the matrix include rules in previous runs, and I ran into them again in latest tests.  Would be really nice to test this locally and avoid the noise on github, but local tools like [`act`](https://github.com/nektos/act) don't work with the `setup-conda` action we're using here.

It looks like every non-empty matrix variable needs to be included in each `include` block to include an additional variant (i.e., include specific version of python even if same as top-level matrix value), and that the value of the variable must not be empty (i.e., empty string to use latest, unpinned version doesn't work here)

At this point, the matrix now includes:
- testing of Python 3.8 and latest GEOS / numpy on all OS using conda defaults channel
- testing of Python 3.8 and latest GEOS / numpy on all OS using conda-forge channel
- testing of Python 3.8 and pinned versions of GEOS / numpy on macos and windows
- testing of Python 3.6 and pinned versions of GEOS / numpy on macos and windows
- testing of Python 3.7 and pinned versions of GEOS / numpy on macos and windows

That seems comparable enough to the previous Travis config, without doing extra testing of pinned versions on Ubuntu.

caspervdw (approved):
Thanks @brendan-ward !

I am not sure how this test matrix will complement the appveyor / travis tests, but we can sort that out when the PR that is working on the source build of GEOS is in.

jorisvandenbossche:
@brendan-ward Thanks for the update! 

I am still not fully convinced that eg the windows builds with different versions here are necessary (it's only MacOS that is not yet covered by the other PR), but let's merge this now, and we can later rationalize the different builds a bit once the dust settles on the different options (eg we might also want to move the result of #219 to github actions at some point, I suppose)

PR #229, opened 2020-11-05
ENH: offset_curve

### LINKED ISSUE #177: ENH: add parallel_offset / offset_curve function
Shapely has a `parallel_offset` method (https://shapely.readthedocs.io/en/latest/manual.html#object.parallel_offset). 

This seems to be based on GEOS' `GEOSOffsetCurve` (and in older GEOS versions on `GEOSSingleSidedBuffer`, see https://github.com/libgeos/geos/blob/34b8d63e72353166d149db71bf9c45faa0148aaa/capi/geos_c.h.in#L466-L482) 

PostGIS also calls this `ST_OffsetCurve` (https://postgis.net/docs/ST_OffsetCurve.html)

jorisvandenbossche:
Closes #177

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -10,6 +10,7 @@ __all__ = [
     "BufferJoinStyles",
     "boundary",
     "buffer",
+    "offset_curve",
     "centroid",
     "convex_hull",
     "delaunay_triangles",
@@ -63,6 +64,7 @@ def boundary(geometry, **kwargs):
     True
     """
     return lib.boundary(geometry, **kwargs)
+
 
 @multithreading_enabled
 def buffer(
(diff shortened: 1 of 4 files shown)


pygeos/constructive.py
```diff
@@ -166,6 +168,69 @@ def buffer(
...
+    mitre_limit=5.0,
+    **kwargs
+):
+    """
+    Returns a (Multi)LineString at a distance from the object
+    on its right or its left side.
+
+    For positive distance the offset will be at the left side of
+    the input line and retain the same direction. For a negative
+    distance it'll be at the right side and in the opposite
```
caspervdw, line 187:
```suggestion
    distance it will be at the right side and in the opposite
```

caspervdw (approved):
One tiny suggestion, otherwise feel free to merge :+1:

src/ufuncs.c
```diff
@@ -1108,6 +1108,79 @@ static void buffer_func(char** args, npy_intp* dimensions, npy_intp* steps, void
...
+  for (i = 0; i < n; i++, ip1 += is1, ip2 += is2) {
+    /* get the geometry: return on error */
+    if (!get_geom(*(GeometryObject**)ip1, &in1)) {
+      errstate = PGERR_NOT_A_GEOMETRY;
+      destroy_geom_arr(ctx, geom_arr, i - 1);
+      break;
+    }
+
+    width = *(double*)ip2;
+    if ((in1 == NULL) | npy_isnan(width)) {
```
jorisvandenbossche, line 1161: One note here: GEOS actually can "handle" the width being NaN, in the sense that it will raise a proper GEOSException. 
It could also make sense to raise an exception here, as I am not sure there are good use case of the width being NaN (so this doesn't go silent)

PR #230, opened 2020-11-06
CLN: deduplicate some code with macros in geometry-creating ufuncs

jorisvandenbossche:
https://github.com/pygeos/pygeos/pull/156 added quite some code to get the GIL release working with geometry creating ufuncs. I think we can improve maintainability a bit by factoring out some of the parts into additional macros.

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -1113,15 +1113,7 @@ static void offset_curve_func(char** args, npy_intp* dimensions, npy_intp* steps
   GEOSGeometry** geom_arr;
   GEOSGeometry* in1 = NULL;
 
-  // Fail if inputs output multiple times on the same place in memory. That would
-  // lead to segfaults as the same GEOSGeometry would be 'owned' by multiple PyObjects.
-  if ((steps[5] == 0) && (dimensions[0] > 1)) {
-    PyErr_Format(PyExc_NotImplementedError,
-                 "Unknown ufunc mode with args[0]=%p, args[5]=%p, steps[0]=%ld, "
-                 "steps[5]=%ld, dimensions[0]=%ld.",
-                 args[0], args[5], steps[0], steps[5], dimensions[0]);
-    return;
-  }
+  CHECK_NO_INPLACE_OUTPUT(5);
 
   if ((is3 != 0) | (is4 != 0) | (is5 != 0)) {
     PyErr_Format(PyExc_ValueError,
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -54,6 +54,12 @@
     return;                                                                       \
   }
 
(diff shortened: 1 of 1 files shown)


src/ufuncs.c
```diff
@@ -43,6 +43,17 @@
   Py_XDECREF(*out);                                      \
   *out = ret
 
+// Fail if inputs output multiple times on the same place in memory. That would
+// lead to segfaults as the same GEOSGeometry would be 'owned' by multiple PyObjects.
+#define CHECK_NO_INPLACE_OUTPUT(N)                                                \
+  if ((steps[N] == 0) && (dimensions[0] > 1)) {                                   \
+    PyErr_Format(PyExc_NotImplementedError,                                       \
```
jorisvandenbossche, line 50: @caspervdw do you remember if it's actually possible to construct an example ufunc call in python that triggers this exception? (I thought to add a test for it, but wasn't able to get one, tried a few things based on the discussion in #156)

jorisvandenbossche:
Other possible snippets that are repeated in all of the geometry-creating ufuncs are:

```
  // allocate a temporary array to store output GEOSGeometry objects
  geom_arr = malloc(sizeof(void*) * dimensions[0]);
  if (geom_arr == NULL) {
    PyErr_SetString(PyExc_MemoryError, "Could not allocate memory");
    return;
  }
```

and 

```
  // fill the numpy array with PyObjects while holding the GIL
  if (errstate == PGERR_SUCCESS) {
    geom_arr_to_npy(geom_arr, args[1], steps[1], dimensions[0]);
  }
  free(geom_arr);
```

Do people think those are also worth replacing with a macro? (those are shorter, so less sure here if it will necessarily improve readability with a macro)

brendan-ward:
Per the snippets above, it seems like a good idea to keep `malloc` and `free` visible within the top-level function, rather than potentially hiding them in a macro.

However, this could potentially get put into a maco like `CHECK_ALLOC(geom_arr)` to centralize the error message without obscuring `malloc` / `free`:

```
  if (geom_arr == NULL) {
    PyErr_SetString(PyExc_MemoryError, "Could not allocate memory");
    return;
  }
```

jorisvandenbossche:
@brendan-ward thanks for the suggestion, added such macro!

brendan-ward (approved):

src/ufuncs.c
caspervdw: No I don't think that is an option. But at the same times, I didn't find any guarantees in numpy that this should not happen.

I see this as a forward compatibility. It is a cheap check and it may prevent segfaults when using this pygeos version with future numpy versions.

caspervdw (approved):

PR #232, opened 2020-11-06
TST: Fix make_valid tests for OverlayNG (normalize result/expected)

jorisvandenbossche:
I tested the new OverlayNG a while ago, and we only had 2 failures related to `make_valid`. But both are just a change in coordinate order and solved by using spatial equality or normalizing the resulting and expected multipolygon first.

### DIFF
--- a/pygeos/test/test_constructive.py
+++ b/pygeos/test/test_constructive.py
@@ -119,12 +119,12 @@ def test_make_valid_none():
         # an L shaped polygon without area is converted to a multilinestring
         (
             Geometry("POLYGON((0 0, 1 1, 1 2, 1 1, 0 0))"),
-            Geometry("MULTILINESTRING ((0 0, 1 1), (1 1, 1 2))"),
+            Geometry("MULTILINESTRING ((1 1, 1 2), (0 0, 1 1))"),
         ),
         # a polygon with self-intersection (bowtie) is converted into polygons
         (
             Geometry("POLYGON((0 0, 2 2, 2 0, 0 2, 0 0))"),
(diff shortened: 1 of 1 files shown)


brendan-ward:
What about updating the test fixtures so that they are pre-normalized (meaning, hard-code them in normalized form)?  Does that work on both existing GEOS and latest w/ OverlayNG?

jorisvandenbossche:
Ah, yes, I could do that indeed. Now, the actual result still needs to be normalized anyway, so it's only a small difference

jorisvandenbossche:
I started changing the expected to use the normalized form. For the first test this was easy, the second test is using `all_types` predefined geometries, and changing those causes a bunch of test failures (20+) for tests that rely on the current format. So not fully sure it is worth changing (while the current change is a one-line change)

brendan-ward:
Sorry for the poor suggestion here, didn't consider that we've hard-coded the expected forms to their non-normalized equivalents.  Updating that many definitely seems out of scope here.

Mostly was just thinking along the lines of avoiding requirement of extra processing unrelated to the function actually being tested.

Perhaps a comment here indicating that normalization is required in this case to handle variation between GEOS versions?

jorisvandenbossche:
> Sorry for the poor suggestion here, didn't consider that we've hard-coded the expected forms to their non-normalized equivalents. Updating that many definitely seems out of scope here.

No, no, not a poor suggestion ;) Ideally we would have started with normalized geometries (now I didn't yet look into it much, it might not be that hard to actually change and fix the tests)

> Mostly was just thinking along the lines of avoiding requirement of extra processing unrelated to the function actually being tested.

The `normalize` should in general not have any effect on what is actaully tested, I think. By definition, it only reorders the existing coordinate sequence, and nothing else (so spatially the geometry doesn't change)

Will add a comment!

caspervdw (approved):
Looks good!

PR #237, opened 2020-11-09
CI: Cull windows tests and shorten name

caspervdw:
Following #222 and #219, I think we should drop the older windows tests. They are covered by appveyor.

Also I shortened the name so that you can see the differences between builds in the Github status reports below a PR.


brendan-ward (approved):
Thanks @caspervdw

jorisvandenbossche (approved):

PR #239, opened 2020-11-10
CLN: remove Cythonized and compiled extension files on python setup.py clean

### LINKED ISSUE #224: Provide command (python setup.py clean?) to remove compiled files (to do a clean build)
Currently, running `python setup.py clean` removes the `/build` directory, but it further does not remove any compiled files that are in the `pygeos/` package directory itself when doing inplace builds (eg `lib.cpython-.....so`), and thus does not actually enable a "clean build".

brendan-ward:
Resolves #224


setup.py
```diff
@@ -129,7 +130,15 @@ def finalize_options(self):
 
 ext_modules = []
 
-if "clean" not in sys.argv and "sdist" not in sys.argv:
+if "clean" in sys.argv:
+    # delete any previously Cythonized or compiled files in pygeos
+    p = Path('pygeos')
+    for pattern in ['*.c','*.so', '*.pyd']:
+        for filename in p.glob(pattern):
+            print("removing '{}'".format(filename))
```
jorisvandenbossche, line 138: I added a print here, is that OK with you? 
That gives a visual indication that actually something is happening, which can be reassuring about what the command does.

jorisvandenbossche (commented):
Thanks!

caspervdw (approved):
Looks good, thanks. Could you still run black on the `setup.py` before merging?

PR #240, opened 2020-11-10
CI: move Travis linux builds to Github Actions

jorisvandenbossche:
After some trial and error fighting with github actions syntax .. (but did it on my fork), I have a working setup.

@mwtoews this is a more or less 1:1 conversion of the travis script to github actions (making use of your script to build GEOS etc)


jorisvandenbossche:
You can see how the second commit uses the cached GEOS builds of the first commit.

Building was failing for GEOS=3.5, see https://github.com/pygeos/pygeos/pull/240/checks?sha=0fc1c5113c90f98ef4016ec4550f68f8526420b6 (related to https://trac.osgeo.org/geos/ticket/784). I "fixed" that by using ubuntu-16.04, instead of the defualt "ubuntu-latest" which is 18.04 at the moment. Using 16.04 / xenial also matches what Travis was using. Once 16.04 becomes end of life, we should look into another fix (if we by that time still want to keep support for GEOS 3.5, at least).

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,92 @@
...
+          - python: 3.9
+            geos: 3.8.1
+            numpy: 1.19.4
+          # dev
+          - python: 3.9
+            geos: master
+
+    env:
+      GEOS_VERSION: ${{ matrix.geos }}
+      GEOS_INSTALL: ${{ github.workspace }}/geosinstall/geos-${{ matrix.geos }}
```
jorisvandenbossche, line 40: I wasn't able to figure out how to use $HOME here for `GEOS_INSTALL`. Now in practice, I suppose it doesn't matter that it is put into the workspace dir (that is where your repo is cloned). 
(only needed to ensure that the cache path is also using this directory)

caspervdw (approved):

mwtoews:
> You can see how the second commit uses the cached GEOS builds of the first commit.
> 
> Building was failing for GEOS=3.5 ... I "fixed" that by using ubuntu-16.04, instead of the defualt "ubuntu-latest" which is 18.04 at the moment... Once 16.04 becomes end of life, we should look into another fix (if we by that time still want to keep support for GEOS 3.5, at least).

This is a new one to me! Reading #116 and #125, Ubuntu 16.04 and GEOS 3.5 are supported until April 2021, so there's really only a few months until this specific test should retire.

mwtoews (approved):
Impressive 1:1 transition with only three attempts!

jorisvandenbossche:
> Impressive 1:1 transition with only three attempts!

I am not that a github actions expert. There were a bit more attempts on my fork .. ;-) https://github.com/jorisvandenbossche/pygeos/actions?query=workflow%3ALinux+branch%3Atravis-to-actions

PR #245, opened 2020-11-12
ENH: relate_pattern

jorisvandenbossche:


jorisvandenbossche:
I opted for now to only support a scalar pattern (so a single pattern for the full arrays of geometries), because this allows to have the python string checking and unpacking *before* the loop, and have the loop itself without the GIL.

We could also push that into the loop, or have a separate version with that checking in the loop to support array-like pattern argument as well.

src/ufuncs.c
```diff
@@ -1535,6 +1535,65 @@ static void relate_func(char** args, npy_intp* dimensions, npy_intp* steps, void
 }
 static PyUFuncGenericFunction relate_funcs[1] = {&relate_func};
 
+static char relate_pattern_dtypes[4] = {NPY_OBJECT, NPY_OBJECT, NPY_OBJECT, NPY_BOOL};
+static void relate_pattern_func(char** args, npy_intp* dimensions, npy_intp* steps,
+                                void* data) {
+  GEOSGeometry *in1 = NULL, *in2 = NULL;
+  const char* pattern;
```
brendan-ward, line 1542:
```suggestion
  const char* pattern = NULL;
```

pygeos/test/test_predicates.py
```diff
@@ -111,6 +111,38 @@ def test_relate_none(g1, g2):
...
+    assert pygeos.relate(line_string, polygon) == "11F00F212"
+    assert pygeos.relate_pattern(line_string, polygon, "11F00F212")
+    assert pygeos.relate_pattern(line_string, polygon, "*********")
+    assert not pygeos.relate_pattern(line_string, polygon, "F********")
+
+
+@pytest.mark.parametrize("g1, g2", [(point, None), (None, point), (None, None)])
+def test_relate_pattern_none(g1, g2):
+    assert pygeos.relate_pattern(g1, g2, "*" * 9).item() is False
+
```
brendan-ward, line 126: Suggestion: add a test for `pattern=None`

src/ufuncs.c
```diff
@@ -1535,6 +1535,65 @@ static void relate_func(char** args, npy_intp* dimensions, npy_intp* steps, void
...
+
+  if (is3 != 0) {
+    PyErr_Format(PyExc_ValueError, "pattern keyword only supports scalar argument");
+    return;
+  }
+  PyObject* in3 = *(PyObject**)ip3;
+  if (PyUnicode_Check(in3)) {
+    pattern = PyUnicode_AsUTF8(in3);
+    if (pattern == NULL) {
+      return;
```
brendan-ward, line 1557: This seems like it indicates an error condition, but isn't reported back as such?  Should this raise an exception?

pygeos/predicates.py
```diff
@@ -768,3 +769,33 @@ def relate(a, b, **kwargs):
...
+    """
+    Returns True if the DE-9IM string code for the relationship between
+    the geometries satisfies the pattern, else False.
+
+    This function compares the DE-9IM code string for two geometries
+    against a specified pattern. If the string matches the pattern then
+    ``True`` is returned, otherwise ``False``. The pattern specified can
+    be an exact match (``0``, ``1`` or ``2``), a boolean match
+    (``T`` or ``F``), or a wildcard (``*``). For example, the pattern
+    for the `within` predicate is ``'T*F**F***'``.
```
brendan-ward, line 785: It looks like this is case-insensitive in GEOS.  Is that correct?  If so, is it worth noting here?

pygeos/test/test_predicates.py
brendan-ward: Also consider adding tests of empty geometries.

brendan-ward (commented):
This looks straightforward to use; limiting this to scalar seems very reasonable.  Thanks @jorisvandenbossche !

src/ufuncs.c
jorisvandenbossche: My understanding is that the error is set by `PyUnicode_AsUTF8`. 
Now, I don't know what case could trigger this (or if it even possible from python), so not sure how to test it.

pygeos/predicates.py
jorisvandenbossche: Do you have an example of it being insensitive. 

Just trying the example of the docstring:

```
In [13]: relate_pattern(point, square, "T*F**F***")
Out[13]: True

In [14]: relate_pattern(point, square, "t*f**f***")
Out[14]: False
```

pygeos/test/test_predicates.py
jorisvandenbossche: Added some more tests

jorisvandenbossche:
@brendan-ward thanks for the review!

src/ufuncs.c
brendan-ward: Right
> In the case of an error, NULL is returned with an exception set

Perhaps a comment would be helpful to note that it set the exception and that `NULL` here indicates a bad string?

pygeos/predicates.py
brendan-ward: Not sure how I saw it as case insenstive now, thought I had tested but cannot reproduce now.

Might be good to note that it is case sensitive.

pygeos/test/test_predicates.py
brendan-ward: I'm assuming that changes (if any) that come out of #253 will impact the tests of `None` vs `None` here as well (since they are hardcoded to `False` here); resolving those changes would be in a different PR (if any).

pygeos/predicates.py
brendan-ward: Or always force to uppercase internally?

src/ufuncs.c
jorisvandenbossche: Added a comment

pygeos/predicates.py
jorisvandenbossche: > Or always force to uppercase internally?

Personally, I don't see much advantage in allowing the user to also pass it as lowercase. 
(except to avoid bugs maybe where you silently get always falls results? If that's the reason, we could actually also check that it is uppercase and otherwise raise an error)

jorisvandenbossche:
Given this was reviewed by Brendan (and no big items are outstanding; we can further discuss / change the case (in)sensitivity in a follow up), I am going to merge this, as I need this for work in Shapely

PR #246, opened 2020-11-12
Integrate previously prepared geometries into STRtree query / bulk query

### LINKED ISSUE #235: ENH: update strtree.query_bulk to use previously prepared geometry if available
Use `get_geom_with_prepared` to get previously prepared geometry.

If not previously prepared, prepare it on the fly.

brendan-ward:
Resolves #235 

The overall strategy is to fetch both the original and prepared geometry from the `GeometryObject`, and only prepare internally within `evaluate_predicate` if not previously prepared.

Checked with valgrind to make sure we're not leaving prepared geometries around accidentally.

I discovered and corrected an autoformatting issue with the doc string for `evaluate_predicate`, hence the larger number of lines changed (the autoformatter previously rolled all params into single paragraph, to avoid this I made them separated by lines).


caspervdw (approved):
Looks good, thanks!

Issue #244, opened 2020-11-11
Proposed: Port ST_Subdivide from PostGIS


brendan-ward:
[ST_Subdivide](https://postgis.net/docs/ST_Subdivide.html) divides complex geometries into smaller rectangular chunks of relatively similar vertex counts, which are then much better to use in spatial indexing, predicate, and related functions.  The algorithm works by recursively cutting geometries along one dimension then the other.

This is not strictly about exposing GEOS internals to Python, so it might not be a good fit here.  But given the recent work with Cython and strong dependence on internals here, it might be a worthwhile experiment (plus, I need this for one of my projects).  

Unless there was opposition, I am hoping to try and implement this within Cython here.

caspervdw:
I am not per se against it, but shouldn't this better be implemented upstream in GEOS?

jorisvandenbossche:
For reference, the PostGIS implementation is here: https://github.com/postgis/postgis/blob/6812b00ed6195e151d04be8c78aa8f1c20c628ea/liblwgeom/lwgeom.c#L2268

It would be nice to have this functionality in PyGEOS. I agree with @caspervdw it sounds as this could be potentially be shared in GEOS (but something to ask there first, then). But I would personally not be against prototyping it already in pygeos on the short term, if you want to do this.

brendan-ward:
Related GEOS ticket: https://trac.osgeo.org/geos/ticket/1075

PR #247, opened 2020-11-12
CLN: Migrate resizeable vector types and functions to dedicated files

brendan-ward:
This moves resizeable vector-related typedefs and functions to `vector.h` / `vector.c` for more generalized use elsewhere.  These currently were defined within strtree, which was their only use so far.

However, for #244 and likely other functions over the long term, we will need more general access to these so that we can build up vectors at runtime.

Types and functions renamed slightly for better clarity (I hope!).

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -11,10 +11,7 @@
 #include <numpy/ndarraytypes.h>
 #include <numpy/npy_3kcompat.h>
 
-#include "geos.h"
 #include "kvec.h"
-#include "pygeom.h"
-#include "strtree.h"
 #include "vector.h"
 
 /* Copy values from arr of indexes to a new numpy integer array.
@@ -24,8 +21,8 @@
  * arr: dynamic vector array to convert to ndarray
  */
 
-PyArrayObject* npy_intp_vec_to_npy_arr(npy_intp_vec* arr) {
-  npy_intp i;
+PyArrayObject* index_vec_to_npy_arr(index_vec_t* arr) {
+  Py_ssize_t i;
   npy_intp size = kv_size(*arr);
 
   npy_intp dims[1] = {size};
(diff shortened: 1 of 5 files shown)


brendan-ward:
Moving to draft: found a few changes needed to better align with standard naming conventions (add `_t` suffix to types).

In order to use in Cython, we also need to provide concrete types (can't pass type as a parameter to functions like `kv_push` from Cython).  I think this is solved by providing a few additional macros for each vector type we use here, which gives us an opportunity to make the functions we call named a little bit more clearly (`kv_A` is pretty non-obvious).

src/vector.h
```diff
@@ -0,0 +1,35 @@
+#ifndef _VECTOR_H
+#define _VECTOR_H
+
+#include <Python.h>
+#include <numpy/ndarraytypes.h>
+
+#include "geos.h"
```
caspervdw, line 7: Only put includes in a header if the header itself needs them. I think you should drop ` geos.h` and `kvec.h` here, only `pygeom.h` is needed for the `GeometryObject` .

src/vector.c
```diff
@@ -0,0 +1,46 @@
...
+#include <structmember.h>
+
+#define NO_IMPORT_ARRAY
+#define PY_ARRAY_UNIQUE_SYMBOL pygeos_ARRAY_API
+
+#include <numpy/arrayobject.h>
+#include <numpy/ndarraytypes.h>
+#include <numpy/npy_3kcompat.h>
+
+#include "geos.h"
```
caspervdw, line 14: There are some includes here that are not needed: I think only `kvec.h` is needed here.

caspervdw (approved):
Some minor remarks about includes, for the rest, feel free to merge this

caspervdw:
Thanks @brendan-ward

PR #252, opened 2020-11-15
ENH: Add is_prepared

### LINKED ISSUE #238: ENH: add is_prepared ufunc
Follow-on to [#92](https://github.com/pygeos/pygeos/pull/92#pullrequestreview-497817510)

The original idea was to make it easy to check if prepared geometries had already been created.

caspervdw:
Closes #238

### DIFF
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -175,7 +175,7 @@ def geometrycollections(geometries):
 
 
 def prepare(geometry, **kwargs):
-    """Add a prepared geometry to a geometry, improving performance of other operations.
+    """Prepare a geometry, improving performance of other operations.
 
     A prepared geometry is a normal geometry with added information such as an
     index on the line segments. This improves the performance of the following operations:
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -195,8 +195,8 @@ def prepare(geometry, **kwargs):
 
     See also
     --------
-    destroy_prepared
-    is_prepared
+    is_prepared : Identify whether a geometry is prepared already.
+    destroy_prepared : Destroy the prepared part of a geometry.
     """
     lib.prepare(geometry, **kwargs)
 
@@ -216,6 +216,5 @@ def destroy_prepared(geometry, **kwargs):
     See also
     --------
     prepare
-    is_prepared : Identify whether a geometry is prepared
     """
     lib.destroy_prepared(geometry, **kwargs)
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -175,7 +175,7 @@ def geometrycollections(geometries):
 
 
 def prepare(geometry, **kwargs):
-    """Compute a prepared geometry.
+    """Add a prepared geometry to a geometry, improving performance of other operations.
 
     A prepared geometry is a normal geometry with added information such as an
     index on the line segments. This improves the performance of the following operations:
@@ -191,16 +191,18 @@ def prepare(geometry, **kwargs):
     Parameters
     ----------
     geometry : Geometry or array_like
+        Geometries are changed inplace
 
     See also
     --------
     destroy_prepared
+    is_prepared
     """
-    return lib.prepare(geometry, **kwargs)
+    lib.prepare(geometry, **kwargs)
 
 
 def destroy_prepared(geometry, **kwargs):
-    """Destroy a previously prepared geometry, freeing up memory.
+    """Destroy the prepared part of a geometry, freeing up memory.
 
     Note that the prepared geometry will always be cleaned up if the geometry itself
     is dereferenced. This function needs only be called in very specific circumstances,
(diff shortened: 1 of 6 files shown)


pygeos/predicates.py
```diff
@@ -186,6 +187,32 @@ def is_missing(geometry, **kwargs):
...
+    --------
+    is_valid_input : check if an object is a geometry or None
+    prepare : prepare a geometry
+
+    Examples
+    --------
+    >>> geometry = Geometry("POINT (0 0)")
+    >>> is_prepared(Geometry("POINT (0 0)"))
+    False
+    >>> pygeos.prepare(geometry);
```
brendan-ward, line 209: To the top of the file, add
```
from . import prepare  # NOQA
```

```suggestion
    >>> prepare(geometry)
```

pygeos/test/test_predicates.py
```diff
@@ -132,6 +133,15 @@ def test_is_ccw(geom, expected):
 @pytest.mark.parametrize("func", BINARY_PREPARED_PREDICATES)
 def test_binary_prepared(a, func):
     actual = func(a, point)
-    pygeos.lib.prepare(a)
+    pygeos.prepare(a)
     result = func(a, point)
     assert actual == result
+
+
+def test_is_prepared():
```
brendan-ward, line 141: Suggestion: add tests for `None` and an empty geometry.

pygeos/predicates.py
```diff
@@ -186,6 +187,32 @@ def is_missing(geometry, **kwargs):
     """
     return lib.is_missing(geometry, **kwargs)
 
+
+@multithreading_enabled
+def is_prepared(geometry, **kwargs):
+    """Returns True if a Geometry is prepared
```
brendan-ward, line 193:
```suggestion
    """Returns True if a Geometry is prepared.  Will always return False if Geometry is None.
```

pygeos/predicates.py
```diff
@@ -186,6 +187,32 @@ def is_missing(geometry, **kwargs):
...
+
+    Examples
+    --------
+    >>> geometry = Geometry("POINT (0 0)")
+    >>> is_prepared(Geometry("POINT (0 0)"))
+    False
+    >>> pygeos.prepare(geometry);
+    >>> is_prepared(geometry)
+    True
+    """
```
brendan-ward, line 212: Suggestion: add example for `None`

```
    >>> geometry = None
    >>> prepare(geometry)
    >>> is_prepared(geometry)
    False
```

brendan-ward (changes requested):
Thanks @caspervdw .  A couple minor fixes needed to handle imports so that doctests pass, and a few suggestions about adding doctest / test entries for `None` and empty geometries.

pygeos/creation.py
```diff
@@ -175,7 +175,7 @@ def geometrycollections(geometries):
 
 
 def prepare(geometry, **kwargs):
-    """Compute a prepared geometry.
+    """Add a prepared geometry to a geometry, improving performance of other operations.
```
jorisvandenbossche, line 178:
```suggestion
    """Prepare a geometry, improving performance of other operations.
```

I find "adding" a bit strange, since it's an implementation detail that it is added to the C struct?

pygeos/predicates.py
jorisvandenbossche: Should we actually return True instead? 
Because why would you use `is_prepared`? To know whether you would still call `prepare`? In that case, doing `is_prepared(geoms).all()` would be useful, but doesn't work with missing values

jorisvandenbossche (commented):
In general, I am wondering a bit what the practical use of this function would be. 

Certainly, in theory it is nice that you can check if geometries are prepared. But would you also use that in practice? Because if you use it because you want to have prepared geometries, but want to check if you need to call `prepare`, it might be faster to simply call `prepare` because the you only need a single pass over the geometries, and we don't re-prepare the ones that are already prepared. 

Now, for testing purposes it can certainly be handy, so that's probably enough reason to add it. But maybe we should point people to simply using `prepare`

caspervdw:
> Should we actually return True instead?
Because why would you use is_prepared? To know whether you would still call prepare? In that case, doing is_prepared(geoms).all() would be useful, but doesn't work with missing values

I think it is best to stick to the general pattern of returning `False` on unary predicates on missing geometries. `None` can't be prepared, so `is_prepared(None)` should be False.

> Now, for testing purposes it can certainly be handy, so that's probably enough reason to add it. But maybe we should point people to simply using prepare

Agreed, I added a note in the docstring.

pygeos/predicates.py
```diff
@@ -187,6 +188,37 @@ def is_missing(geometry, **kwargs):
     """
     return lib.is_missing(geometry, **kwargs)
 
+
+@multithreading_enabled
+def is_prepared(geometry, **kwargs):
+    """Returns True if a Geometry is prepared.
+
+    Note that it is not necessary to check if a geometry is already prepared
+    before preparing it. It is more efficient to call ``prepare`` directly.
```
jorisvandenbossche, line 197: maybe further add something like "... because it will skip geometries that are already prepared"

jorisvandenbossche (approved):
Thanks for the update! Just a tiny doc comment

jorisvandenbossche:
> I think it is best to stick to the general pattern of returning False on unary predicates on missing geometries. None can't be prepared, so is_prepared(None) should be False.

Yeah, I agree that this consistency is good (it just makes the function even less practical to use, although one can probably say the same for other predicates like `is_valid`, where you also need to take missing values into account)

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

Issue #255, opened 2020-11-17
Shims required to support GEOS version-specific functions in Cython 


brendan-ward:
The standard approach of declaring an entry for a GEOS function in `_geos.pxd` does not work when the function is not available in a given older version of GEOS.

In the C extension, we are able to use macros like `GEOS_SINCE_3_7_0` to conditionally include these for the compiler.

I don't see a similar mechanism for Cython.

It looks like other projects (e.g,. Fiona) deal with this by defining shim `*.pxd` files that are dynamically selected and used for the build based on the targeted version of the C library.  I'm not yet sure how those handle functions that were introduced into later versions of the C library.

Is there a better alternative approach?

snowman2:
https://github.com/mapbox/rasterio/pull/2016

snowman2:
Another example: https://github.com/Toblerity/Fiona/pull/979

brendan-ward:
Thanks @snowman2 !

One challenge here compared to rasterio / Fiona is that we don't require an environment variable indicating GEOS version when `geos-config` is not available (e.g., on Windows?).  However, it might be a good idea for us to do so in order to check and enforce minimum GEOS version, and then using `compile_time_env` in `setup.py` becomes straightforward.

PR #256, opened 2020-11-18
WIP: add subdivide function in Cython

### LINKED ISSUE #244: Proposed: Port ST_Subdivide from PostGIS

brendan-ward:
Resolves #244 

This adds an adaption of the algorithm in ST_Subdivide.  It recursively subdivides an ndarray until reaching a target number of vertices.  It uses a quad tile approach vs horizontal or vertical slice (whichever is longer in a given step) as used in ST_Subdivide.  Haven't yet tested which has better performance, but there is a benchmark here so we can track that.

This includes some minor changes I need to backport to #247 around the C vectors.

The actual subdivide function itself is not that complex, but this involved adding a lot of supporting code: we needed several GEOS functions in Cython, and needed a few basic utility functions (`_bounds`, `_box`) that we needed here in a way that couldn't reuse the ufuncs (so they can be within `nogil` blocks / functions).

This currently uses GEOS functions availabe in 3.7 and 3.8, which will require an approach worked out in #255 

Some of the biggest additions here, and most complex to work out, were resizeable vectors that were more obvious / easy to use in Cython.  We couldn't use the existing C vector approach as-is for certain functions because we needed to lock in specific type bindings in C rather than in Python.  This then led to either creating type-specific functions for each vector type (e.g., `index_vec_init`, `index_vec_push`, etc), or a mess of original functions like `kv_init` and type specific functions to bind in the type.  It also felt less friendly to use in Cython, and having a context manager to release the vectors seemed like a good idea.

These still need a bit more work, esp. to release `GEOSGeometries` when manually clearing out the `GeometryVector` in case of an error.


brendan-ward:
Benchmarks for splitting into 4 tiles per recursion vs bisecting longest dimension:
4 tiles:  866±10ms
bisect:  933±20ms

likely highly dependent on the spatial configuration of vertices.  Vertices in the benchmark generated geometries are relatively evenly distributed.

A more optimal approach could use median points rather than means (center x,y) so that there are roughly equal point densities in each quadrant, but there is some overhead to sorting points.  Optimal in this case also depends on how unevenly the points are distributed.

jorisvandenbossche:
Thanks a lot for working on this! 
I didn't yet look at the actual code, but just one remark, or rather concern: this seems to be adding quite some complexity to the cython side (and some duplication of things that are also implemented already in C), which *could* be a sign that this should rather be done in C?  (eg we could say that if you need the resizable vector, that we do such algos in C instead of cython)

(to be clear, I don't have the answer (it might well be that we want this), just noting this is probably something we need to consider / discuss how far we want to go in cython in general)

brendan-ward:
Agreed on the concerns about complexity, and not being sure just how far we want to reimplement things in Cython.

If we kept this in C _and_ modularized some of the shared functions (bounds, box) so that they could be used outside of ufuncs, then that would cut down on reimplementation in Cython a bit.

Most of what is here should port over to C pretty easily; it just involves a bit more overhead (typing effort) to make the function available for use in Python.  I might try that (in a different PR), to see how the two approaches compare.

brendan-ward:
Closing since we did not end up implementing this here; may eventually be implemented in GEOS or Shapely.

PR #257, opened 2020-11-22
ENH: Add get / set precision ufuncs

brendan-ward:
This adds get / set precision functions.

A few things to be aware of:
The Default value for `preserve_topology` is `False` to match simplify function; however, in GEOS it defaults to `True`; the choice was out of consistency, but it may well be appropriate to match GEOS.

There may be a couple GEOS bugs discovered in the tests here: 
1. geometrically invalid inputs do not behave as expected with `preserve_topology = True` (we get back only one of the polygons in the expected multipolygon).
2. polygons always collapse to empty polygons regardless of `keep_collapsed` even though linear rings are preserved.

Will need to dig into GEOS a bit more and log bugs there if appropriate.

I did not yet add additional tests to trace the precision through intersection, etc operations to verify the comment (adapted from GEOS) that the highest precision is used and assigned to outputs of those operations.

### DIFF
--- a/pygeos/geometry.py
+++ b/pygeos/geometry.py
@@ -546,8 +546,8 @@ def get_num_geometries(geometry):
 def get_precision(geometry):
     """Get the precision of a geometry.
 
-    If a precision has not been previously set, it will be 0 (floating point precision).
-    Otherwise, it will return the grid size that was set on a geometry.
+    If a precision has not been previously set, it will be 0 (double precision).
+    Otherwise, it will return the precision grid size that was set on a geometry.
 
     Returns NaN for not-a-geometry values.
 
@@ -561,11 +561,11 @@ def get_precision(geometry):
 
     Examples
     --------
-    >>> get_precision(Geometry("POINT (1 1)")) == 0
-    True
-    >>> geometry = set_precision(Geometry("POINT (1 1)"), 1)
-    >>> get_precision(geometry) == 1
-    True
+    >>> get_precision(Geometry("POINT (1 1)"))
+    0.0
+    >>> geometry = set_precision(Geometry("POINT (1 1)"), 1.0)
+    >>> get_precision(geometry)
+    1.0
     >>> np.isnan(get_precision(None))
     True
     """
@@ -574,8 +574,10 @@ def get_precision(geometry):
 
 @requires_geos("3.6.0")
 @multithreading_enabled
-def set_precision(geometry, grid_size=0, preserve_topology=False):
+def set_precision(geometry, grid_size, preserve_topology=False):
     """Returns geometry with the precision set to a precision grid size.
+
+    By default, geometries use double precision coordinates (grid_size = 0).
 
     Coordinates will be rounded if a precision grid is less precise than the input
     geometry. Duplicated vertices will be dropped from lines and polygons for grid sizes greater
@@ -594,8 +596,10 @@ def set_precision(geometry, grid_size=0, preserve_topology=False):
     Parameters
     ----------
     geometry : Geometry or array_like
-    grid_size : double, optional (default: 0)
-        precision grid size.  If 0, will use floating point precision.
+    grid_size : double
+        precision grid size.  If 0, will use double precision (will not modify geometry
+        if precision grid size was not previously set).  If this value is more
+        precise than input geometry, the input geometry will not be modified.
     preserve_topology : bool, optional (default: False)
         If True, will attempt to preserve the topology of a geometry after rounding
         coordinates.
@@ -606,13 +610,13 @@ def set_precision(geometry, grid_size=0, preserve_topology=False):
 
     Examples
     --------
-    >>> set_precision(Geometry("POINT (0.9 0.9)"), 1)
+    >>> set_precision(Geometry("POINT (0.9 0.9)"), 1.0)
     <pygeos.Geometry POINT (1 1)>
-    >>> set_precision(Geometry("POINT (0.9 0.9 0.9)"), 1)
+    >>> set_precision(Geometry("POINT (0.9 0.9 0.9)"), 1.0)
     <pygeos.Geometry POINT Z (1 1 0.9)>
-    >>> set_precision(Geometry("LINESTRING (0 0, 0 0.1, 0 1, 1 1)"), 1)
+    >>> set_precision(Geometry("LINESTRING (0 0, 0 0.1, 0 1, 1 1)"), 1.0)
     <pygeos.Geometry LINESTRING (0 0, 0 1, 1 1)>
-    >>> set_precision(None) is None
+    >>> set_precision(None, 1.0) is None
     True
     """
 
--- a/pygeos/geometry.py
+++ b/pygeos/geometry.py
@@ -546,8 +546,9 @@ def get_num_geometries(geometry):
 def get_precision(geometry):
     """Get the precision of a geometry.
 
-    If a precision has not been previously set, it will be 0 (double precision).
-    Otherwise, it will return the precision grid size that was set on a geometry.
+    If a precision has not been previously set, it will be 0 (double
+    precision). Otherwise, it will return the precision grid size that was
+    set on a geometry.
 
     Returns NaN for not-a-geometry values.
 
@@ -579,30 +580,32 @@ def set_precision(geometry, grid_size, preserve_topology=False):
 
     By default, geometries use double precision coordinates (grid_size = 0).
 
-    Coordinates will be rounded if a precision grid is less precise than the input
-    geometry. Duplicated vertices will be dropped from lines and polygons for grid sizes greater
-    than 0. Line and polygon geometries may collapse to empty geometries if all vertices are
-    closer together than grid_size. Z values, if present, will not be modified.
+    Coordinates will be rounded if a precision grid is less precise than the
+    input geometry. Duplicated vertices will be dropped from lines and
+    polygons for grid sizes greater than 0. Line and polygon geometries may
+    collapse to empty geometries if all vertices are closer together than
+    grid_size. Z values, if present, will not be modified.
 
-    Note: subsequent operations will always be performed in the precision
-    of the geometry with higher precision (smaller "grid_size"). That same precision will
-    be attached to the operation outputs.
+    Note: subsequent operations will always be performed in the precision of
+    the geometry with higher precision (smaller "grid_size"). That same
+    precision will be attached to the operation outputs.
 
-    Also note: input geometries should be geometrically valid; unexpected results may
-    occur if input geometries are not.
+    Also note: input geometries should be geometrically valid; unexpected
+    results may occur if input geometries are not.
 
     Returns None if geometry is None.
 
     Parameters
     ----------
     geometry : Geometry or array_like
-    grid_size : double
-        precision grid size.  If 0, will use double precision (will not modify geometry
-        if precision grid size was not previously set).  If this value is more
-        precise than input geometry, the input geometry will not be modified.
+    grid_size : float
+        Precision grid size. If 0, will use double precision (will not modify
+        geometry if precision grid size was not previously set). If this
+        value is more precise than input geometry, the input geometry will
+        not be modified.
     preserve_topology : bool, optional (default: False)
-        If True, will attempt to preserve the topology of a geometry after rounding
-        coordinates.
+        If True, will attempt to preserve the topology of a geometry after
+        rounding coordinates.
 
     See also
     --------
--- a/pygeos/geometry.py
+++ b/pygeos/geometry.py
@@ -579,6 +579,9 @@ def set_precision(geometry, grid_size=0, preserve_topology=False):
     coordinates in the geometry to the precision grid if less precise than the input
     geometry.
 
+    Duplicated vertices will be dropped from lines and polygons for grid sizes greater
+    than 0.
+
     Line and polygon geometries may collapse to empty geometries if all vertices are
     closer together than grid_size.
 
@@ -612,9 +615,10 @@ def set_precision(geometry, grid_size=0, preserve_topology=False):
     <pygeos.Geometry POINT (1 1)>
     >>> set_precision(Geometry("POINT (0.9 0.9 0.9)"), 1)
     <pygeos.Geometry POINT Z (1 1 0.9)>
+    >>> set_precision(Geometry("LINESTRING (0 0, 0 0.1, 0 1, 1 1)"), 1)
+    <pygeos.Geometry LINESTRING (0 0, 0 1, 1 1)>
     >>> set_precision(None) is None
     True
     """
 
     return lib.set_precision(geometry, grid_size, preserve_topology)
-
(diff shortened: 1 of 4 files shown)


brendan-ward:
Logged GEOS bugs:
* https://trac.osgeo.org/geos/ticket/1071
* https://trac.osgeo.org/geos/ticket/1072

brendan-ward:
I think this is a case where we shouldn't expose all parameters from GEOS.

In this case, setting the flag to keep collapsed geometries inconsistently produces topologically invalid geometries.  Per [comments in GEOS Trac](https://trac.osgeo.org/geos/ticket/1071#comment:1), they should always collapse, and will always collapse as of GEOS >= 3.9.

I can't think of a good reason why we would want to keep collapsed invalid geometries here, so I stripped this out.

Per the other GEOS issue, we [shouldn't expect invalid geometries to produce valid outputs](https://trac.osgeo.org/geos/ticket/1072#comment:1); hopefully the comment in the docstring is sufficient to indicate this.

src/ufuncs.c
```diff
@@ -862,7 +862,19 @@ static void* get_z_data[1] = {GetZ};
 #endif
 static void* area_data[1] = {GEOSArea_r};
 static void* length_data[1] = {GEOSLength_r};
+
 #if GEOS_SINCE_3_6_0
+static int GetPrecision(void* context, void* a, double* b) {
+  // GEOS returns -1 on error; 0 indicates floating point precision
+  double out = GEOSGeom_getPrecision_r(context, a);
+  if (out == -1) {
+    *(double*)b = NPY_NAN;
```
caspervdw, line 871: If an error is raised, there is no need to output anything

src/ufuncs.c
```diff
@@ -1595,6 +1607,63 @@ static void relate_pattern_func(char** args, npy_intp* dimensions, npy_intp* ste
...
+    }
+    // grid size
+    double in2 = *(double*)ip2;
+    // preserve topology
+    npy_bool in3 = *(npy_bool*)ip3;
+    // flags:
+    // GEOS_PREC_NO_TOPO (1<<0): if set, do not try to preserve topology
+    // GEOS_PREC_KEEP_COLLAPSED  (1<<1): Not used because uncollapsed geometries are
+    // invalid and will not be retained in GEOS >= 3.9 anyway.
+    int flags = (!in3) << 0;
```
caspervdw, line 1640: As the geos c api defines these as macros, it would be more readable to do something like:

```
int flags = 0 ? in3 : GEOS_PREC_NO_TOPO;
```

pygeos/test/test_geometry.py
```diff
@@ -340,26 +340,147 @@ def test_get_parts_invalid_dimensions(geom):
...
+    assert np.all(np.isnan(pygeos.get_precision([None])))
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+def test_set_precision():
+    initial_geometry = pygeos.Geometry("POINT (0.9 0.9)")
+    assert pygeos.get_precision(initial_geometry) == 0
+
+    geometry = pygeos.set_precision(initial_geometry, 0)
+    print("geometry", geometry)
```
caspervdw, line 392: Should be cleaned up

pygeos/geometry.py
```diff
@@ -537,3 +539,86 @@ def get_num_geometries(geometry):
     0
     """
     return lib.get_num_geometries(geometry)
+
+
+@requires_geos("3.6.0")
+@multithreading_enabled
+def get_precision(geometry):
+    """Get the precision of the geometry.
```
caspervdw, line 547:
```suggestion
    """Get the precision of a geometry.
```

pygeos/geometry.py
```diff
@@ -537,3 +539,86 @@ def get_num_geometries(geometry):
...
+
+    If the precision has not been previously set, it will be 0 (floating point precision).
```
caspervdw, line 549:
```suggestion
    If a precision has not been previously set, it will be 0 (floating point precision).
```

pygeos/geometry.py
```diff
@@ -537,3 +539,86 @@ def get_num_geometries(geometry):
...
+    >>> np.isnan(get_precision(None))
+    True
+    """
+    return lib.get_precision(geometry)
+
+
+@requires_geos("3.6.0")
+@multithreading_enabled
+def set_precision(geometry, grid_size=0, preserve_topology=False):
+    """Returns geometry with the precision set to the precision grid size, rounding all
```
caspervdw, line 578: This line is too long, the 'short description' should fit on 1 line.

Also I'd like to keep the option open for plural goemetry input. So instead of "Set the precision of the Geometry" I would like "Set the precision of a Geometry".
- I didn't look at the tests, in general I try to stick with the 1-assert-per-unittest rule, of course a few are not a problem, but these have very many... are some print functions in the tests.

caspervdw (changes requested):
@brendan-ward  Thanks for working on this! Not sure if I am "too early" with the review (you didn't request one yet), but I started on it anyway.

One general remark on unittests:
- I didn't look at the tests to great lengths. In general I try to stick with the 1-assert-per-unittest rule. One unittest should test 1 thing. It is not necessary to test the same thing twice. It's not a big deal (I am fine with leaving it like this), but I thought it is worth mentioning as a generic testing pattern.

src/ufuncs.c
brendan-ward: I think you meant `int flags = in3 ? 0 : GEOS_PREC_NO_TOPO;`

brendan-ward:
@caspervdw Thanks for the review!

I split the tests up a bit more; I think this eliminates the testing of different conditions in a single test.  Because they were stateful, it was helpful to have a few asserts in a single test: assert default condition, assert that it changed, assert that the original geometry was not modified.  There are a couple asserts with `get_precision` after a `set_precision` that may not be strictly necessary since that is also tested in the `test_get_precision`, but I thought it might be helpful to verify expected state.  Certainly open to cutting those back more if this still seems like too many assertions per test.

pygeos/geometry.py
```diff
@@ -537,3 +539,81 @@ def get_num_geometries(geometry):
...
+    geometry : Geometry or array_like
+
+    See also
+    --------
+    set_precision
+
+    Examples
+    --------
+    >>> get_precision(Geometry("POINT (1 1)")) == 0
+    True
```
caspervdw, line 565: Is there a reason for testing `== 0` returns `True`? And not

```
>>> get_precision(...)
0.0
```

pygeos/geometry.py
```diff
@@ -537,3 +539,81 @@ def get_num_geometries(geometry):
...
+    >>> geometry = set_precision(Geometry("POINT (1 1)"), 1)
+    >>> get_precision(geometry) == 1
```
caspervdw, line 567: Same question as above.

pygeos/geometry.py
```diff
@@ -537,3 +539,81 @@ def get_num_geometries(geometry):
...
+    True
+    >>> np.isnan(get_precision(None))
+    True
+    """
+    return lib.get_precision(geometry)
+
+
+@requires_geos("3.6.0")
+@multithreading_enabled
+def set_precision(geometry, grid_size=0, preserve_topology=False):
```
caspervdw, line 577: I don't think it make sense to provide `0` as a default, because nothing will really happen if you do `grid_size=0`.

pygeos/geometry.py
```diff
@@ -537,3 +539,81 @@ def get_num_geometries(geometry):
...
+    Also note: input geometries should be geometrically valid; unexpected results may
+    occur if input geometries are not.
+
+    Returns None if geometry is None.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+    grid_size : double, optional (default: 0)
+        precision grid size.  If 0, will use floating point precision.
```
caspervdw, line 598: Can you specify single or double precision here?

pygeos/test/test_geometry.py
```diff
@@ -340,26 +340,157 @@ def test_get_parts_invalid_dimensions(geom):
...
+    geometry = pygeos.set_precision(initial_geometry, 1)
+    assert pygeos.get_precision(geometry) == 1
+    assert pygeos.equals(geometry, pygeos.Geometry("POINT (1 1)"))
+    # original should remain unchanged
+    assert pygeos.equals(initial_geometry, pygeos.Geometry("POINT (0.9 0.9)"))
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+def test_set_precision_drop_coords():
+    # setting precision of 0 will not drop duplicated points in original
```
caspervdw, line 406: This behavior does not match what is described in the docstring "If 0, will use floating point precision." 

If floating point precision (1E-7 for single or 1E-14 for double) would have been used, the two equal points would be merged into one.

I think the docstring needs adjustment: 0 means, do not change the geometry.

pygeos/test/test_geometry.py
```diff
@@ -340,26 +340,157 @@ def test_get_parts_invalid_dimensions(geom):
...
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+def test_set_precision_z():
+    geometry = pygeos.set_precision(pygeos.Geometry("POINT Z (0.9 0.9 0.9)"), 1)
+    assert pygeos.get_precision(geometry) == 1
+    assert pygeos.equals(geometry, pygeos.Geometry("POINT Z (1 1 0.9)"))
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+def test_set_precision_nan():
```
caspervdw, line 425: This name clashes with another test.

```suggestion
def test_set_precision_nan_coordinates():
```

pygeos/geometry.py
```diff
@@ -537,3 +539,81 @@ def get_num_geometries(geometry):
...
+        If True, will attempt to preserve the topology of a geometry after rounding
+        coordinates.
+
+    See also
+    --------
+    get_precision
+
+    Examples
+    --------
+    >>> set_precision(Geometry("POINT (0.9 0.9)"), 1)
```
caspervdw, line 609: This may seem like bikeshedding, but I think it is important to signal that there is `float` expected here. It also avoids a type conversion to `float` if you give a float default.

```suggestion
    >>> set_precision(Geometry("POINT (0.9 0.9)"), 1.0)
```

Same suggestion for the default value in the function signature and docstring

pygeos/test/test_geometry.py
```diff
@@ -340,26 +340,157 @@ def test_get_parts_invalid_dimensions(geom):
...
+    assert pygeos.equals(
+        pygeos.set_precision(geometry, 5, preserve_topology=True),
+        pygeos.Geometry(
+            "MULTIPOLYGON (((10 10, 15 15, 20 10, 10 10)), ((15 15, 10 20, 20 20, 15 15)))"
+        ),
+    )
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+def test_set_precision_collapse():
```
caspervdw, line 461: This test should be parametrized with `@pytest.mark.parametrize("actual,expected", ...)`. The 3 cases are independent.

pygeos/test/test_geometry.py
```diff
@@ -340,26 +340,157 @@ def test_get_parts_invalid_dimensions(geom):
...
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+def test_set_precision_intersection():
+    """Operations should use the highest precision of the inputs"""
+
+    box1 = pygeos.normalize(pygeos.box(0, 0, 0.9, 0.9))
+    box2 = pygeos.normalize(pygeos.box(0.75, 0, 1.75, 0.75))
+
+    assert pygeos.get_precision(pygeos.intersection(box1, box2)) == 0
+
+    # GEOS will use and keep the lowest precision
```
caspervdw, line 491:
```suggestion
    # GEOS will use and keep the lowest (most precise) value
```

pygeos/test/test_geometry.py
```diff
@@ -340,26 +340,157 @@ def test_get_parts_invalid_dimensions(geom):
...
+    geometry = pygeos.Geometry("POLYGON ((0 0, 0.1 0, 0.1 0.1, 0 0.1, 0 0))")
+    assert pygeos.equals(
+        pygeos.set_precision(geometry, 1),
+        pygeos.Geometry("POLYGON EMPTY"),
+    )
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+def test_set_precision_intersection():
+    """Operations should use the highest precision of the inputs"""
```
caspervdw, line 484:
```suggestion
    """Operations should use the most precise (lowest) input precision"""
```

src/ufuncs.c
```diff
@@ -862,7 +862,18 @@ static void* get_z_data[1] = {GetZ};
 #endif
 static void* area_data[1] = {GEOSArea_r};
 static void* length_data[1] = {GEOSLength_r};
+
 #if GEOS_SINCE_3_6_0
+static int GetPrecision(void* context, void* a, double* b) {
+  // GEOS returns -1 on error; 0 indicates floating point precision
```
caspervdw, line 868: doubtful if it indicates floating point precision (see previous remark) - if you agree please change also here

caspervdw (approved):
Quite some non-critical changes, please have a look and then feel free to merge

pygeos/geometry.py
brendan-ward: Removed default value, now will be required parameter.

pygeos/test/test_geometry.py
brendan-ward: I think it is a bit more general than that; if you provide a more precise grid size than the grid size previously assigned to the geometry, the geometry should not be modified.  This could happen if you first set a precision of 1, then set 0.1.

I'll try to make this more clear in the docstring.

jorisvandenbossche (approved):
Looks good! Only some tiny nitpicks

jorisvandenbossche:
@brendan-ward I took the liberty to push some edits for my review myself (it were only minor nitpicks), and also limited the line length of the docstrings a bit (black doesn't do this). 

Since Casper already approved as well, going to merge this.

jorisvandenbossche:
@brendan-ward coming back to one aspect of this that you raised in the top post:

> The Default value for preserve_topology is False to match simplify function; however, in GEOS it defaults to True; the choice was out of consistency, but it may well be appropriate to match GEOS.

Should we maybe reconsider this? It's indeed nice to have it consistent with other functions like `simplify`, but there might also be good reasons that GEOS has it on by default. It _feels_ the safer option in general for users ("safer but slower"), so in that sense it would maybe be good as default and then users who know what they are doing can turn it off to get fast / less safe.

brendan-ward:
> Should we maybe reconsider this?

Sure - I could go either way with the default.  Does this need to get fixed before 0.9 is released?

/cc @caspervdw

Issue #233, opened 2020-11-06
Failures with GEOS master / 3.9


jorisvandenbossche:
I was testing with GEOS master (in light of the new OverlayNG), and next to 2 failures related to the new overlay implementation (which I am fixing in https://github.com/pygeos/pygeos/pull/232), there are also 2 other failures:

```
_______________________________________ test_points_invalid_ndim ________________________________________

    def test_points_invalid_ndim():
        with pytest.raises(pygeos.GEOSException):
>           pygeos.points([0, 1, 2, 3])
E           Failed: DID NOT RAISE <class 'pygeos.GEOSException'>

pygeos/test/test_creation.py:41: Failed
_______________________________________ test_from_wkb_empty[POLYGON EMPTY] _____________________________

wkt = 'POLYGON EMPTY'

    @pytest.mark.parametrize(
        "wkt", ("POINT EMPTY", "LINESTRING EMPTY", "POLYGON EMPTY", "GEOMETRYCOLLECTION EMPTY")
    )
    def test_from_wkb_empty(wkt):
        wkb = pygeos.to_wkb(pygeos.Geometry(wkt))
        geom = pygeos.from_wkb(wkb)
        assert pygeos.is_geometry(geom).all()
        assert pygeos.is_empty(geom).all()
>       assert pygeos.to_wkb(geom) == wkb
E       AssertionError: assert b'\x01\x03\x0...0\x00\x00\x00' == b'\x01\x03\x0...0\x00\x00\x00'
E         At index 4 diff: b'\x80' != b'\x00'
E         Use -v to get the full diff

pygeos/test/test_io.py:121: AssertionError
```

The first case now seems to ignore the 4th dimension, and create a POINT Z:

```
(Pdb) pygeos.points([0, 1, 2, 3])
<pygeos.Geometry POINT Z (0 1 2)>
```

In the second case, the WKB of an empty polygon differs depending on whether the empty polygon was created from WKT or WKB:

```
(Pdb) geom1 = pygeos.Geometry("POLYGON EMPTY")
(Pdb) geom1
<pygeos.Geometry POLYGON EMPTY>
(Pdb) pygeos.to_wkb(geom1)
b'\x01\x03\x00\x00\x00\x00\x00\x00\x00'
(Pdb) geom2 = pygeos.from_wkb(pygeos.to_wkb(geom1))
(Pdb) geom2
<pygeos.Geometry POLYGON EMPTY>
(Pdb) pygeos.to_wkb(geom2)
[...]

PR #261, opened 2020-11-29
Fix STRtree insert/query to avoid void nullptr

jorisvandenbossche:
xref https://github.com/pygeos/pygeos/issues/233#issuecomment-735285534 and the discussion the geos-devel mailing list (https://lists.osgeo.org/pipermail/geos-devel/2020-November/009870.html). 

This change is what I understand from that discussion should be the solution, but this is not actually working (it's giving garbage / crashing).

One thing I am wondering: I am now passing the address of an integer, but this int is a loop variable, so that is only a local variable that I suppose is cleaned up at the end of `STRtree_new`. 
So if we want to be able to access those integers in `STRtree_query` we might need to store them on the STRtree object? (the python wrapper object, not the GEOS STRtree).

### DIFF
--- a/src/strtree.c
+++ b/src/strtree.c
@@ -172,7 +172,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
       Py_INCREF(obj);
       kv_push(GeometryObject*, _geoms, obj);
       count++;
-      GEOSSTRtree_insert_r(ctx, tree, geom, (void*)i);
+      GEOSSTRtree_insert_r(ctx, tree, geom, &i);
     }
   }
 
@@ -199,7 +199,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
  * */
 
 void query_callback(void* item, void* user_data) {
-  kv_push(npy_intp, *(npy_intp_vec*)user_data, (npy_intp)item);
+  kv_push(npy_intp, *(npy_intp_vec*)user_data, *(npy_intp*)item);
 }
 
 /* Evaluate the predicate function against a prepared version of geom
--- a/src/strtree.c
+++ b/src/strtree.c
@@ -104,6 +104,7 @@ static void STRtree_dealloc(STRtreeObject* self) {
     Py_XDECREF(kv_A(self->_geoms, i));
   }
   kv_destroy(self->_geoms);
+  kv_destroy(self->_tree_indexes);
   // free the PyObject
   Py_TYPE(self)->tp_free((PyObject*)self);
 }
@@ -115,6 +116,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
   npy_intp n, i, count = 0;
   GEOSGeometry* geom;
   pg_geom_obj_vec _geoms;
+  npy_intp_vec _tree_indexes;
   GeometryObject* obj;
 
   if (!PyArg_ParseTuple(args, "Oi", &arr, &node_capacity)) {
@@ -145,7 +147,11 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
   n = PyArray_SIZE((PyArrayObject*)arr);
 
   kv_init(_geoms);
+  kv_init(_tree_indexes);
+  kv_resize(npy_intp, _tree_indexes, n);
+
   for (i = 0; i < n; i++) {
+    kv_push(npy_intp, _tree_indexes, i);
     /* get the geometry */
     ptr = PyArray_GETPTR1((PyArrayObject*)arr, i);
     obj = *(GeometryObject**)ptr;
@@ -159,6 +165,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
         Py_XDECREF(kv_A(_geoms, i));
       }
       kv_destroy(_geoms);
+      kv_destroy(_tree_indexes);
       GEOS_FINISH;
       return NULL;
     }
@@ -172,7 +179,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
       Py_INCREF(obj);
       kv_push(GeometryObject*, _geoms, obj);
       count++;
-      GEOSSTRtree_insert_r(ctx, tree, geom, &i);
+      GEOSSTRtree_insert_r(ctx, tree, geom, &kv_A(_tree_indexes, i));
     }
   }
 
@@ -186,6 +193,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
   self->ptr = tree;
   self->count = count;
   self->_geoms = _geoms;
+  self->_tree_indexes = _tree_indexes;
   return (PyObject*)self;
 }
 
--- a/src/strtree.h
+++ b/src/strtree.h
@@ -22,6 +22,7 @@ typedef struct {
   PyObject_HEAD void* ptr;
   npy_intp count;
   pg_geom_obj_vec _geoms;
+  npy_intp_vec _tree_indexes;
 } STRtreeObject;
 
 extern PyTypeObject STRtreeType;


dbaston:
Right, you don't want to keep a pointer to a local variable that has gone out-of-scope. You can see how it's handled in `sf` here: https://github.com/r-spatial/sf/blob/master/src/geos.cpp#L398

src/strtree.c
```diff
@@ -172,7 +179,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
       Py_INCREF(obj);
       kv_push(GeometryObject*, _geoms, obj);
       count++;
-      GEOSSTRtree_insert_r(ctx, tree, geom, (void*)i);
+      GEOSSTRtree_insert_r(ctx, tree, geom, &kv_A(_tree_indexes, i));
```
brendan-ward, line 182: with array managed in tree

```suggestion
GEOSSTRtree_insert_r(ctx, tree, geom, &(_tree_indexes[i]);
```

src/strtree.c
```diff
@@ -115,6 +116,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
   npy_intp n, i, count = 0;
   GEOSGeometry* geom;
   pg_geom_obj_vec _geoms;
+  npy_intp_vec _tree_indexes;
```
brendan-ward, line 119:
```suggestion
  npy_intp* _tree_indexes;
```

src/strtree.c
```diff
@@ -145,7 +147,11 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
   n = PyArray_SIZE((PyArrayObject*)arr);
 
   kv_init(_geoms);
+  kv_init(_tree_indexes);
```
brendan-ward, line 150:
```suggestion
  _tree_indexes = (npy_intp*)malloc(n * sizeof(npy_intp));
```
(and drop following line)

src/strtree.c
```diff
@@ -145,7 +147,11 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
...
+  kv_resize(npy_intp, _tree_indexes, n);
+
   for (i = 0; i < n; i++) {
+    kv_push(npy_intp, _tree_indexes, i);
```
brendan-ward, line 154:
```suggestion
    _tree_indexes[i] = i;
```

src/strtree.c
```diff
@@ -159,6 +165,7 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
         Py_XDECREF(kv_A(_geoms, i));
       }
       kv_destroy(_geoms);
+      kv_destroy(_tree_indexes);
```
brendan-ward, line 168:
```suggestion
      free(_tree_indexes);
```

src/strtree.h
```diff
@@ -22,6 +22,7 @@ typedef struct {
   PyObject_HEAD void* ptr;
   npy_intp count;
   pg_geom_obj_vec _geoms;
+  npy_intp_vec _tree_indexes;
```
brendan-ward, line 25:
```suggestion
  npy_intp* _tree_indexes;
```

src/strtree.c
```diff
@@ -104,6 +104,7 @@ static void STRtree_dealloc(STRtreeObject* self) {
     Py_XDECREF(kv_A(self->_geoms, i));
   }
   kv_destroy(self->_geoms);
+  kv_destroy(self->_tree_indexes);
```
brendan-ward, line 107:
```suggestion
  free(self->_tree_indexes);
```

brendan-ward (commented):
Thanks for working on this.

What is the advantage to using resizable vectors here?  Per #129, I've been meaning to replace those with allocating an array of known size in advance.

Both the geometry array and the index array should be the same size as the incoming array; this could be managed in the tree via `malloc` and `free` instead of delegating that to the resizable vector.  That should also make passing the pointer to the values into GEOS a little cleaner.  I'll add a few suggestions inline that may help show this (untested); probably also needs an update on usage in query callback but I didn't touch that.

(updating the geometry vector should probably be in a different PR; will also need to add a `_size` member to track the size of that array in that case, since right now that is managed by the vector).

dbaston:
If the geometries you're putting in the tree are stored in fixed-size array, and you need to get the index when you query the tree (instead of getting the geometry directly), you can still store the geometry pointer in the tree and recover the index via pointer arithmetic from the head of the geometry array.

brendan-ward:
> store the geometry pointer in the tree and recover the index via pointer arithmetic

I think this might not work, since the incoming array we need to use for indexing may have geometries that we do not add to the tree (they may include Python `None` values for non-geometries, which we treat as `NULL` here and don't add to tree); thus the pointer offsets in the tree would be for only those geometries added to the tree.

But, if there is something I'm missing, this would be a clever way of avoiding a redundant array of indexes.

dbaston:
Not missing anything. This would only work if all the geometries in the contiguous array are added to the tree.

jorisvandenbossche:
> What is the advantage to using resizable vectors here?

Yeah, I did it mostly out of consistency (to not use a mixture of kvec and plain arrays). And, not having to store the size separately is nice (in general, for the tree_indexes we don't actually need it, that's true). But I know here it has a fixed size, that's also the reason I resized it up front. 
But I agree that using a plain array here would be nice (now, using C++ instead of C would probably even be nicer for such things .. ;)).

> I think this might not work, since the incoming array we need to use for indexing may have geometries that we do not add to the tree 

We indeed don't add each geometry to the *tree*, but looking more closely, we are actually adding something to `_geoms` for each element (just a NULL in case of emtpy/missing geometries). So pointer arithmetic with ``_geoms`` could actually work I think. 

I am only not sure how to do this in the `query_callback` function, as this does not have access to the full array. It receives the address of the `item` to be inserted, but to calculate its "index", I also need the address of the first element of the `_geom` array, which the `query_callback` function doesn't know?

dbaston:
>  I also need the address of the first element of the _geom array, which the query_callback function doesn't know?

You'd have to pass both the start of the `geom` array and the result vector to the callback by wrapping them in a struct. Something like this:

```
struct callback_userdata {
  GEOSGeometry* geom_array;
  npy_intp_vec* query_indexes;
};

void query_callback(void* item, void* user_data) {
  struct callback_userdata* ud = (struct callback_userdata*) user_data;
  size_t item_index = ((GEOSGeometry*) item) - ud->geom_array;

  kv_push(npy_intp, ud->query_indexes, item_index);
}
```

Not sure if it's worth the complexity.

brendan-ward:
New idea (testing now):
* store `_geoms` as an array of `GeometryObject`s that we `malloc` / `free` instead of resizable vector
* store `_size` on the `STRtreeObject` struct so that we know the size of that array at various places (some potential confusion vs `count` that is already there, which is the count of geometries added to the tree, whereas `_size` is the size of `_geoms`, which is the size of the input array )
* on insert to the tree, pass the address of the `GeometryObject`
* in `query_callback`, push the address of a `GeometryObject` onto a resizeable vector of `GeometryObject*`s
* use pointer offsets from the head of `_geoms` to calculate an index when we need it (either to return indexes or to get underlying GEOS geometry for predicate tests)

brendan-ward:
I'm getting some results from the pointer math that I don't completely understand (never used pointer math directly before).

Given `_geoms` is an array of `GeometryObject*` and that `GeometryObject*` has a size of 8 (on my machine), and given these addresses:
* head of _geoms: `0x7f8f52c3fb00`
* first geometry: `0x7f8f52c3fb00` 
* second geometry: `0x7f8f52c3fb08` (offset by 8 from head of _geoms)

When I calculate the offset of the second geometry like `geom - (GeometryObject*)_geoms` (otherwise compiler error comparing `GeometryObject* to GeometryObject**`), I get back 0 instead of 1.

When I calculate what the address would be by incrementing the pointer `(GeometryObject*)_geoms + 1` => `0x7f8f52c3fb20` (offset by 32 from head of _geoms).

Perhaps I'm missing a type cast somewhere?

dbaston:
I think `int index = geom - _geoms[0]` would work.

brendan-ward:
I think I figured it out, just needed to be more explicit about calculating the offsets ourselves and not relying on the compiler to infer the correct item size when using subtraction in this case:

```
char* head_ptr = (char*)_geoms;
char* geom_ptr = (char*)geom;  // e.g., 2nd geom in array
npy_intp geom_index = (npy_intp)((geom_ptr - head_ptr) / sizeof(GeometryObject*));
```

This now gives me the correct indices back out.

I think I might have fallen afoul of the C standard behavior around subtraction only being well defined when the elements are of the same array, and by the time I was running the subtraction (after first adding a geometry to a resizeable vector in `query_callback`) these were likely no longer considered part of the same array.

jorisvandenbossche:
Sorry for the slow follow up here, and thaks @brendan-ward for picking it up! I actually had a version locally with the suggestion of @dbaston, but only the minimal change, your PR is much more comprehensive change, so closing this PR as superseded by #265

PR #262, opened 2020-11-30
FIX: handle non-integer GEOS_VERSION_PATCH like 0beta1 as 0

mwtoews:
Non-releases of GEOS have `geos_c.h` headers like this:
```
#define GEOS_VERSION_PATCH 0beta1
```
which cannot be compiled, as this was expected to be an integer, not a string. This PR declares a quoted macro variable (using [this trick](https://stackoverflow.com/a/6671729/)), and takes the first substring to be the assumed patch digit. Theoretically, there could be more than one digit, but I don't think I've seen this from GEOS' history, so we could assume it will only be one character length.

Review carefully, as I seldom write C.

---

After this is merged, it should be possible to set `GEOS_VERSION: 3.9.0beta1` for CI testing along-side Python 3.9, similar to https://github.com/Toblerity/Shapely/pull/1041

### DIFF
--- a/src/lib.c
+++ b/src/lib.c
@@ -59,11 +59,18 @@ PyMODINIT_FUNC PyInit_lib(void) {
   import_array();
   import_umath();
 
+  /* add quotes to GEOS_VERSION_PATCH to handle (e.g.) 0beta1, then take the first digit */
+#define Q(x) #x
+#define QUOTE(x) Q(x)
+
+#define GEOS_VERSION_PATCH_STR QUOTE(GEOS_VERSION_PATCH)
+  PyObject* py_geos_version_patch_digit = PyUnicode_Substring(PyUnicode_FromString(GEOS_VERSION_PATCH_STR), 0, 1);
+
   /* export the GEOS versions as python tuple and string */
   PyModule_AddObject(m, "geos_version",
                      PyTuple_Pack(3, PyLong_FromLong((long)GEOS_VERSION_MAJOR),
                                   PyLong_FromLong((long)GEOS_VERSION_MINOR),
-                                  PyLong_FromLong((long)GEOS_VERSION_PATCH)));
+                                  PyLong_FromUnicodeObject(py_geos_version_patch_digit, 0)));
   PyModule_AddObject(m, "geos_capi_version",
                      PyTuple_Pack(3, PyLong_FromLong((long)GEOS_CAPI_VERSION_MAJOR),
                                   PyLong_FromLong((long)GEOS_CAPI_VERSION_MINOR),
--- a/src/lib.c
+++ b/src/lib.c
@@ -59,7 +59,9 @@ PyMODINIT_FUNC PyInit_lib(void) {
   import_array();
   import_umath();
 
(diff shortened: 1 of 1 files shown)


mwtoews:
If anyone is aware of a C preprocessor way to substring the first character, that would be more efficient than evaluating this expression each import.

caspervdw:
I think that it is OK to just chop off the beta-version info (as you are proposing). We always have the `geos_version_string` that shows the complete string.

I read a bit about this 'Stringification' (https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html), it turns out you really have to do this in this fashion. Thanks for figuring this out.

src/lib.c
```diff
@@ -59,11 +59,18 @@ PyMODINIT_FUNC PyInit_lib(void) {
   import_array();
   import_umath();
 
+  /* add quotes to GEOS_VERSION_PATCH to handle (e.g.) 0beta1, then take the first digit */
```
caspervdw, line 62: Can you maybe refer to the gcc docs (e.g. https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html)?

caspervdw (approved):

src/lib.c
mwtoews: Sure, although I opted to use more current https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html

It seems they have been shifting their term from "stringification" to "stringizing".

jorisvandenbossche:
It seems this was actually failing on the GEOS master CI build (which doesn't turn red if it fails .. something we should probably change once it is passing):

```
______________________________ test_geos_version _______________________________

    def test_geos_version():
        expected = "{0:d}.{1:d}.{2:d}".format(*pygeos.geos_version)
>       assert pygeos.geos_version_string == expected
E       AssertionError: assert '3.9.0beta2' == '3.9.0'
E         - 3.9.0
E         + 3.9.0beta2

pygeos/test/test_misc.py:23: AssertionError
____________________________ test_geos_capi_version ____________________________

    @pytest.mark.skipif(
        sys.platform.startswith("win") and pygeos.geos_version[:2] == (3, 7),
        reason="GEOS_C_API_VERSION broken for GEOS 3.7.x on Windows",
    )
    def test_geos_capi_version():
        expected = "{0:d}.{1:d}.{2:d}-CAPI-{3:d}.{4:d}.{5:d}".format(
            *(pygeos.geos_version + pygeos.geos_capi_version)
        )
>       assert pygeos.geos_capi_version_string == expected
E       AssertionError: assert '3.9.0beta2-CAPI-1.15.1' == '3.9.0-CAPI-1.15.1'
E         - 3.9.0-CAPI-1.15.1
E         + 3.9.0beta2-CAPI-1.15.1
E         ?      +++++

pygeos/test/test_misc.py:34: AssertionError
```

Those tests were passing before this PR on master (but that build is from 10 days ago, so that's not necessarily an indication the failures are caused by this PR. It's maybe just the GEOS beta release that caused it)

I suppose it's just the creation of the expected value that doesn't work with the beta tag on GEOS.

PR #263, opened 2020-12-01
ENH: Consolidate validation of enum values and raise ValueError on invalid values

brendan-ward:
This fixes a minor bug in the docstring for buffer; it listed an invalid `join_style` option.

While I was at it, I made the enums lowercase to match their usage in the functions for usability, and added a validation of values to give better error message than what you get back now when a value isn't in an `IntEnum`.

Note: this changes the original `KeyError` that would have been emitted to a `ValueError`, which seems more appropriate in this case (it is the _value_ that is wrong; that it is a key in an `IntEnum` is an implementation detail).

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -1,8 +1,8 @@
-from enum import IntEnum
 import numpy as np
 from . import Geometry  # NOQA
 from . import lib
 from .decorators import requires_geos, multithreading_enabled
+from .enum import ParamEnum
 
 
 __all__ = [
@@ -27,22 +27,16 @@ __all__ = [
 ]
 
 
-class BufferCapStyles(IntEnum):
+class BufferCapStyles(ParamEnum):
     round = 1
     flat = 2
     square = 3
 
 
-VALID_BUFFER_CAP_STYLES = {e.name for e in BufferCapStyles}
-
-
-class BufferJoinStyles(IntEnum):
+class BufferJoinStyles(ParamEnum):
     round = 1
     mitre = 2
     bevel = 3
-
-
-VALID_BUFFER_JOIN_STYLES = {e.name for e in BufferJoinStyles}
 
 
 @multithreading_enabled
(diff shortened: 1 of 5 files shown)


pygeos/constructive.py
```diff
@@ -146,9 +152,21 @@ def buffer(
...
     if isinstance(cap_style, str):
-        cap_style = BufferCapStyles[cap_style.upper()].value
+        if not cap_style in VALID_BUFFER_CAP_STYLES:
+            raise ValueError(
+                "cap_style '{}' is not valid, must be one of {}".format(
+                    cap_style, ", ".join(VALID_BUFFER_CAP_STYLES)
+                )
+            )
+        cap_style = BufferCapStyles[cap_style].value
     if isinstance(join_style, str):
```
jorisvandenbossche, line 162: Alternative could be to add a method like 

```
    @classmethod
    def get_value(cls, item):
        try:
            return cls[item].value
        except KeyError:
            raise ValueError("invalid ...")
```

to the Enum definitions (or a base class from which the two enums can inherit), and then here we can just do `cap_style = BufferCapStyles.get_value(cap_style)`. 
Which might give a little bit less code repetition

jorisvandenbossche (approved):
Looks good!

+1 on changing KeyError to ValueError

pygeos/constructive.py
brendan-ward: I like the idea of simplifying these via something like a new `ParamEnum` class that provides validation.  One possible disadvantage is that handling this validation in the API function (e.g., `buffer`), we're able to specifically name the parameter in the `ValueError` message, whereas that may not abstract as well to putting validation into the enum.

We're also using the same pattern for STRtree predicates.  However, there, the set of validation options is being used in GeoPandas as per [geopandas/pull/1698](https://github.com/geopandas/geopandas/pull/1698).

Perhaps one path forward would be to pull validation into the enum, and let downstream packages like geopandas use that enum to build their own set of valid options (e.g., via `{e.name for e in BinaryPredicate}` on their end) if they want to handle validation at that level?   (there was a preference for handling this at the downstream level too, for consistency of error messages)

This would have the advantage of exposing only the enums via our public API, but not a corresponding set of values.

/cc @adriangb

brendan-ward:
This consolidates validation of enums used in buffer and strtree.  I did not modify `GeometryType` enum since it does not look like that is used as a function parameter in the same way.

brendan-ward:
@caspervdw this is ready for final review and merge

caspervdw (approved):
Looks good to merge, thanks!

martinfleis:
Shouldn't the removal of `VALID_PREDICATES` be considered a hard breaking change? An environment with stable geopandas and master pygeos now fails. I know it will be fixed by geopandas/geopandas#1872 but it requires synchronization of releases between the two to avoid the same situation as we had with pygeos 0.9 which forced us to release geopandas 0.8.2.

jorisvandenbossche:
Ah, yes. Forgot about that, and easy to add back.

jorisvandenbossche:
-> https://github.com/pygeos/pygeos/pull/309

brendan-ward:
@martinfleis this is fixed on `master` here now, sorry for the issue.

PR #266, opened 2020-12-08
[API] Set None as default axis for reduction operations

### LINKED ISSUE #194: Default axis of union_all
Currently the `union_all` function has a default to union over axis 0:

https://github.com/pygeos/pygeos/blob/e83fc031d97ed232fcd3c117bfb4fc22df925da0/pygeos/set_operations.py#L162-L172

I was wondering if `axis=None` wouldn't be a better default (in which case it always gives a single geometry by default, regardless of the shape of the array), since that is also the default of numpy reductions (eg sum). Or did we have a specific reason to have the axis=0 deviation?

caspervdw:
Closes #194

This was not as easy as expected. I got the error that I could not reduce over multiple axes because the array elements were not reorderable for this operation. I traced that back to the numpy source:

https://github.com/numpy/numpy/blob/5cae51e794d69dd553104099305e9f92db237c53/numpy/core/src/umath/reduction.c#L207
https://github.com/numpy/numpy/blob/509b5ae4a7bb3c99324fd302ead1bbf4b130c741/numpy/core/src/umath/ufunc_object.c#L3637
https://github.com/numpy/numpy/blob/4d290795d4e8c60053eb789acf173159dc4c1575/numpy/core/src/umath/ufunc_object.c#L2445

It appears to be undocumented how to signal numpy that the operation is "reorderable" (commutative?). The trick was providing `PyUFunc_ReorderableNone` as `identity` in the ufunc definition. I can see this to be present as far back as numpy 1.14.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,7 +33,7 @@ Version 0.9 (unreleased)
   all input values are None (#249).
 * API change: the default axis of ``union_all``, ``intersection_all``, ``symmetric_difference_all``,
   and ``coverage_union_all`` can now reduce over multiple axes. The default changed from the first
-  axis (``0``) all all axes (``None``) (#266).
+  axis (``0``) to all axes (``None``) (#266).
 * Fixed internal GEOS error code detection for ``get_dimensions`` and ``get_srid`` (#218).
 * Addition of ``prepare`` function that generates a GEOS prepared geometry which is stored on
   the Geometry object itself. All binary predicates (except ``equals``) make use of this (#92).
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,9 @@ Version 0.9 (unreleased)
   instead of returning None if any value is None (#249).
 * API change: ``union_all`` now returns None (instead of ``GEOMETRYCOLLECTION EMPTY``) if
   all input values are None (#249).
+* API change: the default axis of ``union_all``, ``intersection_all``, ``symmetric_difference_all``,
+  and ``coverage_union_all`` can now reduce over multiple axes. The default changed from the first
+  axis (``0``) all all axes (``None``) (#266).
 * Fixed internal GEOS error code detection for ``get_dimensions`` and ``get_srid`` (#218).
 * Addition of ``prepare`` function that generates a GEOS prepared geometry which is stored on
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
I was going to say: but why does it then already work for `pygeos.union_all(arr, axis=None)` right now? But of course we have a custom implementation for union_all, not using a reducing ufunc ..

Thanks for figuring this out!

CHANGELOG.rst
```diff
@@ -31,6 +31,9 @@ Version 0.9 (unreleased)
   instead of returning None if any value is None (#249).
 * API change: ``union_all`` now returns None (instead of ``GEOMETRYCOLLECTION EMPTY``) if
   all input values are None (#249).
+* API change: the default axis of ``union_all``, ``intersection_all``, ``symmetric_difference_all``,
+  and ``coverage_union_all`` can now reduce over multiple axes. The default changed from the first
+  axis (``0``) all all axes (``None``) (#266).
```
jorisvandenbossche, line 36:
```suggestion
  axis (``0``) to all axes (``None``) (#266).
```

src/ufuncs.c
```diff
@@ -2425,10 +2430,13 @@ int init_ufuncs(PyObject* m, PyObject* d) {
...
-  DEFINE_YY_Y(intersection);
+  // 'REORDERABLE' sets PyUFunc_ReorderableNone instead of PyUFunc_None as 'identity',
+  // meaning that the order of arguments does not matter. This enables reduction over
+  // multiple (or all) axes.
+  DEFINE_YY_Y_REORDERABLE(intersection);
   DEFINE_YY_Y(difference);
-  DEFINE_YY_Y(symmetric_difference);
-  DEFINE_YY_Y(union);
+  DEFINE_YY_Y_REORDERABLE(symmetric_difference);
+  DEFINE_YY_Y_REORDERABLE(union);
```
jorisvandenbossche, line 2439: For union it's not needed strictly for `union_all` to make this a reorderable ufunc, but this enables `pygeos.lib.union.reduce(.., axis=None)` ?

caspervdw: It is indeed not implemented like the others. But still, I like to be consistent here, union is reorderable.

PR #267, opened 2020-12-09
ENH: add contains_properly predicate function

### LINKED ISSUE #180: ENH: Add support for ContainsProperly predicate
Note: this is exposed by GEOS only as an operation on prepared geometries.

jorisvandenbossche:
Closes #180

Since this is only available as a prepared function on the GEOS side, the question is what we do here:

- Automatically prepare the geometry within the `contains_properly` function (eg this can be done on the python side, by calling `pygeos.prepare` on the input before calling ``pygeos.lib.contains_properly``). 
- Still leave this as a responsibility to the user to provide prepared geometries (in which case I should probably add a custom error type on the C level so we can give a custom error message in python, because now I am just raising a GEOSException)

The second option is more consistent with the other predicates, but of course those all work out of the box for non-prepared geometries (so in terms of *usage*, the first option might look more consistent)

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -256,7 +256,19 @@ static PyUFuncGenericFunction YY_b_funcs[1] = {&YY_b_func};
 /* Define the geom, geom -> bool functions (YY_b) prepared */
 static void* contains_func_tuple[2] = {GEOSContains_r, GEOSPreparedContains_r};
 static void* contains_data[1] = {contains_func_tuple};
-static void* contains_properly_func_tuple[2] = {NULL, GEOSPreparedContainsProperly_r};
+static char GEOSContainsProperly(void* context, void* g1, void* g2) {
+  const GEOSPreparedGeometry* prepared_geom_tmp;
+  char ret;
+
+  prepared_geom_tmp = GEOSPrepare_r(context, g1);
+  if (prepared_geom_tmp == NULL) {
+      return 2;
+    }
+  ret = GEOSPreparedContainsProperly_r(context, prepared_geom_tmp, g2);
+  GEOSPreparedGeom_destroy_r(context, prepared_geom_tmp);
+  return ret;
+}
+static void* contains_properly_func_tuple[2] = {GEOSContainsProperly, GEOSPreparedContainsProperly_r};
 static void* contains_properly_data[1] = {contains_properly_func_tuple};
 static void* covered_by_func_tuple[2] = {GEOSCoveredBy_r, GEOSPreparedCoveredBy_r};
 static void* covered_by_data[1] = {covered_by_func_tuple};
@@ -300,10 +312,6 @@ static void YY_b_p_func(char** args, npy_intp* dimensions, npy_intp* steps, void
       ret = 0;
     } else {
       if (in1_prepared == NULL) {
-        if (func == NULL) {
-          errstate = PGERR_GEOS_EXCEPTION;
-          goto finish;
-        }
         /* call the GEOS function */
         ret = func(ctx, in1, in2);
       } else {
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
I included a few changes from https://github.com/pygeos/pygeos/pull/252, but once that PR is merged, will update here

brendan-ward:
@jorisvandenbossche  thanks for working on this!

General comment: I think that this requires prepared geometries is an implementation detail in GEOS, not a user-facing concern.

I think I would prefer a variant of your option 1, which behaves more like STRtree: use prepared geometry if available, create an internal prepared geometry and destroy it after if not available (all in C, not Python).  This has the benefit of creating no side effects on the geometry objects.  

Then in the docstring, simply note that this is being done.

This seems like it would create a more uniform predicates API; all predicates can accept prepared or unprepared geometries; if prepared are available it will use them for possible speedups.

jorisvandenbossche:
Thanks for the feedback! 

I understand that doing this on the C level would give a more uniform API without potentially unexpected side effects. However, I am personally not sure it is worth the added complexity ..
IMO, if we want this to be an implementation detail (which I agree with), I would just ensure that the geometries are prepared, without keeping track of which ones where already prepared beforehand and which not. And a note in the docstring that this function will prepare your geometries should IMO be sufficient for those who care about the increased memory usage (and they can then destroy the prepared-aspect manually if the want). 

(BTW, also for STRtree, we could IMO move to store the prepared geometries in the GeometryObject, but that's for a another discussion (although quite related))

Now, if others prefer to have a version without side effects, no problem and I am happy to implement it (the implementation itself shouldn't be difficult, it's just yet another `PyUFuncGenericFunction` version, and every time we add one, the complexity of ufuncs.c increases a little bit)

jorisvandenbossche:
(IMO we should also still discuss what *long term* user facing API we want related to preparing geometries. For example, IMO there is something to say for *always* preparing in the predicates by *default*, with the option to turn it off if you don't want it, but so that as a user you don't need to think about doing it manually to benefit from the speed-ups. Again, related but different discussion ;))

jorisvandenbossche:
> I think I would prefer a variant of your option 1, which behaves more like STRtree: use prepared geometry if available, create an internal prepared geometry and destroy it after if not available (all in C, not Python). This has the benefit of creating no side effects on the geometry objects.

@brendan-ward Looking at this again, this is actually quite straightforward (and not that much additional complexity, as I was arguing above), since we already have a switch between preprared and non-prepared functions in the predicate ufunc inner loop implementation. So I just added a wrapper around `GEOSPreparedContainsProperly_r` that prepares+destroys to use for the case of non-prepared input geometries. This doesn't add much complexity, and you get the exact same behaviour as for the other predicates

src/ufuncs.c
```diff
@@ -256,7 +256,19 @@ static PyUFuncGenericFunction YY_b_funcs[1] = {&YY_b_func};
 /* Define the geom, geom -> bool functions (YY_b) prepared */
 static void* contains_func_tuple[2] = {GEOSContains_r, GEOSPreparedContains_r};
 static void* contains_data[1] = {contains_func_tuple};
-static void* contains_properly_func_tuple[2] = {NULL, GEOSPreparedContainsProperly_r};
+static char GEOSContainsProperly(void* context, void* g1, void* g2) {
+  const GEOSPreparedGeometry* prepared_geom_tmp;
```
brendan-ward, line 260:
```suggestion
  const GEOSPreparedGeometry* prepared_geom_tmp = NULL;
```

brendan-ward (approved):
Thanks for the changes, I'm glad it was an easy change.

I just tested this with some real-world data where I was really wanting to use `contains_properly` anyway (due to boundary issues) (20k pairs of multipolygons):

```
contains: 4.59s
contains_properly: 0.77s
```

Quite the speedup!

caspervdw (approved):
Thanks @jorisvandenbossche !

Issue #268, opened 2020-12-09
Failures with numpy 1.20.0rc1


jorisvandenbossche:
See eg https://github.com/pygeos/pygeos/runs/1519984556

```
 ____________________________ test_from_shapely_arr _____________________________

    @mock.patch("pygeos.io.ShapelyGeometry", ShapelyGeometryMock)
    @mock.patch("pygeos.io.shapely_geos_version", pygeos.geos_capi_version_string)
    def test_from_shapely_arr():
>       actual = pygeos.from_shapely([ShapelyGeometryMock(point), None])

pygeos/test/test_io.py:341: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pygeos/io.py:243: in from_shapely
    arr = np.asarray(geometry, dtype=object)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = [<pygeos.test.test_io.ShapelyGeometryMock object at 0x7f65862e0190>, None]
dtype = <class 'object'>, order = None

    @set_array_function_like_doc
    @set_module('numpy')
    def asarray(a, dtype=None, order=None, *, like=None):
        """
        ....
        """
        if like is not None:
            return _asarray_with_like(a, dtype=dtype, order=order, like=like)
    
>       return array(a, dtype, copy=False, order=order)
E       ValueError: Invalid __array_interface__ value, must be a dict
```

This `ShapelyGeometryMock` is defined as 

https://github.com/pygeos/pygeos/blob/211fc48d3314bdd402f8071293659a5368e1cc7f/pygeos/test/test_io.py#L23-L30

but so apparently numpy now calls the `__array_interface__` on individual objects in a list in `np.asarray(..)`.

PR #269, opened 2020-12-09
Update for compatibility with numpy 1.20 (builtin type aliases, array coercion)

jorisvandenbossche:
Closes https://github.com/pygeos/pygeos/issues/268

Mostly fixing warnings. In addition (the example shown in #268), the `from_shapely` function was broken in case the input was a list-like (not already an array, or not a scalar geometry)

### DIFF
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -241,13 +241,11 @@ def from_shapely(geometry, **kwargs):
         arr = np.empty(1, dtype=object)
         arr[0] = geometry
         arr.shape = ()
-    elif geometry is None:
-        arr = geometry
     elif not isinstance(geometry, np.ndarray) and isinstance(geometry, Collection):
         # geometry is a list/array-like
         arr = np.empty(len(geometry), dtype=object)
         arr[:] = geometry
     else:
-        # we already have a numpy array
+        # we already have a numpy array or we are None
         arr = geometry
     return lib.from_shapely(arr, **kwargs)
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -93,9 +93,9 @@ def to_wkt(
     return lib.to_wkt(
         geometry,
         np.intc(rounding_precision),
-        np.bool(trim),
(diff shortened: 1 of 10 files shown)


pygeos/io.py
```diff
@@ -239,6 +239,13 @@ def from_shapely(geometry, **kwargs):
         arr = np.empty(1, dtype=object)
         arr[0] = geometry
         arr.shape = ()
+    elif geometry is None:
+        arr = geometry
+    elif not isinstance(geometry, np.ndarray):
+        # geometry is a list/array-like
+        arr = np.empty(len(geometry), dtype=object)
+        arr[:] = geometry
```
jorisvandenbossche, line 247: @caspervdw this fails for the `ShapelyGeometryMock` that we use in our tests (because this is not a subclass of shapely's `BaseGeometry`) .. Would you be fine to have those tests actually use Shapely instead? 
I am not fully sure how to make it work otherwise (unless I would make the above `isinstance(geometry, ShapelyGeometry)` or duck-type check like `hasattr(geometry, "__geom__")`, but then that change would *only* be to please the mock in the tests ..

jorisvandenbossche: Actually I needed a fix for this for the shapely dev branch as well, since I am using a DummyGeometry class there to pass pointers to GEOS, which is also not a BaseGeometry subclass. So added an additional check for the object having a length (being a Collection abc)

caspervdw: I think that it is sufficient to skip the unittests if shapely is not present.

jorisvandenbossche: I think it was actually fixed now by adding the additional check for being a collection here (so the ShapelyGeometryMock then passes through to the next `else:` block)

jorisvandenbossche:
Whoops, merged without checking the test log again (because for the build with GEOS/numpy master, we don't let it be "red" if there are failures) -> https://github.com/pygeos/pygeos/pull/278

PR #270, opened 2020-12-11
ENH: Add STRtree benchmarks

brendan-ward:
Adds basic benchmarks of STRtree query bulk functionality against polygons.

On my local machine, this shows:
```
time_tree_create                                                                                 4.47±0.4μs
time_tree_query_bulk                                                                             3.39±0.5μs
time_tree_query_bulk_contains                                                                      67.6±4ms
time_tree_query_bulk_contains_properly                                                          1.04±0.06ms
time_tree_query_bulk_covered_by                                                                    51.2±6ms
time_tree_query_bulk_covers                                                                        65.0±3ms
time_tree_query_bulk_crosses                                                                       59.8±4ms
time_tree_query_bulk_intersects                                                                    5.70±1ms
time_tree_query_bulk_overlaps                                                                      48.3±4ms
time_tree_query_bulk_touches                                                                       55.2±7ms
time_tree_query_bulk_within                                                                        49.8±3ms
```

Among other things, this highlights which predicates have a prepared predicate implementation in GEOS (`intersects`, `contains_properly`).

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -81,9 +81,7 @@ class STRtree:
     """Benchmarks queries against STRtree"""
 
     def setup(self):
-        # create irregular polygons using union of buffered random points
-        points = pygeos.points(np.random.random((1000, 2)))
-        self.polygons = pygeos.get_parts(pygeos.union_all(pygeos.buffer(points, 0.025)))
+        self.polygons = pygeos.polygons(np.random.random((100, 3, 2)))
         self.tree = pygeos.STRtree(self.polygons)
 
     def time_tree_create(self):
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -56,9 +56,9 @@ class ConstructiveSuite:
         pygeos.delaunay_triangles(self.points)
 
 
-
 class GetParts:
     """Benchmarks for getting individual parts from 100 multipolygons of 100 polygons each"""
+
     def setup(self):
         self.multipolygons = np.array([pygeos.multipolygons(pygeos.polygons(np.random.random((2, 100, 2)))) for i in range(10000)], dtype=object)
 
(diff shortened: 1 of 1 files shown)


brendan-ward:
Previous test geometries didn't really work as well as I hoped; switched to random triangles and now see more divergence in benchmarks:

```
time_tree_create                                                                                 25.2±0.4μs
time_tree_query_bulk                                                                                322±9μs
time_tree_query_bulk_contains                                                                    6.42±0.4ms
time_tree_query_bulk_contains_properly                                                          1.42±0.07ms
time_tree_query_bulk_covered_by                                                                  10.9±0.7ms
time_tree_query_bulk_covers                                                                      6.18±0.2ms
time_tree_query_bulk_crosses                                                                        150±6ms
time_tree_query_bulk_intersects                                                                  23.2±0.4ms
time_tree_query_bulk_overlaps                                                                       154±4ms
time_tree_query_bulk_touches                                                                        153±3ms
time_tree_query_bulk_within                                                                      12.0±0.2ms
```

benchmarks/benchmarks.py
```diff
@@ -75,3 +75,44 @@ def time_get_parts_python(self):
             parts.append(pygeos.get_geometry(self.multipolygons[i], range(num_parts)))
 
         parts = np.concatenate(parts)
+
+
+class STRtree:
+    """Benchmarks queries against STRtree"""
+
+    def setup(self):
+        self.polygons = pygeos.polygons(np.random.random((100, 3, 2)))
```
jorisvandenbossche, line 84: Although this is an easy test creation (I also use it myself often), one concern is that this might not be typical data *for spatial indexing*, because many of those polygons intersect: from a quick test, often more than 60% of all polygons combinations in the array are overlapping (for other benchmarks, like element-wise operations where dependence between elements doesn't matter, this should not be a concern, but for a spatial index it might matter)

Now, it's probably fine to add those benchmarks anyway, but it might be good to also think about other potential (random or not) datasets.

benchmarks/benchmarks.py
```diff
@@ -75,3 +75,44 @@ def time_get_parts_python(self):
...
+        self.tree = pygeos.STRtree(self.polygons)
+
+    def time_tree_create(self):
+        pygeos.STRtree(self.polygons)
```
jorisvandenbossche, line 88: It's fine to benchmark this, but we just need to be aware it's not the "tree" creation, but only the object construction (the tree is only created on first query)

benchmarks/benchmarks.py
```diff
@@ -75,3 +75,44 @@ def time_get_parts_python(self):
...
         parts = np.concatenate(parts)
+
+
+class STRtree:
+    """Benchmarks queries against STRtree"""
+
+    def setup(self):
+        self.polygons = pygeos.polygons(np.random.random((100, 3, 2)))
+        self.tree = pygeos.STRtree(self.polygons)
+
```
jorisvandenbossche, line 86: Maybe can do here a tiny query with a single point or so, to already trigger the tree creation

jorisvandenbossche (commented):
Cool, thanks for adding those! 
Maybe we can increase the size of the array a bit more (eg from 100 polygons to 1000 or 10000 polygons?)

benchmarks/benchmarks.py
jorisvandenbossche: I was thinking, if we could add a random offset (an affine translate) to spread them out over a bit larger area, that would reduce the overlap quite a bit. Something like:

```
a, b = -2, 2
offset = (b - a) * np.random.random_sample(100) + a
pygeos.polygons(np.random.random((100, 3, 2)) + offset[:, None, None])
``

File: benchmarks/benchmarks.py (hunk 2 above)
Comment (brendan-ward) [reply benchmarks/benchmarks.py:88]: Right.  My goal was more to ensure that we didn't have a performance regression in the pygeos code to construct the tree from an array.  Perhaps it would be better to indicate this as `time_tree_insert`

**Comment by brendan-ward** [comment]
Yeah, should have visually inspected the test geometries more.  Reworked that a bit to give us a mix of irregular and isolated polygons, now it looks more like this:
image

It is a bit slow to create these, so I limited to 2,000 polygons.

Latest results:
```
time_tree_create                                                                                   287±10μs
time_tree_query_bulk                                                                                555±7μs
time_tree_query_bulk_contains                                                                      298±40ms
time_tree_query_bulk_contains_properly                                                          6.80±0.08ms
time_tree_query_bulk_covered_by                                                                     188±2ms
time_tree_query_bulk_covers                                                                         290±2ms
time_tree_query_bulk_crosses                                                                        237±3ms
time_tree_query_bulk_intersects                                                                  23.4±0.8ms
time_tree_query_bulk_overlaps                                                                       235±2ms
time_tree_query_bulk_touches                                                                        237±2ms
time_tree_query_bulk_within                                                                         190±2ms
```

jorisvandenbossche:
Cool, that looks good. 
Maybe we should use a fixed random seed in this case? As I can imagine that the exact number of geometries that get created (which will vary) can influence the timings, and we want to use this for comparing multiple runs of it.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #273, opened 2020-12-16
ENH: clip_by_rect

jorisvandenbossche:
(we could certainly make xmin/ymin/xmax/ymax accept arrays as well, but didn't include that for a first pass)

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -12,6 +12,7 @@ __all__ = [
     "buffer",
     "offset_curve",
     "centroid",
+    "clip_by_rect",
     "convex_hull",
     "delaunay_triangles",
     "envelope",
@@ -253,6 +254,46 @@ def centroid(geometry, **kwargs):
     <pygeos.Geometry POINT EMPTY>
     """
     return lib.centroid(geometry, **kwargs)
+
+
+@multithreading_enabled
+def clip_by_rect(geometry, xmin, ymin, xmax, ymax, **kwargs):
+    """
+    Returns the portion of a geometry within a rectangle.
+
+    The geometry is clipped in a fast but possibly dirty way. The output is
+    not guaranteed to be valid. No exceptions will be raised for topological
+    errors.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+        The geometry to be clipped
+    xmin : float
+        Minimum x value of the rectangle
+    ymin : float
+        Minimum y value of the rectangle
+    xmax : float
+        Maximum x value of the rectangle
+    ymax : float
+        Maximum y value of the rectangle
+
+    Examples
+    --------
+    >>> line = Geometry("LINESTRING (0 0, 10 10)")
+    >>> clip_by_rect(line, 0., 0., 1., 1.)
+    <pygeos.Geometry LINESTRING (0 0, 1 1)>
+    """
+    if not all(np.isscalar(val) for val in [xmin, ymin, xmax, ymax]):
+        raise TypeError("xmin/ymin/xmax/ymax only accepts scalar values")
+    return lib.clip_by_rect(
+        geometry,
+        np.double(xmin),
+        np.double(ymin),
+        np.double(xmax),
+        np.double(ymax),
+        **kwargs
+    )
 
 
 @multithreading_enabled
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -265,6 +265,9 @@ def clip_by_rect(geometry, xmin, ymin, xmax, ymax, **kwargs):
     not guaranteed to be valid. No exceptions will be raised for topological
     errors.
 
+    Note: empty geometries or geometries that do not overlap with the
+    specified bounds will result in GEOMETRYCOLLECTION EMPTY.
+
     Parameters
     ----------
     geometry : Geometry or array_like
--- a/pygeos/test/test_constructive.py
+++ b/pygeos/test/test_constructive.py
@@ -308,8 +308,68 @@ def test_reverse_invalid_type(geom):
         pygeos.reverse(geom)
 
 
+@pytest.mark.parametrize(
+    "geom,expected",
+    [
+        # Point outside
+        ("POINT (0 0)", "GEOMETRYCOLLECTION EMPTY"),
+        # Point inside
+        ("POINT (15 15)", "POINT (15 15)"),
+        # Point on boundary
+        ("POINT (15 10)", "GEOMETRYCOLLECTION EMPTY"),
+        # Line outside
+        ("LINESTRING (0 0, -5 5)", "GEOMETRYCOLLECTION EMPTY"),
+        # Line inside
+        ("LINESTRING (15 15, 16 15)", "LINESTRING (15 15, 16 15)"),
+        # Line on boundary
+        ("LINESTRING (10 15, 10 10, 15 10)", "GEOMETRYCOLLECTION EMPTY"),
+        # Line splitting rectangle
+        ("LINESTRING (10 5, 25 20)", "LINESTRING (15 10, 20 15)"),
+    ],
+)
+def test_clip_by_rect(geom, expected):
+    geom, expected = pygeos.Geometry(geom), pygeos.Geometry(expected)
+    actual = pygeos.clip_by_rect(geom, 10, 10, 20, 20)
+    assert pygeos.equals(actual, expected)
+
+
+@pytest.mark.parametrize(
+    "geom, rect, expected",
+    [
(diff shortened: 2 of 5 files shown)


pygeos/test/test_constructive.py
```diff
@@ -306,3 +306,28 @@ def test_reverse_none():
...
+
+
+def test_clip_by_rect_missing():
+    actual = pygeos.clip_by_rect(None, 0, 0, 1, 1)
+    assert actual is None
+
+
+def test_clip_by_rect_empty():
+    actual = pygeos.clip_by_rect(empty_polygon, 0, 0, 1, 1)
+    assert actual == Geometry("GEOMETRYCOLLECTION EMPTY")
```
brendan-ward, line 325: This seems odd that an empty polygon would be converted to an empty geometry collection

brendan-ward: This fails outright for empty points:
```python
>>> from pygeos.test.common import empty_point
>>> pygeos.clip_by_rect(empty_point, 0,0,10,10)
...
pygeos.GEOSException: UnsupportedOperationException: getX called on empty Point
```

Whereas intersection does not:
```python
>>> pygeos.intersection(empty_point, pygeos.box(0,0,10,10))
<pygeos.Geometry GEOMETRYCOLLECTION EMPTY>
```

src/ufuncs.c
```diff
@@ -1339,6 +1339,67 @@ static void snap_func(char** args, npy_intp* dimensions, npy_intp* steps, void*
...
+    if (!get_geom(*(GeometryObject**)ip1, &in1)) {
+      errstate = PGERR_NOT_A_GEOMETRY;
+      destroy_geom_arr(ctx, geom_arr, i - 1);
+      break;
+    }
+
+    if (in1 == NULL) {
+      // in case of a missing value: return NULL (None)
+      geom_arr[i] = NULL;
+    } else {
```
brendan-ward, line 1383: It might be good to guard for empty geometries here.  It looks like right now they tend to be coalesced to empty geometry collections by this and by `intersection`, but I wonder if we should just return a clone of the input so that we aren't changing geometry type or dimensions by this operation.

pygeos/test/test_constructive.py
```diff
@@ -306,3 +306,28 @@ def test_reverse_none():
 def test_reverse_invalid_type(geom):
     with pytest.raises(TypeError, match="One of the arguments is of incorrect type"):
         pygeos.reverse(geom)
+
+
+@pytest.mark.parametrize("geometry", all_types)
+def test_clip_by_rect(geometry):
+    actual = pygeos.clip_by_rect([geometry, geometry], 0.0, 0.0, 1.0, 1.0)
+    assert actual.shape == (2,)
+    assert actual[0] is None or isinstance(actual[0], Geometry)
```
brendan-ward, line 315: This test doesn't validate expected results from the clipping operation.  I'd recommend using specific pairs of inputs and clipped outputs.

Do watch for this case though: https://trac.osgeo.org/geos/ticket/1056
I was able to reproduce the issue on 3.8

pygeos/test/test_constructive.py
```diff
@@ -306,3 +306,28 @@ def test_reverse_none():
...
+    actual = pygeos.clip_by_rect(None, 0, 0, 1, 1)
+    assert actual is None
+
+
+def test_clip_by_rect_empty():
+    actual = pygeos.clip_by_rect(empty_polygon, 0, 0, 1, 1)
+    assert actual == Geometry("GEOMETRYCOLLECTION EMPTY")
+
+
+def test_clip_by_rect_missing_non_scalar_kwargs():
```
brendan-ward, line 328:
```suggestion
def test_clip_by_rect_non_scalar_kwargs():
```

brendan-ward (changes requested):
Thanks for working on this!

A few suggested changes around testing and empty geometries.

pygeos/test/test_constructive.py
jorisvandenbossche: The empty points seems a bit strange (and something we can certainly catch), but for the rest this is the behaviour of GEOS. I am not sure to what extent we should start changing this (or otherwise bring it up with them if we think this is strnage behaviour). As you show, `intersection` also returns a geometry collection for empty input:

```
In [12]: pygeos.intersection(pygeos.Geometry("POLYGON EMPTY"), pygeos.box(0,0,10,10))
Out[12]: <shapely.geometry.GeometryCollection GEOMETRYCOLLECTION EMPTY>
```

So that's probably a reason to leave it here as is as well?

pygeos/test/test_constructive.py
jorisvandenbossche: > This test doesn't validate expected results from the clipping operation. I'd recommend using specific pairs of inputs and clipped outputs.

I know, but that's what we do for most other functions in this test file as well (I suppose in the rationale: we test that it works, but rely for correctness on GEOS). Now, I can certainly add a test with a few cases that ensures it's doing a clipping operation and not just something random (suppose that I called the wrong GEOS function in ufuncs.c), but it's always going to be limited.

src/ufuncs.c
jorisvandenbossche: See also the answer above. I certainly understand the concern, and it was also my first reaction (why can't the empty geometry not preserve its geometry type?). But we also don't do such check for other overlay operations like intersection etc, so this might warrant a more general discussion/change?

pygeos/test/test_constructive.py
jorisvandenbossche: I added some test cases from shapely!

pygeos/test/test_constructive.py
brendan-ward: Yeah, we should probably leave this at the default response from GEOS.

pygeos/constructive.py
```diff
@@ -255,6 +256,46 @@ def centroid(geometry, **kwargs):
...
 
+@multithreading_enabled
+def clip_by_rect(geometry, xmin, ymin, xmax, ymax, **kwargs):
+    """
+    Returns the portion of a geometry within a rectangle.
+
+    The geometry is clipped in a fast but possibly dirty way. The output is
+    not guaranteed to be valid. No exceptions will be raised for topological
+    errors.
+
```
brendan-ward, line 267: Based on the results for empty and your test cases here, it might be a good idea to note that empty geometries or geometries that do not overlap with the specified bounds will result in `GEOMETRYCOLLECTION EMPTY`.

brendan-ward (approved):
Thanks for adding extra test cases!

Other than possibly adding a note about `GEOMETRYCOLLECTION EMPTY` results, this looks good to me.

brendan-ward:
Here is a benchmark to consider adding (could certainly be improved, too):
```python

class BoxCutter:
    """Benchmarks for different methods of cutting geometries by boxes"""

    def setup(self):
        # create irregular polygons by merging overlapping point buffers
        self.polygon = pygeos.union_all(
                pygeos.buffer(pygeos.points(np.random.random((1000, 2)) * 500), 10)
            )
        xmin = np.random.random(100) * 100
        xmax = xmin + 100
        ymin = np.random.random(100) * 100
        ymax = ymin + 100
        self.bounds = np.array([xmin, ymin, xmax, ymax]).T
        self.boxes = pygeos.box(xmin, ymin, xmax, ymax)


    def time_clip_by_box(self):
        pygeos.intersection(self.polygon, self.boxes)

    def time_clip_by_box_loop(self):
        for box in self.boxes:
            # use a for loop for direct comparison to clip by rect
            pygeos.intersection(self.polygon, box)

    def time_clip_by_rect(self):
        for bounds in self.bounds:
            pygeos.clip_by_rect(self.polygon, *bounds)

```

`clip_by_rect` is substantially faster even though we do a bit of looping:

```
time_clip_by_box                                                                    153±5ms
time_clip_by_box_loop                                                               157±8ms
time_clip_by_rect                                                                19.5±0.6ms
```

(benchmark is against GEOS master branch, which uses overlayNG for intersection - may impact results...)

jorisvandenbossche:
Thanks, will add that. I only don't fully understand what the goal of `time_clip_by_box_loop` is ?

brendan-ward:
The goal of `time_clip_by_box_loop` was to make it more directly comparable to `time_clip_by_rect` since both use `for` loops, though for this particular benchmark it doesn't incur enough overhead from doing the loop in Python vs in the ufunc that it diverges much from the pure ufunc function.

Feel free to drop it.

jorisvandenbossche:
Updated to include the note about empties, add a changelog entry, and add the proposed benchmark.

PR #275, opened 2020-12-30
ENH: add polygonize ufunc

jorisvandenbossche:
There is also `polygonize_full` to add

I am not sure the ufunc version will be that useful (since with a 2D array, you need the same number of lines for each entry). But most consistent to expose it this way of course.

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -490,7 +490,7 @@ def point_on_surface(geometry, **kwargs):
 def polygonize(geometries, **kwargs):
     """Creates polygons formed from the linework of a set of Geometries.
 
-    Polygonizes a set of Geometries which contain linework that
+    Polygonizes an array of Geometries that contain linework which
     represents the edges of a planar graph. All types of Geometry are
     accepted as input; the constituent linework is extracted as the edges
     to be polygonized.
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -491,9 +491,13 @@ def polygonize(geometries, **kwargs):
     """Creates polygons formed from the linework of a set of Geometries.
 
     Polygonizes an array of Geometries that contain linework which
-    represents the edges of a planar graph. All types of Geometry are
-    accepted as input; the constituent linework is extracted as the edges
-    to be polygonized.
+    represents the edges of a planar graph. Any type of Geometry may be
+    provided as input; only the constituent lines and rings will be used to
+    create the output polygons.
+
+    Lines or rings that when combined do not completely close a polygon
+    will result in an empty GeometryCollection.  Duplicate segments are
+    ignored.
 
     This function returns the polygons within a GeometryCollection.
     Individual Polygons can be obtained using `get_geometry` to get
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -459,9 +459,9 @@ def polygonize(geometries, **kwargs):
     Examples
     --------
     >>> lines = [
-    ...     pygeos.Geometry("LINESTRING (0 0, 1 1)"),
-    ...     pygeos.Geometry("LINESTRING (0 0, 0 1)"),
-    ...     pygeos.Geometry("LINESTRING (0 1, 1 1)"),
+    ...     Geometry("LINESTRING (0 0, 1 1)"),
+    ...     Geometry("LINESTRING (0 0, 0 1)"),
+    ...     Geometry("LINESTRING (0 1, 1 1)"),
     ... ]
     >>> polygonize(lines)
     <pygeos.Geometry GEOMETRYCOLLECTION (POLYGON ((1 1, 0 0, 0 1, 1 1)))>
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -495,6 +495,12 @@ def polygonize(geometries, **kwargs):
     accepted as input; the constituent linework is extracted as the edges
     to be polygonized.
 
+    This function returns the polygons within a GeometryCollection.
+    Individual Polygons can be obtained using `get_geometry` to get
+    a single polygon or `get_parts` to get an array of polygons.
+    MultiPolygons can be constructed from the output using
+    ``pygeos.multipolygons(pygeos.get_parts(pygeos.polygonize(geometries)))``.
+
     Parameters
     ----------
     geometries : array_like
@@ -503,6 +509,14 @@ def polygonize(geometries, **kwargs):
         Axis along which the geometries are polygonized.
         The default is to perform a reduction over the last dimension
         of the input array. A 1D array results in a scalar geometry.
+
+    Returns
+    -------
+    GeometryCollection or array of GeometryCollections
+
+    See Also
+    --------
+    get_parts, get_geometry
 
     Examples
     --------
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -20,6 +20,7 @@ __all__ = [
     "make_valid",
     "normalize",
     "point_on_surface",
+    "polygonize",
     "reverse",
     "simplify",
     "snap",
@@ -440,6 +441,32 @@ def point_on_surface(geometry, **kwargs):
     <pygeos.Geometry POINT EMPTY>
     """
     return lib.point_on_surface(geometry, **kwargs)
+
+
+def polygonize(geometries, **kwargs):
+    """Creates polygons formed from the linework of a set of Geometries.
+
+    Polygonizes a set of Geometries which contain linework that
+    represents the edges of a planar graph. All types of Geometry are
+    accepted as input; the constituent linework is extracted as the edges
+    to be polygonized.
+
+    Parameters
+    ----------
+    geometries : array_like
+        An array of geometries.
+
+    Examples
+    --------
+    >>> lines = [
+    ...     pygeos.Geometry("LINESTRING (0 0, 1 1)"),
+    ...     pygeos.Geometry("LINESTRING (0 0, 0 1)"),
+    ...     pygeos.Geometry("LINESTRING (0 1, 1 1)"),
+    ... ]
+    >>> polygonize(lines)
+    <pygeos.Geometry GEOMETRYCOLLECTION (POLYGON ((1 1, 0 0, 0 1, 1 1)))>
+    """
+    return lib.polygonize(geometries, **kwargs)
 
 
 @requires_geos("3.7.0")
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -499,6 +499,10 @@ def polygonize(geometries, **kwargs):
     ----------
     geometries : array_like
         An array of geometries.
+    axis : int
+        Axis along which the geometries are polygonized.
+        The default is to perform a reduction over the last dimension
+        of the input array. A 1D array results in a scalar geometry.
 
     Examples
     --------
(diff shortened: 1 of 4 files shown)


pygeos/constructive.py
```diff
@@ -486,6 +487,32 @@ def point_on_surface(geometry, **kwargs):
...
+    """Creates polygons formed from the linework of a set of Geometries.
+
+    Polygonizes a set of Geometries which contain linework that
+    represents the edges of a planar graph. All types of Geometry are
+    accepted as input; the constituent linework is extracted as the edges
+    to be polygonized.
+
+    Parameters
+    ----------
+    geometries : array_like
```
caspervdw, line 500: I'd expect this function to have an `axis` kwarg, and that to have `None` as a sensible default (see #194) - but this function is not a `ufunc.reduce` but a generalized ufunc.

src/ufuncs.c
```diff
@@ -1682,6 +1682,50 @@ static void relate_pattern_func(char** args, npy_intp* dimensions, npy_intp* ste
...
+    if (ret_ptr == NULL) {
+      errstate = PGERR_GEOS_EXCEPTION;
+      goto finish;
+    }
+    OUTPUT_Y;
+  }
+
+finish:
+  if (geoms != NULL) {
+    free(geoms);
```
caspervdw, line 1723: In case of an error, the already created geometries need to be destroyed. You now deallocate the array holding the geometry pointers, but the geometries itself will form a memory leak if the array was partially filled already.

I think such a thing is already done in the Y_Y functions.

pygeos/constructive.py
```diff
@@ -486,6 +487,32 @@ def point_on_surface(geometry, **kwargs):
     return lib.point_on_surface(geometry, **kwargs)
 
 
+def polygonize(geometries, **kwargs):
+    """Creates polygons formed from the linework of a set of Geometries.
+
+    Polygonizes a set of Geometries which contain linework that
```
caspervdw, line 493:
```suggestion
    Polygonizes an array of Geometries that contain linework which
```

A small that-vs-which bikeshedding here. Also: `set` is confusing in the context of python.

caspervdw (changes requested):
Thanks @jorisvandenbossche ! I left some remarks

pygeos/constructive.py
jorisvandenbossche: Yeah, so our other generalized ufuncs like `points` also don't have an axis keyword? 

-> That's what I wanted to answer, but actually the actual ufunc seems to support this (which I wasn't aware of, or didn't think about before):

```
In [32]: pygeos.lib.points(np.array([[1, 2], [3, 4]]), axis=1)
Out[32]: 
array([<pygeos.Geometry POINT (1 2)>, <pygeos.Geometry POINT (3 4)>],
      dtype=object)

In [33]: pygeos.lib.points(np.array([[1, 2], [3, 4]]), axis=0)
Out[33]: 
array([<pygeos.Geometry POINT (1 3)>, <pygeos.Geometry POINT (2 4)>],
      dtype=object)
```

So by default the last axis is reduced, but you can actually specify a different axis to collapse (assuming it has a correct number of elements in case of `points`)

But, those generalized ufuncs don't support the `axis=None` option.

---

Now, I could of course still add an `axis` keyword to the python-level `polygonize` function, and in the `axis=None` case pass the flattened array to the ufunc.

src/ufuncs.c
jorisvandenbossche: Is it possible that this (`destroy_geom_arr(ctx, geom_arr, i - 1);`) is only needed for functions that release the GIL where we store the output in a temporary C array `geom_arr` which we only in the end transfer to a numpy array. 

While for the non-threaded versions, I suppose we directly write in the output object-dtype array, and numpy takes care of cleaning up this array? 
(in any case, I don't see such a manual clean-up in eg the geometry creation functions (linestrings, linearrings, polygons_with_holes, etc)

jorisvandenbossche:
Thanks for the feedback!

src/ufuncs.c
caspervdw: You're right, this is not necessary. All geometries in `geoms` are borrowed from Python because they come directly from `get_geom`

pygeos/constructive.py
caspervdw: I don' t think it is necessary to mimic this `axis=None` behaviour for generalized ufuncs. This is more something for numpy to implement. But forwarding `**kwargs` and documenting `axis` would be a good thing.

jorisvandenbossche: >  I encountered precisely what you are suggesting in `axis=None` at the `union_all` implementation: ... maybe good to do that after all for internal API consistency

On the other hand, you could also see `polygonize` as similar to `pygeos.polygons()` (but where the input doesn't need to be strictly rings), and there the default is `axis=0`. 
Now, it can certainly be seen both ways, given that the inputs are actually "processed" in polygonize, it's maybe more alike `union_all`

jorisvandenbossche: I already started with documenting the keyword. Can still add the `axis=None` functionality if we want.

pygeos/constructive.py
```diff
@@ -486,6 +487,36 @@ def point_on_surface(geometry, **kwargs):
     return lib.point_on_surface(geometry, **kwargs)
 
 
+def polygonize(geometries, **kwargs):
+    """Creates polygons formed from the linework of a set of Geometries.
+
+    Polygonizes an array of Geometries that contain linework which
+    represents the edges of a planar graph. All types of Geometry are
+    accepted as input; the constituent linework is extracted as the edges
```
brendan-ward, line 495: Suggestion:
```
Any type of Geometry may be provided as input; only lines 
will be used to create the output polygons.
```

pygeos/constructive.py
```diff
@@ -486,6 +487,36 @@ def point_on_surface(geometry, **kwargs):
     return lib.point_on_surface(geometry, **kwargs)
 
 
+def polygonize(geometries, **kwargs):
+    """Creates polygons formed from the linework of a set of Geometries.
+
```
brendan-ward, line 492: The return value of `GeometryCollection` is a bit unexpected (though based on what GEOS provides), so it would be good to note this specifically.

```
Returns Polygons within a GeometryCollection.  
Individual Polygons can be obtained using get_parts, 
and MultiPolygons can be constructed from the output using
multipolygons(get_parts(polygonize(geometries)))


See also
--------
get_parts
```

(suggestion to use `multipolygons` and `get_parts` may be too specific, but I'd want an easy way to get outputs as Polygons or MultiPolygons)

brendan-ward (commented):
A few comments on the docstring...

pygeos/constructive.py
jorisvandenbossche: @brendan-ward the original text is certainly not the best worded, but I think "only lines will be used to create output polygons" is a bit misleading, since all the rings (i.e. lines) from polygons will be used as well

pygeos/constructive.py
jorisvandenbossche: Yes, indeed, that's useful to highlight that the return is a collection and how to deal with it. See updated docstring.

pygeos/test/test_constructive.py
```diff
@@ -391,3 +391,79 @@ def test_clip_by_rect_non_scalar_kwargs():
...
+    assert arr.shape == (3, 2, 3)
+    result = pygeos.polygonize(arr)
+    assert isinstance(result, np.ndarray)
+    assert result.shape == (3, 2)
+    for res in result.flatten():
+        assert res == expected
+
+
+@pytest.mark.skipif(
+    np.__version__ < "1.15", reason="axis keyword for gufunc introduced in np 1.15"
```
jorisvandenbossche, line 451: https://numpy.org/devdocs/release/1.15.0-notes.html#generalized-ufuncs-now-accept-axes-axis-and-keepdims-arguments

pygeos/constructive.py
brendan-ward: Perhaps "only lines or rings..." would clarify?

Mostly what I was going for was to try and clarify that any types can be passed in but only parts of them are used; just a slightly different phrasing to in particular change "all types" to "any type".

pygeos/test/test_constructive.py
brendan-ward: Suggestion `gufunc` => `generalized ufunc`

pygeos/constructive.py
brendan-ward: It would be good to add a bit more to the docstring to clarify output behavior, something like:
```
Lines or rings that when combined do not completely close a polygon 
will result in an empty GeometryCollection.  Duplicate segments are 
ignored.
```
(this happens for the first one or two of your example lines below)

and 

```
Any input arrays of lines or rings must have the same shape: each 
polygon to be created must have the same number input lines or 
rings.
```
If you try to pass in a ragged array (e.g., array with first subarray 1 element long and second subarray 2 elements long), you get back `TypeError: One of the arguments is of incorrect type. Please provide only Geometry objects.` which is not particularly helpful in this case.  Perhaps since that may be a common error, specifically validating shape of the inputs would be helpful?

jorisvandenbossche:
We should also expose GEOSNode to make this more useful (https://postgis.net/docs/ST_Node.html)

caspervdw (approved):

pygeos/constructive.py
jorisvandenbossche: > Perhaps since that may be a common error, specifically validating shape of the inputs would be helpful?

We can't simply check the shape of the input, since you can also pass a 1D array to get a scalar output. So for a 1D array, we would need to check the actual values (whether it are geometries or arrays), which will become a bit costly I think.

jorisvandenbossche: I already updated the docstring for your first suggestions about only lines/rings being used as input. 

I am still struggling a bit to find a good wording this suggestion:

```
Any input arrays of lines or rings must have the same shape: each 
polygon to be created must have the same number input lines or 
rings.
```

I also find that a bit confusing, since its sounds as it is talking about multiple arrays, while it's actually about a 2D array?

(now, in general I don't think the fact that this is a ufunc is going to be very useful. I expect that most people will only use this with a 1D array)

brendan-ward: Feel free to ignore any confusing suggestions. 😄 

We can always refine the docstring later as we better identify where users are getting confused.  Don't let the docstring suggestions here hold things up.

jorisvandenbossche: Rereading your previous comment, I think I now better understand the point you were trying to make. 
Probably easiest to see if you pass a list of arrays/lists, and for example using one of the other constructive functions:

```
In [4]: pygeos.linestrings([[(0, 1), (2, 3)], [(0, 1), (2, 3)]])
Out[4]: 
array([<pygeos.Geometry LINESTRING (0 1, 2 3)>,
       <pygeos.Geometry LINESTRING (0 1, 2 3)>], dtype=object)

In [5]: pygeos.linestrings([[(0, 1), (2, 3)], [(0, 1), (2, 3), (4, 5)]])
...
ValueError: linestrings: Input operand 0 does not have enough dimensions (has 1, gufunc core with signature (i, d)->() requires 2)
```

A different error message, but also not an informative one, and similar in the underlying reason: if you pass a '"ragged" array, it's not a proper multidimensional array but rather a 1D array of wrong objects, and you get an error. 

But since the other constructive functions have a similar "trap", maybe can leave this for a separate issue to think about how we can make this more user friendly? 
(and then will merge this, as I have a follow-up on this for `polygonize_full`)

PR #276, opened 2021-01-13
ENH: Add fixed precision set operations for GEOS >= 3.9

brendan-ward:
This PR adds support for the fixed precision variants for set operations introduced in GEOS 3.9.

In `ufuncs.c`, these were copied and adapted from the `YY_Y` functions to handle an extra scalar parameter.  Since these needed to be version-guarded in C, it seemed reasonable to call into these only if the Python wrapper function determines that the `grid_size` parameter is non-None and GEOS >= 3.9; otherwise it falls back to the original functions.

The `grid_size` is scalar for now; this seems like a reasonable simplification because mixed-precision within an array seems a bit awkward.

It looks like there are a few errors due to mixed dimension GeometryCollections not being accepted in the new fixed precision functions.  I think this is a known issue in GEOS / JTS, but need to track down the cross-references.

Note: this is built against #257


brendan-ward:
Based on my understanding of `ufunc.reduce()`, it doesn't look possible to pass in the `grid_size` parameter in order to add fixed precision to the set operations that use reduce: `difference_all`, `intersection_all`, `symmetric_difference_all`.

Luckily `union_all` uses a different approach, so it was easily adaptable to use fixed precision.

May need to postpone adding fixed precision to the other *_all functions to a later PR.

jorisvandenbossche:
Cool, thanks a lot!

I suppose the easiest is first to get #257 in, so this gets a bit easier to review? (is there actually a reason it built on the other PR, it doesn't seem to depend on it?)

> The `grid_size` is scalar for now; this seems like a reasonable simplification because mixed-precision within an array seems a bit awkward.

+1

> May need to postpone adding fixed precision to the other *_all functions to a later PR.

Yes, that seems fine to leave for a follow-up. Having the grid_size option is the plain element-wise ufuncs would already be great!

jorisvandenbossche:
One user question that comes up: doing a `set_precision(geom, grid_size=1)` followed by a normal `intersection`, does that give the same result as doing an `intersection(...., grid_size=1)` ?

brendan-ward:
GEOS ticket for mixed dimension overlay errors: https://trac.osgeo.org/geos/ticket/1092

brendan-ward:
> is there actually a reason it built on the other PR, it doesn't seem to depend on it

I was using the other PR to test (manually) if the precision was set on the output geometries after using one of these fixed-precision operations.  It is not set (e.g., the precision of the result of union with `grid_size=1` is 0, even though results have been rounded).  Does this seem like something we should raise with GEOS?  The documentation there is not specific on this front.

> doing a set_precision(geom, grid_size=1) followed by a normal intersection, does that give the same result as doing an intersection(...., grid_size=1)

Based on simple tests, these are the same.  My understanding is that for GEOS >= 3.9, calling `set_precision` will use the same precision model in GEOS as the fixed-precision overlay functions - so the effect to the inputs of the operation should be the same.  It is also my understanding that if a grid_size was set for an input geometry to the set operations, that they would automatically use whichever precision is higher of the inputs, so in this case of both are set to 1 the result should be the same as if the `grid_size=1` parameter is used for the set operation.

brendan-ward:
Benchmarks (*_prec1 denotes `grid_size=1` and *_prec2 denotes `grid_size=2`):


```
time_difference                                    11.3±0.2ms
time_difference_prec1                                29.1±1ms
time_difference_prec2                              30.1±0.6ms
time_intersection                                  12.6±0.6ms
time_intersection_prec1                              30.4±4ms
time_intersection_prec2                            29.4±0.4ms
time_symmetric_difference                          12.0±0.3ms
time_symmetric_difference_prec1                      30.6±2ms
time_symmetric_difference_prec2                      33.3±5ms
time_union                                         12.1±0.6ms
time_union_all                                       33.6±7ms
time_union_all_prec1                                 76.7±8ms
time_union_all_prec2                                82.0±10ms
time_union_prec1                                   28.0±0.7ms
time_union_prec2                                     31.0±2ms
```

brendan-ward:
Benchmark test geometry is randomly generated to look something like this (transparency to show overlap):
image

pygeos/set_operations.py
```diff
@@ -16,45 +16,87 @@
...
+    If grid_size is nonzero, input coordinates will be snapped to a precision grid of that
+    size and resulting coordinates will be snapped to that same grid.  If 0, this
+    operation will use double precision coordinates.
+
     Parameters
     ----------
     a : Geometry or array_like
     b : Geometry or array_like
+    grid_size : float, optional (default: None).
+        precision grid size; requires GEOS >= 3.9.0.
```
jorisvandenbossche, line 31: Maybe mention here what the default means?

jorisvandenbossche, line 31:
```suggestion
    grid_size : float, optional (default: None)
        Precision grid size; requires GEOS >= 3.9.0.
```

(some formatting nitpicks)

pygeos/set_operations.py
```diff
@@ -16,45 +16,87 @@
...
     --------
+    >>> from pygeos.constructive import normalize
     >>> line = Geometry("LINESTRING (0 0, 2 2)")
     >>> difference(line, Geometry("LINESTRING (1 1, 3 3)"))
     <pygeos.Geometry LINESTRING (0 0, 1 1)>
     >>> difference(line, Geometry("LINESTRING EMPTY"))
     <pygeos.Geometry LINESTRING (0 0, 2 2)>
     >>> difference(line, None) is None
     True
+    >>> normalize(difference(box(0,0,2,2),box(1,1,3,3)))
```
jorisvandenbossche, line 43:
```suggestion
    >>> normalize(difference(box(0, 0, 2, 2), box(1, 1, 3, 3)))
```

(PEP8 nitpick)

pygeos/set_operations.py
```diff
@@ -16,45 +16,87 @@
...
+    <pygeos.Geometry POLYGON ((0 0, 0 2, 1 2, 1 1, 2 1, 2 0, 0 0))>
+    >>> normalize(difference(box(0.1,0.2,2.1,2.1),box(1,1,3,3), grid_size=1)) # doctest: +SKIP
```
jorisvandenbossche, line 45: Is the doctest skip still needed with adding the `normalize`? (or if we keep skipping it, the normalize can be removed)

pygeos/set_operations.py
```diff
@@ -16,45 +16,87 @@
 ]
 
 @multithreading_enabled
-def difference(a, b, **kwargs):
+def difference(a, b, grid_size=None, **kwargs):
     """Returns the part of geometry A that does not intersect with geometry B.
 
+    If grid_size is nonzero, input coordinates will be snapped to a precision grid of that
```
jorisvandenbossche, line 22: Can you limit the line length here to 80? (black doesn't fix this in a docstring, since this one is even longer as 88, and in addition for docstring readability (and eg how it is displayed in jupyter popups) it is good to even limit to 80 for docstrings)

pygeos/test/test_set_operations.py
```diff
@@ -41,6 +50,35 @@ def test_set_operation_array(a, func):
...
+    with pytest.raises(
+        ValueError, match="grid_size parameter only accepts scalar values"
+    ):
+        func(point, point, grid_size=[1])
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 9, 0), reason="GEOS < 3.9")
+@pytest.mark.parametrize("a", all_single_types)
+@pytest.mark.parametrize("func", SET_OPERATIONS)
+@pytest.mark.parametrize("grid_size", [0])
```
jorisvandenbossche, line 75: Parametrize with something else as 0 as well here?

pygeos/test/test_set_operations.py
```diff
@@ -105,6 +143,92 @@ def test_set_operation_reduce_all_none_arr(n, func, related_func):
...
+    assert pygeos.equals(actual, expected)
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 9, 0), reason="GEOS < 3.9")
+@pytest.mark.parametrize("n", range(1, 3))
+@pytest.mark.parametrize("func, related_func", REDUCE_SET_OPERATIONS_PREC)
+def test_set_operation_prec_reduce_all_none(n, func, related_func):
+    # API change: before, union_all([None]) yielded EMPTY GEOMETRYCOLLECTION
+    # The new behaviour is that it returns None if all inputs are None.
+    assert func([None] * n, grid_size=1) is None
```
jorisvandenbossche, line 229: Are those None-handling tests strictly needed? Because that's implemented on the python side and is the same regardless of passing grid_size or not? (now, implementation can change in the future, of course)

pygeos/test/test_set_operations.py
```diff
@@ -41,6 +50,35 @@ def test_set_operation_array(a, func):
...
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 9, 0), reason="GEOS < 3.9")
+@pytest.mark.parametrize("a", all_single_types)
+@pytest.mark.parametrize("func", SET_OPERATIONS)
+@pytest.mark.parametrize("grid_size", [0])
+def test_set_operation_prec_array(a, func, grid_size):
+    actual = func([a, a], point, grid_size=grid_size)
+    assert actual.shape == (2,)
+    assert isinstance(actual[0], Geometry)
```
jorisvandenbossche, line 79: Maybe also assert that the precision of the result is equal to grid size? (so we actually are sure we properly passed the grid size param, as I suppose this test would pass as well at the moment if the implementation had a mistake there)

(Eg for union_all below you have a few cases with actual input vs expected based on grid_size, which ensures things work as expected)

pygeos/test/test_set_operations.py
```diff
@@ -168,3 +292,43 @@ def test_coverage_union_non_polygon_inputs(geom_1, geom_2):
...
+            [pygeos.box(0.1, 0.1, 5, 5), pygeos.box(0, 0.2, 5.1, 10)],
+            0,
+            pygeos.Geometry(
+                "POLYGON ((0 0.2, 0 10, 5.1 10, 5.1 0.2, 5 0.2, 5 0.1, 0.1 0.1, 0.1 0.2, 0 0.2))"
+            ),
+        ),
+        # grid_size is at effective precision, expect no change
+        (
+            [pygeos.box(0.1, 0.1, 5, 5), pygeos.box(0, 0.2, 5.1, 10)],
+            0,
```
jorisvandenbossche, line 312:
```suggestion
            0.1,
```

? 

(otherwise it is the same test as the one above)

pygeos/test/test_set_operations.py
```diff
@@ -168,3 +292,43 @@ def test_coverage_union_non_polygon_inputs(geom_1, geom_2):
...
+            1,
+            pygeos.Geometry("POLYGON ((0 5, 0 10, 5 10, 5 5, 5 0, 0 0, 0 5))"),
+        ),
+        # grid_size much larger than effective precision causes rounding to nearest
+        # multiple of 10
+        (
+            [pygeos.box(0.1, 0.1, 5, 5), pygeos.box(0, 0.2, 5.1, 10)],
+            10,
+            pygeos.Geometry('POLYGON ((0 10, 10 10, 10 0, 0 0, 0 10))')
+        )
```
jorisvandenbossche, line 329: What happens with eg `grid_size=100` (where everything would be rounded to 0, so result collapses). Empty polygon? (similar as set_precision does I assume)

src/ufuncs.c
```diff
@@ -1151,6 +1156,157 @@ static void YYd_d_func(char** args, npy_intp* dimensions, npy_intp* steps, void*
...
+        // Keep in2 as 'outcome' of the operation.
+        out = in2;
+        // Ensure that it will not be destroyed (it is owned by python)
+        do_destroy = 0;
+      }
+
+      // 4. (not NULL, NULL); When a None value is encountered after a not-None
+      //    Don't do `out = in1`, as that is already the case.
+      // 5. (NULL, NULL); When we have not yet encountered any not-None
+      //    Do nothing; out will remain NULL
```
jorisvandenbossche, line 1235: I am wondering if we could somehow reduce the duplication between `YYd_Y_func_reduce` and `YY_Y_func_reduce` ... (but my C skills are not enough to directly "see" it, except for some ugly templating, or a separate function that does the inner part and receives a flag to know if it should check in3or not or something ..) 

(now if we want this, can certainly be left for later)

jorisvandenbossche (commented):
Looks good!

Mainly a bunch of minor docstring comments (I made them on the first, but they also apply for the other docstrings), and some comments on the tests.

pygeos/set_operations.py
brendan-ward: We need the skip for GEOS < 3.9 (otherwise using `grid_size` raises exceptions); I did not find a way to conditionally enable some doctests based on GEOS version.

pygeos/test/test_set_operations.py
brendan-ward: Unfortunately, GEOS currently returns 0 for the precision of the outputs when `grid_size` is provided as a parameter here.

I added a test to compare the results to the output of the operation against geometries that have been set to the same `grid_size` in advance via `set_precision`.

pygeos/test/test_set_operations.py
brendan-ward: Yes.  Added this to the test since it is probably good to demonstrate it.

src/ufuncs.c
brendan-ward: Oops - I should have deleted `YYd_Y_func_reduce` because we cannot use the `reduce` functionality with the additional `grid_size` parameter.  Gone now.

pygeos/test/test_set_operations.py
brendan-ward: It is implemented at the C level instead of Python level, and since `grid_size` goes to a different C function, I wanted to verify that all the existing behavior was consistent.

pygeos/set_operations.py
jorisvandenbossche: > We need the skip for GEOS < 3.9 (otherwise using `grid_size` raises exceptions); I did not find a way to conditionally enable some doctests based on GEOS version.

Ah, yes of course. Skipping is fine then (we should maybe at some point consider to run the doctests only on a limited GEOS version range, eg only latest release). 

If we skip it, though, I would maybe still leave out the `normalize` to make the example easier to read.

pygeos/test/test_set_operations.py
jorisvandenbossche: > Unfortunately, GEOS currently returns 0 for the precision of the outputs when grid_size is provided as a parameter here.

Ah, that's interesting to know (it's probably not "unfortunately", but intentionally done that way in GEOS?). So that is then still a difference between doing a `set_precision` first vs specifying the `grid_size` keyword. When setting the precision, that gets propagated to the output, while passing the keyword, the grid size is only used "within" the operation, but not set on the output.

jorisvandenbossche (approved):
Thanks for the updates!

jorisvandenbossche:
(needs a merge of master, conflict with https://github.com/pygeos/pygeos/pull/266)

pygeos/test/test_set_operations.py
brendan-ward: Added clarifying note to docstring.  Not sure if this is intentional in GEOS.

jorisvandenbossche: Just checked on the geos chat channel, and Martin Davies confirmed that it is intentional

jorisvandenbossche: Quoting him "The idea is to give more flexibility in using precision for specific operations, without affecting the underlying geometry precision model"

caspervdw:
Thanks @brendan-ward and @jorisvandenbossche ! Merging right before the 0.9 release

PR #278, opened 2021-01-14
TST: fix from_shapely test for numpy 1.20

jorisvandenbossche:
Small mistake I did in https://github.com/pygeos/pygeos/pull/269


jorisvandenbossche:
This works now

PR #279, opened 2021-01-17
CI: add GEOS 3.9.0 build to linux CI

jorisvandenbossche:
GEOS 3.9.0 is already released a while, so adding it to the Linux github actions builds. 
In addition, it was added to conda-forge a few days ago, so I think it should automatically be picked up for the "GEOS latest" builds using the conda-forge channel in the conda github action builds.


jorisvandenbossche:
@caspervdw all passing here!

PR #283, opened 2021-01-21
DOC: fix style issue with numpydoc parameter names and sphinx_rtd_theme

jorisvandenbossche:
So I think the issue is due to a sphinx > 2 compatibility issue with numpydoc (https://github.com/numpy/numpydoc/issues/215). I included a workaround for that in the past in geopandas (https://github.com/geopandas/geopandas/pull/1299). 
But, it should normally now also be fixed in the rtd theme itself in the latest versions. The problem seems to be that we are getting an older version (I assume because rtd installs it from defaults). So testing if installing it in our specified env from conda-forge solves that.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,12 @@
+/*Copied from sphinx' basic.css to ensure the sphinx >2.0 docstrings are
+rendered somewhat properly (xref https://github.com/numpy/numpydoc/issues/215) */
+
+.classifier {
+    font-style: oblique;
+}
+
+.classifier:before {
+    font-style: normal;
+    margin: 0.5em;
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Nope, didn't solve it, because readthedocs forces the main channel, which then causes downgrades after our environment was created ..

They have experimental feature flags for this: https://docs.readthedocs.io/en/stable/guides/feature-flags.html#available-flags, but those needs to asked by contacting their support. 

So the custom css might be the easier short term fix

caspervdw:
Thanks @jorisvandenbossche

caspervdw:
pygeos.readthedocs.io looks good again!

PR #285, opened 2021-01-26
Update ASV configuration

jorisvandenbossche:
Add cython, which is now needed to build pygeos, and update the python version. I also set conda-forge as channel, because defaults is typically a bit lagging behind (eg in GEOS version)


brendan-ward (approved):
Thanks @jorisvandenbossche !

PR #286, opened 2021-01-26
Change default STRtree leaf lize (node capacity) to 10

jorisvandenbossche:
I ran our STRtree benchmarks, but it didn't give any significant change (it might be that our benchmarks are too small / too "dummy" random geometries to notice a difference, if there would be one)

This came up in Shapely, where this was actually done wrong (set at size of geometry array, https://github.com/Toblerity/Shapely/pull/1064#discussion_r562376957). While this is not the case here, @dbaston mentioned that based on his experiments leaf size of 8 to 16 was optimal, and GEOS internally defauls to 10. So unless we have a good reason to deviate (@caspervdw?), probably best to just follow GEOS' default.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,11 @@ Version 0.10 (unreleased)
 
 * ...
 
+**API Changes**
+
+* STRtree default leaf size is now 10 instead of 5, for somewhat better performance
+  under normal conditions.
+
 **Added GEOS functions**
 
 * Addition of a ``contains_properly`` function (#267).
--- a/pygeos/strtree.py
+++ b/pygeos/strtree.py
@@ -51,7 +51,7 @@ class STRtree:
     [[0, 0, 0, 1, 1], [2, 3, 4, 5, 6]]
     """
 
-    def __init__(self, geometries, leafsize=5):
+    def __init__(self, geometries, leafsize=10):
         self.geometries = np.asarray(geometries, dtype=np.object_)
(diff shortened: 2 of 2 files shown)


brendan-ward:
I tested this with some messy real-world global polygon data (~270 query multipolygons, ~20k target multipolygons), using `STRtree.query_bulk` for a variety of leaf sizes (200 times each):

```
4: 3.10s
6: 2.74s
8: 2.60s
10: 2.86s
12: 3.23s
14: 2.52s
16: 2.75s
18: 2.60s
20: 2.76s
```

And the 20k polygons against themselves (there are also many non-self overlaps); 100 times each:

```
4: 6.12s
6: 4.97s
8: 5.45s
10: 4.80s
12: 5.08s
14: 5.09s
16: 5.23s
18: 5.29s
20: 5.36s
```

The results are more variable with respect to leaf size than I would have expected outright but look to support the 8-16 range with a default of 10.

brendan-ward (approved):
Thanks @jorisvandenbossche - this seems like a reasonable default and early tests indicate a performance improvement with real-world data.

jorisvandenbossche:
Thanks for testing Brendan!

caspervdw (approved):

brendan-ward:
Thanks @jorisvandenbossche !  I took the liberty of adding this to the changelog in order to merge.

jorisvandenbossche:
Thanks!

PR #288, opened 2021-01-29
DOC: clarify overlaps versus crosses predicates

### LINKED ISSUE #287: DOC: Linestring and Polygon with overlaps vs crosses
Based on the documentation for [overlaps](https://pygeos.readthedocs.io/en/latest/predicates.html#pygeos.predicates.overlaps)
> Returns True if A and B intersect, but one does not completely contain the other.

I am not sure why but `overlaps` returns `False` for this case while `crosses` returns `True` ? 
image
Is it expected, maybe because one of the points of the linestring is not inside the polygon?

I can provide proper coordinates if that case is interesting.

brendan-ward:
resolves #287 

The term `dimension` is prominent here (area, line, or point), but may cause confusion compared to coordinate dimensions (2D or 3D).  Does it need further clarification here?

### DIFF
--- a/pygeos/predicates.py
+++ b/pygeos/predicates.py
@@ -369,13 +369,11 @@ def is_valid_reason(geometry, **kwargs):
 
 @multithreading_enabled
 def crosses(a, b, **kwargs):
-    """Returns True if the intersection of two geometries spatially crosses.
+    """Returns True if A and B spatially cross.
 
-    That is: the geometries have some, but not all interior points in common.
-    The geometries must intersect and the intersection must have a
-    dimensionality less than the maximum dimension of the two input geometries.
-    Additionally, the intersection of the two geometries must not equal either
-    of the source geometries.
+    A crosses B if they have some but not all interior points in common,
+    the intersection is one dimension less than the maximum dimension of A or B,
+    and the intersection is not equal to either A or B.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


pygeos/predicates.py
```diff
@@ -637,8 +635,12 @@ def intersects(a, b, **kwargs):
...
 @multithreading_enabled
 def overlaps(a, b, **kwargs):
-    """Returns True if A and B intersect, but one does not completely contain
-    the other.
+    """Returns True if A and B spatially overlap.
+
+    A and B overlap if they have some but not all points in common, have the
+    same dimension, and the intersection of the interiors of the two geometries
+    has the same dimension as the geometries themselves.  That is, only polyons
+    can overlap other polygons and only lines can overlap other lines.
```
jorisvandenbossche, line 643: And eg lines that only cross at a single point don't overlap? (that might be useful to add as well to illustrate the "same dimension" rule?)

pygeos/predicates.py
```diff
@@ -650,14 +652,23 @@ def overlaps(a, b, **kwargs):
 
     Examples
     --------
-    >>> line = Geometry("LINESTRING(0 0, 1 1)")
-    >>> overlaps(line, line)
+    >>> poly = Geometry("POLYGON ((0 0, 0 4, 4 4, 4 0, 0 0))")
+    >>> overlaps(poly, poly)
+    False
+    >>> overlaps(poly, Geometry("POLYGON ((0 0, 0 2, 2 2, 2 0, 0 0))"))
     False
```
jorisvandenbossche, line 659: I think we should put some hints here about the answer, like:

```suggestion
    A polygon that is fully contained does not overlap:

    >>> overlaps(poly, Geometry("POLYGON ((0 0, 0 2, 2 2, 2 0, 0 0))"))
    False
```

and maybe similar for some others that might not be directly clear

jorisvandenbossche (commented):
Thanks!

caspervdw:
Is this good to merge?

brendan-ward:
This is done from my opinion, provided the updated docstrings are sufficiently clear to others.

caspervdw:
Thanks @brendan-ward , the docstrings are much better now.

Issue #241, opened 2020-11-10
RFC: Bulk geometry creation


caspervdw:
There have been quite some issues revolving around the geometry-creation functions:

#138 - Create geometries with different number of points in a vectorized way
#149 - Allow creating empty geometries in constructive functions

And the inverse of this, disassembling geometries into ndarrays:

#75 - ENH: get coordinates + offset arrays for "ragged array" representation
#93 - Add cython algo to get offset arrays for flat coordinates representation
#127 - Function to "flatten" or "explode" multi-geometries
#128 - Function to return all parts of a multi-geometry as ndarray
#197 - Cython + implement `get_parts`

The current geometry creation API was modeled after shapely's constructors. Now that we are integrating shapely and pygeos, I think this similarity will become less important. Shapely will keep wrapping the low level functions of pygeos (or later: shapely itself).
So at this point I think it is a good idea to make an RFC for the geometry creating functions in pygeos.

**General**
It mostly boils down to: how to you represent a ragged array in numpy? I think we have two options:

- An ndarray of python lists. While not optimal in terms of performance, it is still sufficiently performant for pygeos to support this. Lists are actually not that bad.
- We have discussed the (coordinates, indices) approach extensively in the context of `get_parts` (see #128). This seems to be the most optimal approach.

Other considerations:
- We can easily support multiple creation algorithms.
- Each function needs an inverse (like `get_parts`). Naming proposal: prefix the function with `unpack_`, and change `from` to ` to`.
- We have to consider 2D and 3D
- We have to consider empty geometry creation

**Points**
*Create an n-dimensional array of 2D/3D points:*
Call signature: `points(coordinates : ndarray[shape=(..., 2/3), dtype=float])`
```
>>> points(np.array([[0, 0], [0, 1]]))
ndarray([<Geometry POINT (0 0)>, <Geometry POINT (0 1)>])
>>> points(np.array([[0, 0, 0], [0, 1, 1]]))
ndarray([<Geometry POINT Z (0 0 0)>, <Geometry POINT Z (0 1 1)>])
```
*Create an n-dimensional array of mixed points from lists*
Call signature: `points_from_lists(coordinates : ndarray[dtype=list])`
```
>>> points_from_lists([[0, 0], [0, 1, 1], []]))
ndarray([<Geometry POINT (0 0)>, <Geometry POINT Z (0 1 1)>, <Geometry POINT EMPTY>])
```

**Linestrings and linearrings**
[...]

PR #290, opened 2021-01-31
Accept indices in collection constructors

caspervdw:
The `collections_1d` is the reverse of the `get_parts` that was implemented previously. This feature was proposed in RFC #241 .

There are some things to figure out still (better error messaging, docstrings, tests) but I think that this could be a blueprint for the other proposed functions that take  geometries & indices (=ragged array) and return a single 1D array of 'collections'.


pygeos/_bulk_creation.pyx
```diff
@@ -0,0 +1,93 @@
...
+    PyGEOS_GetGEOSGeometry
+)
+
+# initialize PyGEOS C API
+import_pygeos_c_api()
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def collections_1d(object[:] geometries, int[:] indices, int geom_type = 7, int ndim = 2):
```
caspervdw, line 31: Note that `indices` is expected to be sorted.

brendan-ward (commented):
Thanks for working on this @caspervdw.

Apologies for the long delay in review.  I think this provides a nice start at the inverse of `get_parts` and provides a model for how to work with ragged inputs.

I added few comments below to consider.  I think perhaps the most substantive change is to calculate the counts of geometries per output collection before the main loop, then use that as the basis for the inner loop.  I think this will simplify handling of the temporary geometries.

pygeos/_bulk_creation.pyx
```diff
@@ -0,0 +1,140 @@
...
+        # return immediately if there are no geometries to return
+        return np.empty(shape=(0, ), dtype=np.object_)  
+
+    if np.any(indices[1:] < indices[:-1]):
+        raise ValueError("The indices array should be sorted.")  
+
+    # A temporary array for the geometries that will be given to CreateCollection.
+    # Its size equals n_geoms, which is much too large in most cases. But it will
+    # give overhead to investigate the optimal size at this point.
+    temp_geoms = np.empty(shape=(n_geoms, ), dtype=np.intp)
```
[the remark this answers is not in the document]
caspervdw: I like the idea, but after doing some timings I decided to go for `np.bincount` instead of your suggestion.

```
In [19]: %timeit np.unique(arr, return_counts=True)[1]
1.97 ms ± 12 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [16]: def counts_2(arr):
    ...:     run_start = np.r_[1, arr[:-1] != arr[1:]]
    ...:     return np.diff(np.r_[np.nonzero(run_start)[0], len(arr)])
In [20]: %timeit counts_2(arr)
643 µs ± 1.09 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [30]: %timeit np.bincount(arr)
239 µs ± 377 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

It has the added advantage of returning `0` for indices that do not occur:

```
np.bincount([1, 1, 5, 5])
Out[31]: array([0, 2, 0, 0, 0, 2])
```

pygeos/_bulk_creation.pyx
```diff
@@ -0,0 +1,140 @@
...
+    else:
+        raise ValueError(f"Invalid geometry_type: {geometry_type}.")
+
+    if n_geoms == 0:
+        # return immediately if there are no geometries to return
+        return np.empty(shape=(0, ), dtype=np.object_)  
+
+    if np.any(indices[1:] < indices[:-1]):
+        raise ValueError("The indices array should be sorted.")  
+
```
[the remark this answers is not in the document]
caspervdw: If they are not incremental, then that is fine, we'll just get 1 collection.

caspervdw:
Thanks for your feedback @brendan-ward . The counting before the loop makes sure that I can make a more tight temporary buffer, but for the rest, I don't see any simplification of the loop. Especially because we still skip missing values.

pygeos/_bulk_creation.pyx
```diff
@@ -0,0 +1,144 @@
...
+                        raise TypeError(
+                            f"One of the arguments has unexpected geometry type {curr_type}."
+                        )
+
+                # ignore missing values
+                if geom == NULL:
+                    n_missing += 1
+                else:
+                    # assign to the temporary geometry array
+                    temp_geoms_view[this_geom_idx] = <np.intp_t>GEOSGeom_clone_r(geos_handle, geom)
```
brendan-ward, line 128: `this_geom_idx` is the position within the inner loop, but shouldn't be used for saving this into the array because it doesn't account for missing geometries.

Instead, I'd suggest using another counter that is incremented as you set non-empty geometries into `temp_geoms_view` and then use that value as the length of the geometry collection for `GEOSGeom_createCollection_r` below.


Here is a test case to add to `test_collections_1d` for that (currently segfaults):
```python
([point, None, line_string], [0, 0, 0], [geom_coll([point, line_string])])
```

caspervdw:
Thanks @brendan-ward  for your feedback, I went over it once more and renamed the various integer variables in the cython function to make it more understandable.

Also, I decided to diverge from the RFC I wrote and instead of exposing a new function, incorporate this into the existing `multipoints`, `multilinestrings` etc. functions. What do you think about this approach? Also pinging @jorisvandenbossche

pygeos/_geometry.pyx
```diff
@@ -75,3 +79,114 @@ def get_parts(object[:] array):
...
     return parts, index
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def collections_1d(
+    object geometries,
+    object indices,
+    int geometry_type = 7,
+    int ndim = 2
```
brendan-ward, line 90: I don't see that `ndim` is ever passed in from the Python side?  Should this parameter be removed?

pygeos/_geometry.pyx
```diff
@@ -75,3 +79,114 @@ def get_parts(object[:] array):
...
+                    if curr_type != expected_type and curr_type != expected_type_alt:
+                        # deallocate previous temp geometries (preventing memory leaks)
+                        for coll_geom_idx in range(coll_size):
+                            GEOSGeom_destroy_r(geos_handle, <GEOSGeometry *>temp_geoms_view[coll_geom_idx])
+                        raise TypeError(
+                            f"One of the arguments has unexpected geometry type {curr_type}."
+                        )
+
+                # ignore missing values
+                if geom != NULL:
```
brendan-ward, line 175: The `NULL` check should come above line 163 so that we don't call `GEOSGeomTypeId_r` with a `NULL` geometry.

pygeos/_geometry.pyx
```diff
@@ -75,3 +79,114 @@ def get_parts(object[:] array):
...
+                    # assign to the temporary geometry array                   
+                    temp_geoms_view[coll_size] = <np.intp_t>GEOSGeom_clone_r(geos_handle, geom)
+                    coll_size += 1
+
+            # create the collection
+            coll = GEOSGeom_createCollection_r(
+                geos_handle,
+                geometry_type, 
+                <GEOSGeometry**> &temp_geoms_view[0],
+                <unsigned int>coll_size
```
brendan-ward, line 185: To avoid the type cast here, is it possible to define this as an `unsigned int` up front?

pygeos/_geometry.pyx
```diff
@@ -75,3 +79,114 @@ def get_parts(object[:] array):
...
+        raise TypeError("indices is not a one-dimensional array.")
+
+    if geometries.shape[0] != indices.shape[0]:
+        raise ValueError("geometries and indices do not have equal size.")
+
+    if geometries.shape[0] == 0:
+        # return immediately if there are no geometries to return
+        return np.empty(shape=(0, ), dtype=np.object_)
+
+    if np.any(indices[1:] < indices[:-1]):
```
brendan-ward, line 130: This is raising a Cython warning
> warning: pygeos/_geometry.pyx:130:38: the result of using negative indices inside of code sections marked as 'wraparound=False' is undefined

Probably need to avoid using `@cython.wraparound(False)`.

brendan-ward (changes requested):
A few changes needed.

I think it works well to handle this via parameters to the existing geometry construction functions; 👍 on the API change.

brendan-ward (approved):
Thanks for the updates!  It needs an entry in the changelog, but after that it is ready to merge.

PR #295, opened 2021-02-03
Update pin for numpy version for Python 3.9 in pyproject.toml

### LINKED ISSUE #294: Intermittent numpy-related test failures
We're getting this error `numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject` on a few recent PR CI runs, unrelated to the substance of those PRs.

Appears limited to our runner for GEOS 3.9 / numpy 1.19.5 / linux runners.

Examples:
* [https://github.com/pygeos/pygeos/runs/1824297876?check_suite_focus=true](https://github.com/pygeos/pygeos/runs/1824297876?check_suite_focus=true)
* [https://github.com/pygeos/pygeos/runs/1801237559?check_suite_focus=true](https://github.com/pygeos/pygeos/runs/1801237559?check_suite_focus=true)

We're not alone, this appears to be happening in `scikit-learn`'s [hdbscan](https://github.com/scikit-learn-contrib/hdbscan/issues/457) as well.

jorisvandenbossche:
@brendan-ward I _think_ this should fix #294

Now that numpy 1.20 is released, we have the problem that in the Python 3.9 build we pinned numpy to 1.19 in the environment, but pygeos itself was still built against the latest (so a newer) numpy. And when it comes to building, numpy is forward compatible (built with older version will work with newer), but not the other way around

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,8 @@ requires = [
     "numpy==1.13.3; python_version=='3.6'",
     "numpy==1.14.5; python_version=='3.7'",
     "numpy==1.17.3; python_version=='3.8'",
+    "numpy==1.19.3; python_version=='3.9'",
     # do not pin numpy on future versions of python to avoid incompatible numpy and python versions
-    "numpy; python_version>='3.9'",
+    "numpy; python_version>='3.10'",
     "Cython>=0.29",
 ]


pyproject.toml
```diff
@@ -7,7 +7,8 @@ requires = [
     "numpy==1.13.3; python_version=='3.6'",
     "numpy==1.14.5; python_version=='3.7'",
     "numpy==1.17.3; python_version=='3.8'",
+    "numpy==1.19.3; python_version=='3.9'",
```
jorisvandenbossche, line 10: 1.19.3 is also what is used at https://github.com/scipy/oldest-supported-numpy/commit/aa14bc2b75ed09865e20a91d5ee578a3493f7dd1

brendan-ward (approved):
Thanks for working on this!  I can confirm that bumping the CI numpy version to `1.20.0` also avoids this ([CI run here](https://github.com/brendan-ward/pygeos/runs/1825364031?check_suite_focus=true)) - but better to pin the build dependency.  (I often forget the build deps will install a different version even if the dep is already installed)

jorisvandenbossche:
> but better to pin the build dependency.

Yes, that's something we need to do anyway, unrelated from our CI (to ensure users installing from source/sdist don't run into this). 

Now, it would be good to actually have a CI build using numpy 1.20, though, certainly given it had quite some substantial changes (although I suppose the conda-based CI builds using "latest" will currently test against 1.20 already, so that might be good enough)

brendan-ward:
The Python 3.9 / GEOS master / numpy latest is [using numpy 1.20](https://github.com/pygeos/pygeos/runs/1825498403?check_suite_focus=true#step:6:56) as well, so that is covered for now - but also since we allow that build to fail with lower severity, it might be good to add a CI run with numpy 1.20 if we start consistently letting that build fail due to very recent changes in GEOS or elsewhere.

jorisvandenbossche:
Ah, and we could also use that GEOS master build to test against pre-releases of numpy (so another reason to not rely on it for 1.20 testing)

PR #293, opened 2021-02-03
Fix signed char portability issue

brendan-ward:
Resolves #291 

We were using a `char` with a -1 sentinel value, which fails on ARM architectures because it is unsigned.  This PR assigns it to an `int` which is the proper return value for `GEOSGeomTypeId_r` in this case anyway.

Issue #291, opened 2021-02-02
Use of signed char fails on ARM architectures


girgink:
Hi,

On NVIDIA Jetson AGX (ARMv8) I do a fresh install of pygeos using Docker:

```
docker run -it --user root --rm arm64v8/ubuntu:20.04

apt-get update
apt-get install -y python3-dev python3-pip libgeos-dev
pip3 install --no-cache-dir cython 
pip3 install --no-cache-dir numpy
pip3 install --no-cache-dir --no-build-isolation pygeos
```

The following [example given in the documentation](https://pygeos.readthedocs.io/en/latest/set_operations.html) for `union_all()` method results in TypeError: 

```
import pygeos

line_1 = pygeos.Geometry("LINESTRING(0 0, 2 2)")
line_2 = pygeos.Geometry("LINESTRING(2 2, 3 3)")
pygeos.union_all([line_1, line_2])
```

Error message:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/pygeos/decorators.py", line 55, in wrapped
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/pygeos/set_operations.py", line 353, in union_all
    collections = lib.create_collection(geometries, GeometryType.GEOMETRYCOLLECTION)
TypeError: One of the Geometry inputs is of incorrect geometry type.
```

girgink:
Tests also fail with the same error if I install from the repository:

```
apt-get install git
pip3 install --no-cache-dir pytest
cd /tmp
git clone https://github.com/pygeos/pygeos.git
cd pygeos
pip3 install -e .[test] --no-build-isolation
pytest pygeos
```

Output:

```
=============================================== test session starts ================================================
platform linux -- Python 3.8.5, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /tmp/pygeos
collected 18 items / 10 errors / 8 selected

====================================================== ERRORS ======================================================
________________________________ ERROR collecting pygeos/test/test_constructive.py _________________________________
pygeos/test/test_constructive.py:7: in <module>
    from .common import (
pygeos/test/common.py:22: in <module>
    geometry_collection = pygeos.geometrycollections(
pygeos/creation.py:174: in geometrycollections
    return lib.create_collection(geometries, GeometryType.GEOMETRYCOLLECTION)
[...]

PR #296, opened 2021-02-04
Add travis for different CPU archs

caspervdw:


jorisvandenbossche:
Cool! 

* We will need to migrate the account to travis-ci.com (from .org), see the warning on the travis page
* As mentioned in https://github.com/pygeos/pygeos/issues/291#issuecomment-772305610, we might need to only run this nightly or on demand if we want to keep it within the free limits of Travis (or contact Travis to ask for more)

caspervdw:
OK I migrated to travis-ci.com , switched off the PR and commit triggers, and configured a daily run.

Not sure how this migration will take effect though.

jorisvandenbossche:
It's a bit unfortunate that Travis has no way to trigger a build on a PR on demand (AFAIK). 
For example for github actions, whether it runs for a PR or only nightly is coded inside the yml, so you can always activate it inside the PR where you want to check this by editing the yml file (and remove again before merging).

Eg for #293 we probably want to run the tests to ensure they pass with the patch.

caspervdw:
The total build time is now about 4 minutes. The `--no-build-isolation` prevented a numpy build, which saves a lot of time.

The free credits are (if I understand it correctly) 1000 minutes/month , so we have about 250 builds per month. I think not enough to build on every PR, but we may switch build on every commit to master?

jorisvandenbossche:
Yes, every commit on master certainly sounds good I think

PR #298, opened 2021-02-08
ENH: add polygonize_full

jorisvandenbossche:
Follow-up on https://github.com/pygeos/pygeos/pull/275, adding the "full" version.

Current behaviour: returns 4 collections (polygons, dangles, cut edges, invalid rings):

```
In [17]: lines = np.array([pygeos.Geometry("LINESTRING (0 0, 1 1)"), pygeos.Geometry("LINESTRING (0 0, 0 1, 1 1)"), pygeos.Geometry("LINESTRING (0 1, 1 1)")])

In [18]: pygeos.lib.polygonize_full(lines)
Out[18]: 
(<pygeos.Geometry GEOMETRYCOLLECTION (POLYGON ((1 1, 0 0, 0 1, 1 1)))>,
 <pygeos.Geometry GEOMETRYCOLLECTION EMPTY>,
 <pygeos.Geometry GEOMETRYCOLLECTION (LINESTRING (0 1, 1 1))>,
 <pygeos.Geometry GEOMETRYCOLLECTION EMPTY>)
```

### DIFF
--- a/pygeos/test/test_constructive.py
+++ b/pygeos/test/test_constructive.py
@@ -482,10 +482,12 @@ def test_polygonize_missing():
 
 def test_polygonize_full():
     lines = [
+        None,
         pygeos.Geometry("LINESTRING (0 0, 1 1)"),
         pygeos.Geometry("LINESTRING (0 0, 0 1)"),
         pygeos.Geometry("LINESTRING (0 1, 1 1)"),
         pygeos.Geometry("LINESTRING (1 1, 1 0)"),
+        None,
         pygeos.Geometry("LINESTRING (1 0, 0 0)"),
         pygeos.Geometry("LINESTRING (5 5, 6 6)"),
         pygeos.Geometry("LINESTRING (1 1, 100 100)"),
--- a/pygeos/test/test_constructive.py
+++ b/pygeos/test/test_constructive.py
@@ -521,12 +521,18 @@ def test_polygonize_full_array():
     assert len(result) == 4
     assert all(isinstance(geom, pygeos.Geometry) for geom in result)
     assert result[0] == expected
+    assert all(
+        geom == pygeos.Geometry("GEOMETRYCOLLECTION EMPTY") for geom in result[1:]
+    )
 
     result = pygeos.polygonize_full(np.array([lines]))
     assert len(result) == 4
     assert all(isinstance(geom, np.ndarray) for geom in result)
     assert all(geom.shape == (1,) for geom in result)
     assert result[0][0] == expected
+    assert all(
+        geom[0] == pygeos.Geometry("GEOMETRYCOLLECTION EMPTY") for geom in result[1:]
+    )
 
     arr = np.array([lines, lines])
     assert arr.shape == (2, 3)
@@ -536,6 +542,10 @@ def test_polygonize_full_array():
     assert all(arr.shape == (2,) for arr in result)
     assert result[0][0] == expected
     assert result[0][1] == expected
+    assert all(
+        g == pygeos.Geometry("GEOMETRYCOLLECTION EMPTY")
+        for geom in result[1:] for g in geom
+    )
 
     arr = np.array([[lines, lines], [lines, lines], [lines, lines]])
     assert arr.shape == (3, 2, 3)
@@ -545,6 +555,9 @@ def test_polygonize_full_array():
     assert all(arr.shape == (3, 2) for arr in result)
     for res in result[0].flatten():
         assert res == expected
+    for arr in result[1:]:
+        for res in arr.flatten():
+            assert res == pygeos.Geometry("GEOMETRYCOLLECTION EMPTY")
 
 
 @pytest.mark.skipif(
--- a/pygeos/test/test_constructive.py
+++ b/pygeos/test/test_constructive.py
@@ -513,9 +513,7 @@ def test_polygonize_full_array():
         pygeos.Geometry("LINESTRING (0 0, 0 1)"),
         pygeos.Geometry("LINESTRING (0 1, 1 1)"),
     ]
-    expected = pygeos.Geometry(
-        "GEOMETRYCOLLECTION (POLYGON ((1 1, 0 0, 0 1, 1 1)))"
-    )
+    expected = pygeos.Geometry("GEOMETRYCOLLECTION (POLYGON ((1 1, 0 0, 0 1, 1 1)))")
     result = pygeos.polygonize_full(np.array(lines))
     assert len(result) == 4
     assert all(isinstance(geom, pygeos.Geometry) for geom in result)
@@ -543,7 +541,8 @@ def test_polygonize_full_array():
     assert result[0][1] == expected
     assert all(
         g == pygeos.Geometry("GEOMETRYCOLLECTION EMPTY")
-        for geom in result[1:] for g in geom
+        for geom in result[1:]
+        for g in geom
     )
 
     arr = np.array([[lines, lines], [lines, lines], [lines, lines]])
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
Hmm seems to be segfaulting with older numpy (<= 1.15)

jorisvandenbossche:
@brendan-ward @caspervdw Aside from that I need to update this PR / check the failures with older numpy, what are your thoughts on the API here? 
Currently the function returns 4 arrays (or 4 scalars for 1D input), so having a generalized ufunc signature of "(d)->(),(),(),()" (but in theory it could also be a single array with first/last dimension of shape 4, I suppose, but not fully sure how that would be done in C)

caspervdw:
> Aside from that I need to update this PR / check the failures with older numpy, what are your thoughts on the API here?

The API looks good as it is now, I think there should indeed be separate 4 return values, because they differ semantically.

jorisvandenbossche:
Turns out it was not failing on builds with older numpy (<= 1.15), but it was actually failing with older GEOS (<= 3.7), but both constraints result in the same failing builds ..

And the segfault seems to be related to the memory management of the intermediate collection object that is created.

src/ufuncs.c
```diff
@@ -1844,6 +1850,71 @@ static void polygonize_func(char** args, npy_intp* dimensions, npy_intp* steps,
...
+        GEOSPolygonize_full_r(ctx, collection, &cuts, &dangles, &invalidRings);
+    if (ret_ptr == NULL) {
+      errstate = PGERR_GEOS_EXCEPTION;
+      goto finish;
+    }
+    OUTPUT_Y_I(1, ret_ptr);
+    OUTPUT_Y_I(2, cuts);
+    OUTPUT_Y_I(3, dangles);
+    OUTPUT_Y_I(4, invalidRings);
+    // GEOSGeom_destroy_r(ctx, collection);
```
jorisvandenbossche, line 1903: @caspervdw So I expected that I would need to destroy the collection that is created here (and not returned), but doing so results in segfaults on clean-up / del.

Now, I suppose it is related to this aspect of `GEOSGeom_createCollection`:

> The caller **retains ownership** of the containing array,
> but the ownership of the pointed-to objects is transferred
> to the returned \ref GEOSGeometry.

So the problem here is that `GEOSGeom_destroy_r(ctx, collection)` also destroys the geometries that were put into the collection. But those geometries were the values passed to the ufunc, and thus should not be destroyed. How to solve this?

* Clone all input geometries before putting them in the collection, so the full collection can be destroyed (but that seems wasteful)
* How do I destroy the collection object without destroying the pointed-to objects? Or is that simply not needed (because the collection is nothing else as pointing to those underlying objects)?

caspervdw: I think the first option you state is the only viable one. `CreateCollection` transfers ownership of the geometries, so as soon as you use it on geometries which are also owned by a `pygeos.Geometry`, the geometries will have two owners, which is wrong.

Skipping the destroy of the `collection` will lead to a memory leak. So copying is the only way.

pygeos/test/test_constructive.py
```diff
@@ -480,6 +480,99 @@ def test_polygonize_missing():
...
+        pygeos.Geometry("LINESTRING (0 1, 1 1)"),
+    ]
+    expected = pygeos.Geometry(
+        "GEOMETRYCOLLECTION (POLYGON ((1 1, 0 0, 0 1, 1 1)))"
+    )
+    result = pygeos.polygonize_full(np.array(lines))
+    assert len(result) == 4
+    assert all(isinstance(geom, pygeos.Geometry) for geom in result)
+    assert result[0] == expected
+
```
brendan-ward, line 524: Here and below, might be good to assert that the last 3 parts of the result are empty geometry collections.

pygeos/test/test_constructive.py
```diff
@@ -480,6 +480,99 @@ def test_polygonize_missing():
...
+    arr = np.array([lines, lines])  # shape (2, 3)
+    result = pygeos.polygonize_full(arr, axis=1)
+    assert len(result) == 4
+    assert all(arr.shape == (2,) for arr in result)
+    result = pygeos.polygonize_full(arr, axis=0)
+    assert len(result) == 4
+    assert all(arr.shape == (3,) for arr in result)
+
+
+def test_polygonize_full_missing():
```
brendan-ward, line 569: It is probably a good idea to add another test case or cases that include inputs that are mixed valid geometries and `None` to be sure that `None`s are properly ignored.

brendan-ward (commented):
A few comments to consider to verify that `None`s in mixed inputs are handled properly, and otherwise agree that geometries need to be cloned before creating the intermediate collection.

pygeos/test/test_constructive.py
jorisvandenbossche: Added

pygeos/test/test_constructive.py
jorisvandenbossche: That should already be covered by `test_polygonize_full` (which has a `None` in the input as well), I think?

src/ufuncs.c
jorisvandenbossche: OK, updated with cloning all input before creating collections

pygeos/test/test_constructive.py
brendan-ward: Adding a test with `None` in the first position followed by valid geometries and possibly (though overkill) one in between two valid geometries is probably a good idea; this has helped us catch segfaults in the past (when geometries added to a malloc'd array are first tested for `NULL` before inserting into a position in that array)

brendan-ward: Yes, adding another `None` to `test_polygonize_full` would be just fine.

src/ufuncs.c
```diff
@@ -1844,6 +1850,78 @@ static void polygonize_func(char** args, npy_intp* dimensions, npy_intp* steps,
...
+        errstate = PGERR_NOT_A_GEOMETRY;
+        goto finish;
+      }
+      if (geom == NULL) {
+        continue;
+      }
+      // need to copy the input geometries, because the Collection takes ownership
+      geom_copy = GEOSGeom_clone_r(ctx, geom);
+      if (geom_copy == NULL) {
+        errstate = PGERR_GEOS_EXCEPTION;
```
brendan-ward, line 1887: There may be an edge case here if `GEOSGeom_clone_r` fails partway through.  There will be some geometry objects in `geoms` (up to `n_geoms`) that need to be destroyed before freeing the array.

To be clear, this should only be done if things fail at this step.  As soon as ownership of the cloned geometries transfers to the new GeometryCollection, things are OK (they are properly deleted on failure of creation of GeometryCollection)

jorisvandenbossche: I added some code for this, but I don't think this will actually be covered? (I am not sure how a clone can fail, except something like out of memory)

brendan-ward: Agreed - not sure how to make it fail - but seems good from a defensive programming standpoint even if we can't get coverage for this.

It would be possible to proof this with a manual test: just fail if `i == 2` in order to verify that the implementaion is correct, then take that back out (it looks correct).

jorisvandenbossche: Making the check above `if ((geom_copy == NULL) | (n_geoms == 2)) {` gives a "correct" GEOSException in python, so the destroy code should have run fine, I suppose.

brendan-ward (approved):
Looks good, thanks @jorisvandenbossche

caspervdw (approved):
Looks good! Maybe document the `**kwargs` as done in #327 , else it will probably be forgotten.

jorisvandenbossche:
Thanks @caspervdw and @brendan-ward for the reviews here!

Issue #258, opened 2020-11-24
PROPOSAL: Add densifier


snowman2:
https://geos.osgeo.org/doxygen/classgeos_1_1geom_1_1util_1_1Densifier.html

jorisvandenbossche:
This isn't exposed in the GEOS C API, so this first needs an enhancement (request) at the GEOS side.

snowman2:
https://trac.osgeo.org/geos/ticket/1077

PR #299, opened 2021-02-09
ENH: Add densify ufunc

brendan-ward:
This adds `GEOSDensify`, which will be available when GEOS 3.10.0 is released; currently available on GEOS `master`.

crossref to Shapely issue: https://github.com/Toblerity/Shapely/issues/1037

/cc @snowman2

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,13 +11,13 @@ Version 0.10 (unreleased)
 **API Changes**
 
 * STRtree default leaf size is now 10 instead of 5, for somewhat better performance
-  under normal conditions (#286).
+  under normal conditions (#286)
 
 **Added GEOS functions**
 
-* Addition of a ``contains_properly`` function (#267).
-* Addition of a ``polygonize`` function (#275).
-* Addition of a ``densify`` function.
+* Addition of a ``contains_properly`` function (#267)
+* Addition of a ``polygonize`` function (#275)
+* Addition of a ``densify`` function (#299)
 
 **Bug fixes**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,7 +17,7 @@ Version 0.10 (unreleased)
 
 * Addition of a ``contains_properly`` function (#267)
 * Addition of a ``polygonize`` function (#275)
-* Addition of a ``densify`` function for GEOS >= 3.10 (#299)
+* Addition of a ``segmentize`` function for GEOS >= 3.10 (#299)
 
 **Bug fixes**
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.rst
```diff
@@ -11,14 +11,16 @@ Version 0.10 (unreleased)
...
-  under normal conditions.
+  under normal conditions (#286)
 
 **Added GEOS functions**
 
-* Addition of a ``contains_properly`` function (#267).
-* Addition of a ``polygonize`` function (#275).
+* Addition of a ``contains_properly`` function (#267)
+* Addition of a ``polygonize`` function (#275)
+* Addition of a ``densify`` function (#299)
```
jorisvandenbossche, line 20: maybe mention that it's for GEOS 3.10+ only?

jorisvandenbossche (commented):
PostGIS calls this `segmentize` (https://postgis.net/docs/ST_Segmentize.html, which I personally find a clearer name). Was there any discussion on the GEOS PR about naming?

brendan-ward:
There was no discussion in the GEOS PR about naming.  It exposes operations against the `Densifier` C++ class (same naming in JTS as well), so I think "densify" was aligned with those terms, whereas "segmentize" would have been foreign.

Personally, I find "segmentize" less clear since "densify" implies increasing the density of points (in a linear geometry), whereas "segmentize" (i.e., make into segments) implies breaking a LineString into segments and returning individual segments (also would be useful later on).

brendan-ward:
There was no discussion in the GEOS PR about naming.  It exposes operations against the `Densifier` C++ class (same naming in JTS as well), so I think "densify" was aligned with those terms, whereas "segmentize" would have been foreign.

Personally, I find "segmentize" less clear since "densify" implies increasing the density of points (in a linear geometry), whereas "segmentize" (i.e., make into segments) implies breaking a LineString into segments and returning individual segments (also would be useful later on).

caspervdw:
Naming things is one of the hardest parts. In my opinion, both names are not concise, but that is the problem if you only have one word. Next best thing is aligning to what people know.

I'd like to follow PostGIS, this is what we have always done in pygeos (e.g. line_interpolate_point). End users do not care about namings of internal C++ classes, while they will probably know PostGIS.

caspervdw (commented):
Looks good, only the name of the function is still under discussion

brendan-ward:
Possible consistency issue: `frechet_distance` and `hausdorff_distance` include a `densify` parameter that looks to do something similar to this stand-alone function:
> The parameter 'densify' makes this approximation less coarse by splitting
    the line segments between vertices before computing the distance.

Does that make us want to rethink changing `densify` => `segmentize`?

snowman2:
Side note: This closes #258. I created duplicate issues in pygeos and shapely.

caspervdw (approved):

PR #301, opened 2021-02-12
TST: Handle GEOS version strings with dev / beta qualifiers in the tests

brendan-ward:
We are getting version test failures on GEOS master or latest on the `3.9.1` branch because the version string includes modifiers like `dev`.

This PR is one way to resolve that within C, by building up the string from the integer parts.  I'm not sure I like this option; it has the risk of being a tautology (i.e., version string is no longer a validity check against the tuple of version parts).

Do we want the version string to reflect the original string from GEOS with these modifiers?  This seems like it might be useful from a troubleshooting perspective.  In which case, we just need to modify the tests so that they ignore these modifiers, but we allow the versions to be a bit different: (3,9,1) => `3.9.1dev`

### DIFF
--- a/pygeos/test/test_misc.py
+++ b/pygeos/test/test_misc.py
@@ -1,4 +1,4 @@
-from string import ascii_letters
+from string import ascii_lowercase
 
 import pygeos
 from pygeos.decorators import requires_geos, multithreading_enabled
@@ -25,7 +25,7 @@ def test_geos_version():
     actual = pygeos.geos_version_string
 
     # strip any beta / dev qualifiers
-    actual = actual.lower().rstrip(ascii_letters)
+    actual = actual.lower().rstrip(ascii_lowercase)
 
     assert actual == expected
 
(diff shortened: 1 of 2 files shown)


brendan-ward:
If we strip "dev" off from the string, then we start getting warnings about mismatched versions vs shapely in GeoPandas:

```
The Shapely GEOS version (3.10.0dev-CAPI-1.15.0) is incompatible with the GEOS version PyGEOS was compiled with (3.10.0-CAPI-1.15.0)
```

Seems like the correct thing to do here might be to leave the qualifiers in the string version and fix our test cases to allow these.

brendan-ward:
@caspervdw Do we want to strip version qualifiers (as here), or modify our tests to allow them for the GEOS version string?

I'm leaning toward modifying the tests but wanted to be sure before I refactor this and get it merged soon; this is causing failures across our other PRs.

caspervdw:
I agree with your preference, leave the version strings as supplied by GEOS and fix our tests.

jorisvandenbossche:
+1 on updating our tests

pygeos/test/test_misc.py
```diff
@@ -20,7 +20,12 @@ def test_version():
 
 def test_geos_version():
     expected = "{0:d}.{1:d}.{2:d}".format(*pygeos.geos_version)
-    assert pygeos.geos_version_string == expected
+    actual = pygeos.geos_version_string
+
+    # strip any beta / dev qualifiers
+    actual = actual.lower().rstrip("abcdefghijklmnopqrstuvwxyz")
```
mwtoews, line 26: one suggestion is to use [string constants](https://docs.python.org/3/library/string.html#string.ascii_letters) via `from string import ascii_letters` at the top, then:
```suggestion
    actual = actual.rstrip(ascii_letters)
```
similar suggestion for below around line 46

jorisvandenbossche (approved):

PR #304, opened 2021-02-26
NaN equality tests for GEOS 3.10

caspervdw:
See #233 for a discussion. For GEOS 3.9 and before, we were able to construct a POINT (nan nan). In GEOS master, this is not possible anymore and it transforms to a POINT EMPTY. This caused some of our tests to fail.

These tests just use a point because it is the simplest geometry. Now I updated the tests so that they use linestrings and everything passes on my machine.

### DIFF
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -32,6 +32,9 @@ def _wrap_construct_ufunc(func, coords, y=None, z=None):
 def points(coords, y=None, z=None):
     """Create an array of points.
 
+    Note that GEOS >=3.10 automatically converts POINT (nan nan) to
+    POINT EMPTY.
+
     Parameters
     ----------
     coords : array_like
--- a/pygeos/test/common.py
+++ b/pygeos/test/common.py
@@ -33,7 +33,7 @@ empty_point = pygeos.Geometry("POINT EMPTY")
 empty_line_string = pygeos.Geometry("LINESTRING EMPTY")
 empty_polygon = pygeos.Geometry("POLYGON EMPTY")
 empty = pygeos.Geometry("GEOMETRYCOLLECTION EMPTY")
-point_nan = pygeos.points(np.nan, np.nan)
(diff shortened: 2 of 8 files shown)


jorisvandenbossche (approved):
Change itself looks good!

I am wondering if it would still be good to have a specific test for the behaviour of `pygeos.points(..)` with NaN coordinates. Because while the change in behaviour of GEOS is fine to me, it's still something that users might actually use, and good to keep that way in the future.

caspervdw:
I added the test as you suggested.

Also I used preprocessor `!GEOS_SINCE_3_10_0` directives to disable my self-built POINT EMPTY to POINT (nan nan) conversion on GEOS >=3.10. Everything still checks out.

brendan-ward:
I'm having a little bit of a hard time following the logic for version-guarding (via `!GEOS_SINCE_3_10_0`) some things but not others.

For example, why are `geometrycollection_has_point_empty` and `has_point_empty` version-guarded but not `multipoint_has_point_empty`?  Perhaps more can be added to their individual docstrings to indicate that they are only needed for GEOS < 3.10.0?  The leading comment above the version-guard is about conversion of points, so it isn't as obvious as to how that applies to these conditional tests.

caspervdw:
> I'm having a little bit of a hard time following the logic for version-guarding (via `!GEOS_SINCE_3_10_0`) some things but not others.
> 
> For example, why are `geometrycollection_has_point_empty` and `has_point_empty` version-guarded but not `multipoint_has_point_empty`? Perhaps more can be added to their individual docstrings to indicate that they are only needed for GEOS < 3.10.0? The leading comment above the version-guard is about conversion of points, so it isn't as obvious as to how that applies to these conditional tests.

These functions are also used for fixing an issue with empty points in WKT strings, which has not been fixed in GEOS master yet. So that is why I pulled `multipoint_has_point_empty` outside of the version conditional.

Issue #297, opened 2021-02-08
TST: Tests fail on big-endian systems


brendan-ward:
The `s390` runner is big-endian, whereas everything else we are testing is little-endian.  Our WKB tests expect results in little-endian order, and fail on inequality ([ref](https://travis-ci.com/github/pygeos/pygeos/jobs/481129309)).

The simple solution may be to skip or xfail WKB tests on `s390`; otherwise we need to have a set of expected WKB output for each endianness.

caspervdw:
We can also force byte order in the tests by putting `to_wkb(..., byte_order=1)` ?

jorisvandenbossche:
There are also some doctest failures like this one:

```
063     Examples
064     --------
065     >>> from pygeos import normalize
066     >>> points = Geometry("MULTIPOINT (2 2, 4 2)")
067     >>> normalize(voronoi_polygons(points))
Expected:
    <pygeos.Geometry GEOMETRYCOLLECTION (POLYGON ((3 0, 3 4, 6 4, 6 0, 3 0)), PO...>
Got:
    <pygeos.Geometry GEOMETRYCOLLECTION EMPTY>
```

That might not be WKB related?

brendan-ward:
Correct - not all the failures on `s390` are necessarily related to WKB; this particular one will require a bit more investigation.

caspervdw:
The remainder can be divided into two categories:

## Failures on big endian systems with `voronoi_polygons`:

`voronoi_polygons` (`GEOSVoronoiDiagram_r`) fails in all cases:
```
>>> voronoi_polygons(Geometry("MULTIPOINT (2 2, 4 2)"))
<pygeos.Geometry GEOMETRYCOLLECTION EMPTY>
>>> voronoi_polygons(Geometry("LINESTRING (0 0, 1 0, 1 1)"))
E pygeos.GEOSException: TopologyException: Input geom 1 is invalid: Self-intersection at or near point -2.2558952938275252e+307 -2.2558952938275252e+307 at -2.2558952938275252e+307 -2.2558952938275252e+307
```
etc. The large numbers in the error message point in the direction of an assumed endianness somewhere in `GEOSVoronoiDiagram_r`

## Failures on big endian systems with predicates:

```
>>> disjoint(Geometry("LINESTRING(0 0, 1 1)"), Geometry("LINESTRING(0 2, 2 0)"))
True # expected False
[...]

PR #305, opened 2021-03-01
Pin endianness in to_wkb tests

caspervdw:
Partially addresses #297 

Unfortunately, it is quite elaborate to emulate s390x on linux, so testing occurs on Travis.
I can trigger this manually from a test branch, see https://travis-ci.com/github/caspervdw/pygeos/jobs/487227201

### DIFF
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -137,9 +137,9 @@ def to_wkb(
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


brendan-ward (approved):
Thanks for adding this!

(not sure yet what is driving the other failures on Travis, but they appear unrelated to WKB ordering so out of scope here)

PR #307, opened 2021-03-03
Also clean build/lib.*/pygeos in setup.py clean

caspervdw:
I have been scratching my head over linking issues over the past hour, and it appeared that `python setup.py clean` did not actually clean all the compiled libraries. Now it does.


jorisvandenbossche (approved):

jorisvandenbossche:
Yes, I noticed this before as well that I still needed to remove the build directory as well. +1 on removing it with the clean command.

PR #308, opened 2021-03-04
Add box ufunc based on create_box C function

### LINKED ISSUE #306: Handle GEOS errors on geometry constructors
Box fails if the second or third coordinates are NaN.

Steps to reproduce:

```
In [1]: import pygeos
In [5]: pygeos.box(1, 1, 2, float("nan"))                                                                                                                          
Out[5]: <pygeos.Geometry POLYGON ((2 1, 2 nan, 1 nan, 1 1, 2 1))>
In [6]: pygeos.box(float("nan"), 1, 2, 2)                                                                                                                          
Out[6]: <pygeos.Geometry POLYGON ((2 1, 2 2, nan 2, nan 1, 2 1))>
In [7]: pygeos.box(1, float("nan"), 2, 2)                                                                                                                          
In [2]: pygeos.box(1, 1, float("nan"), 2)                                                                                                                          
```

brendan-ward:
This adds a ufunc for `box` based on the `create_box` function introduced in #272  and replaces the prior implementation in Python.

Also resolves #306 

Also resolves a latent bug in `create_box` because it was testing for equality to `NP_NAN` (which always fails) instead of using `npy_isnan()`.

This is about 2x as fast as the prior implementation:

```
benchmarks.ConstructiveSuite.time_box                                        8.76±0.3ms
benchmarks.ConstructiveSuite.time_box_prev                                   16.1±0.4ms
```

Some tests were adjusted slightly; the polygons returned by `box` are now in normalized form because `create_box` creates them that way.

The only bit that was a little awkward was that errors are not thrown by `create_box`; it always returns `NULL` when geometries cannot be created: either because coordinates are `nan` or because one of the GEOS functions fails.  This means that if we get back `NULL` we have to check to see if any coordinates are `nan` - which is a valid result - or capture the error otherwise.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -48,7 +48,8 @@ class ConstructiveSuite:
     """Benchmarks constructive functions on a set of 10,000 points"""
 
     def setup(self):
-        self.points = pygeos.points(np.random.random((10000, 2)))
+        self.coords = np.random.random((10000, 2))
+        self.points = pygeos.points(self.coords)
 
     def time_voronoi_polygons(self):
         pygeos.voronoi_polygons(self.points)
@@ -61,6 +62,16 @@ class ConstructiveSuite:
 
     def time_delaunay_triangles(self):
         pygeos.delaunay_triangles(self.points)
+
+    def time_box(self):
+        pygeos.box(*np.hstack([self.coords, self.coords + 100]).T)
+
+    def time_box_prev(self):
+        # previous implementation of box
+        x1, y1, x2, y2 = np.hstack([self.coords, self.coords + 100]).T
+        rings = np.array(((x2, y1), (x2, y2), (x1, y2), (x1, y1)))
+        rings = rings.transpose(list(range(2, rings.ndim)) + [0, 1])
+        pygeos.polygons(rings)
 
 
 class ClipSuite:
@@ -247,12 +258,14 @@ class STRtree:
         # result
         l, r = self.grid_point_tree.nearest(self.grid_points)
         # calculate distance to nearest neighbor
-        dist = pygeos.distance(self.grid_points.take(l), self.grid_point_tree.geometries.take(r))
+        dist = pygeos.distance(
+            self.grid_points.take(l), self.grid_point_tree.geometries.take(r)
+        )
         # include a slight epsilon to ensure nearest are within this radius
         b = pygeos.buffer(self.grid_points, dist + 1e-8)
 
         # query the tree for others in the same buffer distance
-        left, right = self.grid_point_tree.query_bulk(b, predicate='intersects')
+        left, right = self.grid_point_tree.query_bulk(b, predicate="intersects")
         dist = pygeos.distance(
             self.grid_points.take(left), self.grid_point_tree.geometries.take(right)
         )
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -66,13 +66,6 @@ class ConstructiveSuite:
     def time_box(self):
         pygeos.box(*np.hstack([self.coords, self.coords + 100]).T)
 
-    def time_box_prev(self):
-        # previous implementation of box
-        x1, y1, x2, y2 = np.hstack([self.coords, self.coords + 100]).T
-        rings = np.array(((x2, y1), (x2, y2), (x1, y2), (x1, y1)))
-        rings = rings.transpose(list(range(2, rings.ndim)) + [0, 1])
-        pygeos.polygons(rings)
-
 
 class ClipSuite:
     """Benchmarks for different methods of clipping geometries by boxes"""
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -111,11 +111,7 @@ def box(x1, y1, x2, y2):
     x2 : array_like
     y2 : array_like
     """
-    x1, y1, x2, y2 = np.broadcast_arrays(x1, y1, x2, y2)
-    rings = np.array(((x2, y1), (x2, y2), (x1, y2), (x1, y1)))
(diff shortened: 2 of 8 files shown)


brendan-ward:
Per further clarification in #306 and discussion on GEOS IRC, in general we should assume that GEOS takes ownership of the geometry components (coordinate sequence, linear ring) that are passed to the constructor for higher-order geometries (e.g., polygon).  Thus we should not attempt to clean them up ourselves.

pygeos/creation.py
```diff
@@ -111,11 +111,7 @@ def box(x1, y1, x2, y2):
     x2 : array_like
     y2 : array_like
     """
-    x1, y1, x2, y2 = np.broadcast_arrays(x1, y1, x2, y2)
-    rings = np.array(((x2, y1), (x2, y2), (x1, y2), (x1, y1)))
-    # bring first two axes to the last two positions
-    rings = rings.transpose(list(range(2, rings.ndim)) + [0, 1])
-    return polygons(rings)
+    return lib.box(x1, y1, x2, y2)
```
caspervdw, line 114: Can you forward the `**kwargs` here? We do it with most ufuncs.

pygeos/test/test_creation.py
```diff
@@ -298,13 +298,27 @@ def test_create_collection_wrong_geom_type(func, sub_geom):
 
 def test_box():
     actual = pygeos.box(0, 0, 1, 1)
-    assert str(actual) == "POLYGON ((1 0, 1 1, 0 1, 0 0, 1 0))"
+    assert str(actual) == "POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))"
```
caspervdw, line 301: This is an API change that might be breaking to some applications. In the previous implementation it followed shapely (see here: https://shapely.readthedocs.io/en/stable/manual.html#shapely.geometry.box)

Did you have any reason to adapt the order of coordinates?

caspervdw (changes requested):
Thanks! Two suggestions, please have a look

pygeos/test/test_creation.py
jorisvandenbossche: I don't know the exact reason for shapely's behaviour, but I think returning the normalized version also makes sense:

```
In [3]: pygeos.normalize(pygeos.Geometry("POLYGON ((1 0, 1 1, 0 1, 0 0, 1 0))"))
Out[3]: <pygeos.Geometry POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))>
```

Although that's still something else ..

benchmarks/benchmarks.py
```diff
@@ -62,6 +63,16 @@ def time_convex_hull(self):
     def time_delaunay_triangles(self):
         pygeos.delaunay_triangles(self.points)
 
+    def time_box(self):
+        pygeos.box(*np.hstack([self.coords, self.coords + 100]).T)
+
+    def time_box_prev(self):
```
jorisvandenbossche, line 69: If we keep it, I would call it `time_box_from_rings` or something like that

pygeos/test/test_creation.py
jorisvandenbossche: Hmm, the output of normalize is actually no longer counter-clock-wise. I didn't expect that normalize would change that (at least not change to *not* be CCW).

BTW, Shapely has a `ccw=True` keyword (so which we should still add at some point for compatibility with shapely, unless we deprecate it there)

benchmarks/benchmarks.py
brendan-ward: We can delete it now.

pygeos/test/test_creation.py
brendan-ward: Whoops, right, normalized is clockwise.

No particular reason to change coordinate order other than was trying to start from the bottom left (minimum) coordinate similar to normalize.

Can easily change back to match shapely, no compelling reason to diverge from prior precedent here if that might break people.  Will also look into adding `ccw=True` parameter.

brendan-ward:
This now matches shapely as the canonical form for both orderings.

I updated the parameter names to indicate min vs max values, but used `xmin`, etc rather than shapely's `minx` since we use the former in a few other places but not the latter.

brendan-ward:
Does adding the `ccw` parameter justify an entry in the changelog?

caspervdw (changes requested):
I really like the addition of `ccw`, and that is it True by default (e.g. GeoJSON uses CCW by default)
One remark about the datatype. In other place we use `NPY_BOOL` for booleans.

src/ufuncs.c
```diff
@@ -1249,6 +1249,48 @@ static PyUFuncGenericFunction YYd_Y_funcs[1] = {&YYd_Y_func};
...
+
+  // allocate a temporary array to store output GEOSGeometry objects
+  geom_arr = malloc(sizeof(void*) * n);
+  CHECK_ALLOC(geom_arr);
+
+  GEOS_INIT_THREADS;
+
+  for (i = 0; i < n; i++, ip1 += is1, ip2 += is2, ip3 += is3, ip4 += is4, ip5 += is5) {
+    geom_arr[i] = create_box(ctx, *(double*)ip1, *(double*)ip2, *(double*)ip3,
+                             *(double*)ip4, *(int*)ip5);
```
caspervdw, line 1271: I guess you get a compiler warning here, it should be `char` or `npy_bool`:

```suggestion
                             *(double*)ip4, *(npy_bool*)ip5);
```

brendan-ward: Whoops!  Good catch, didn't see the warning.

caspervdw (approved):

caspervdw:
I think this deserves an entry in the changelog, for the rest, please feel free to merge.

PR #309, opened 2021-03-04
Add back pygeos.strtree.VALID_PREDICATES for now

jorisvandenbossche:
See https://github.com/pygeos/pygeos/pull/263#issuecomment-790463108. Removing it broke geopandas. I think it's easy to keep it a bit longer in pygeos for now (and we can still revert this before doing the next release)


brendan-ward (approved):
@jorisvandenbossche thanks for adding this back in.  In retrospect we should have deprecated rather than removed it outright.  I made a few minor changes to clarify that.

jorisvandenbossche:
Thanks for the clarification in the changelog!

PR #310, opened 2021-03-05
[Done] Fix creation error handling and release GIL

### LINKED ISSUE #306: Handle GEOS errors on geometry constructors

caspervdw:
Closes #306

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,8 @@ Version 0.10 (unreleased)
 * Deprecated ``VALID_PREDICATES`` set from ``pygeos.strtree`` package; these can be constructed
   in downstream libraries using the ``pygeos.strtree.BinaryPredicate`` enum.
   This will be removed in a future release.
+* ``points``, ``linestrings``, ``linearrings``, and ``polygons`` now return a ``GEOSException``
+  instead of a ``ValueError`` for invalid input.
 
 **Added GEOS functions**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,7 +9,7 @@ Version 0.10 (unreleased)
 * Addition of ``nearest`` and ``nearest_all`` functions to ``STRtree`` for
   GEOS >= 3.6 to find the nearest neighbors (#272).
 * Released GIL for ``points``, ``linestrings``, ``linearrings``, and 
-  (partially) ``polygons`` (#310).
+  ``polygons`` (without holes) (#310).
 
 **API Changes**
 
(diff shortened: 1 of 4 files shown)


brendan-ward (commented):
👍 more release of GIL and resolving segfaults.

This isn't the same as #308 and shouldn't close it; do you mean #306?

I don't see that there is a specific test of `nan` to ensure no segfaults per #306?  Might be a good idea to add one.

I'm still not 100% clear on what your handling of `nan` values should be; is this supposed to include `nan` coordinates?

```
>>> pygeos.linestrings([[0, 0], [np.nan, 1]])
<pygeos.Geometry LINESTRING (0 0, NaN 1)>

>>> pygeos.linearrings([[0, 0], [np.nan, 1], [2, np.nan]])
<pygeos.Geometry LINEARRING (0 0, NaN 1, 2 NaN, 0 0)>
```

That is consistent with points, unless all values are `nan`:
```
>>> pygeos.points([np.nan,1])
<pygeos.Geometry POINT (NaN 1)>
```

They all seem awkward from a usability standpoint.


However, linear features don't all coalesce on emtpy when all coordinates are `nan`; should they?

```
>>> pygeos.linestrings([[np.nan,np.nan], [np.nan,np.nan]])
<pygeos.Geometry LINESTRING (NaN NaN, NaN NaN)>
```

And we fail due to lack of comparability of `nan` for linear rings:
```
>>> pygeos.linearrings([[np.nan,np.nan], [np.nan,np.nan], [np.nan, np.nan], [np.nan,np.nan]])
...
pygeos.GEOSException: IllegalArgumentException: Points of LinearRing do not form a closed linestring
```

caspervdw:
Thanks for your feedback @brendan-ward . I addressed it and I think it is good to merge now.

It is true that we allow `nan` in the geometry constructors. It may be better to block this (and return `None` for geometries that have at least 1 `nan` coordinate) - or return EMPTY geometries. I don't know, I don't want to think too much for the end user.

Issue #119, opened 2020-03-21
BUG: from_shapely with PreparedGeometry segfaults


jorisvandenbossche:
Instead of raising a TypeError:

```
import pygeos 
import shapely.geometry  
import shapely.prepared 
p = shapely.geometry.Point(1, 1)
pp = shapely.prepared.prep(p)

In [7]: pp  
Out[7]: <shapely.prepared.PreparedGeometry at 0x7f02605e20d0>

In [9]: pygeos.from_shapely([pp])  
Out[9]: Segmentation fault (core dumped)
```

caspervdw:
Fixed in #312

PR #312, opened 2021-03-09
to_shapely and from_shapely updates

0phoff:
I noticed issue #54 was pushed to a next milestone and thus decided to put some work into it.
This PR adds a `to_shapely()` function and improves the `from_shapely()` function to also work when the GEOS versions of both library do not match, by going through the WKB format (similarly to how geopandas handles things).

This implementation is implemented purely in python, as I do not know how we should create shapely objects from C.
I still believe it is worth adding, as this fills a "gap" in the current pygeos pipeline, and we can still optimize it without affecting the public API, later down the road...

Remarks / TODO:
- [ ] Unit Test `to_shapely()` when GEOS versions are the same. This would mean mocking `shapely.geos.lgeos.GEOSGeom_clone()` and `shapely.geometry.base.geom_factory()`. I am not completely sure if it is worth it / necessary, as we would effectively be mocking 90% of the functionality...
- [X] ~Contrary to geopandas, these implementations do not work with the `__geo_interface__`. I don't have any experience with this interface and don't really know what to expect here, so I currently cannot work on this.~
  Let's keep the to_* and from_* functions simple.
- [X] ~Instead of using `shapely.geos.lgeos.GEOSGeom_clone()`, we could probably use the `get_geos_handle` and `GEOSGeom_clone_r` from _pygeos/\_geos.pxd_. This would mean either rewriting this function in cython or exposing this functionality in python (get_geos_handle should be quite trivial with cpdef, not sure about the geos functions). Again,  I am not completely sure if it is worth it / necessary.~
  Not implementing this for now. If anyone wants to try to get an even faster conversion, feel free...
- [X] ~Add `ignore_shapely2_warnings`, similarly to [GeoPandas](https://github.com/geopandas/geopandas/blob/master/geopandas/_compat.py#L140). I have not done this yet, as I do not use Shapely V2 and did not immediately see a need for this. Let me know if this is necessary and I can add it.~
  I do not think this is necessary, as Shapely V2 will contain PyGEOS, hence making this library unnecessary.
- [x] Add `to_shapely()` to documentation.

---

__Performance Note:__

At first, I implemented things more like geopandas, by having a single python loop and performing the necessary transformations in  that loop:
- pygeos -> wkb -> shapely
- shapely -> wkb -> pygeos

However, after performing some tests, I found out it is faster to perform the "pygeos -> wkb" / "wkb -> pygeos" seperately and run it on the entire array at once. Obviously, having the same GEOS version is still much faster!
The test also shows that `from_shapely()` is quite a bit faster when you have the same GEOS version, compared to `to_shapely()`. My guess is that this is because most of the compute happens in C, whilst `to_shapely()` is built with python.
You can find my test-code and charts below.

```python
# Note: I only tested with boxes, so maybe the results can be slightly different when having different geometries

# pygeos -> shapely
times = []
nums = []
for i in range(0, 5001, 50):
    if i == 0:
        box = None
    else:
        box = pygeos.box(0,0,range(i),range(i))
        
    result = %timeit -o pygeos.to_shapely(box)
    nums.append(i if i != 0 else 1)
    times.append(result.best)

# shapely -> pygeos
times = []
nums = []
for i in range(0, 5001, 50):
    if i == 0:
        box = None
    else:
        box = np.array([shapely.geometry.box(0,0,j,j) for j in range(i)], dtype='object')
        
    result = %timeit -o pygeos.from_shapely(box)
    nums.append(i if i != 0 else 1)
    times.append(result.best)
```

image
image

### DIFF
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -9,18 +9,39 @@ from . import geos_capi_version_string
 
 
 ShapelyGeometry = None
-shapely_compatible = None
 shapely_lgeos = None
 shapely_geom_factory = None
 shapely_wkb_loads = None
+shapely_compatible = None
 _shapely_checked = False
 
 def check_shapely_version():
+    """
+    This function will try to import shapely and extracts some necessary classes and functions from the package.
+    It also looks if Shapely and PyGEOS use the same GEOS version, as this means the conversion can be faster.
+
+    This function sets a few global variables:
+
+    - ShapelyGeometry: 
+        shapely.geometry.base.BaseGeometry
+    - shapely_lgeos: 
+        shapely.geos.lgeos
+    - shapely_geom_factory: 
+        shapely.geometry.base.geom_factory
+    - shapely_wkb_loads: 
+        shapely.wkb.loads
+    - shapely_compatible: 
+        ``None`` if shapely is not installed, 
+        ``True`` if shapely and PyGEOS use the same GEOS version, 
+        ``False`` otherwise
+    - _shapely_checked: 
+        Mostly internal variable to mark that we already tried to import shapely
+    """
     global ShapelyGeometry
-    global shapely_compatible
     global shapely_lgeos
     global shapely_geom_factory
     global shapely_wkb_loads
+    global shapely_compatible
     global _shapely_checked
 
     if not _shapely_checked:
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -30,21 +30,21 @@ def check_shapely_version():
 
     This function sets a few global variables:
 
-    - ShapelyGeometry: 
+    - ShapelyGeometry:
         shapely.geometry.base.BaseGeometry
-    - ShapelyPreparedGeometry: 
+    - ShapelyPreparedGeometry:
         shapely.prepared.PreparedGeometry
-    - shapely_lgeos: 
+    - shapely_lgeos:
         shapely.geos.lgeos
-    - shapely_geom_factory: 
+    - shapely_geom_factory:
         shapely.geometry.base.geom_factory
-    - shapely_wkb_loads: 
+    - shapely_wkb_loads:
         shapely.wkb.loads
-    - shapely_compatible: 
-        ``None`` if shapely is not installed, 
-        ``True`` if shapely and PyGEOS use the same GEOS version, 
+    - shapely_compatible:
+        ``None`` if shapely is not installed,
+        ``True`` if shapely and PyGEOS use the same GEOS version,
         ``False`` otherwise
-    - _shapely_checked: 
+    - _shapely_checked:
         Mostly internal variable to mark that we already tried to import shapely
     """
     global ShapelyGeometry
@@ -159,12 +159,7 @@ def to_wkt(
 
 
 def to_wkb(
-    geometry,
-    hex=False,
-    output_dimension=3,
-    byte_order=-1,
-    include_srid=False,
-    **kwargs
+    geometry, hex=False, output_dimension=3, byte_order=-1, include_srid=False, **kwargs
 ):
     r"""
     Converts to the Well-Known Binary (WKB) representation of a Geometry.
@@ -225,7 +220,7 @@ def to_wkb(
 def to_shapely(geometry):
     """
     Converts PyGEOS geometries to Shapely.
-    
+
     Parameters
     ----------
     geometry : shapely Geometry object or array_like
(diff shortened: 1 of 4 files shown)


pygeos/io.py
```diff
@@ -219,33 +296,47 @@ def from_shapely(geometry, **kwargs):
...
-        # we already have a numpy array or we are None
-        arr = geometry
-    return lib.from_shapely(arr, **kwargs)
+        unpack = geometry is None or isinstance(geometry, ShapelyGeometry)
+        if unpack:
+            geometry = (geometry,)
+
+        arr = [
+            None
+            if g is None
```
caspervdw, line 332: This double `if ... else`  in a list comprehension is a bit too long for my taste. 
Also I think that the "empty point" exception is so rare that this should be a case of "exception handling" instead of preemptive checking.

pygeos/io.py
```diff
@@ -219,33 +296,47 @@ def from_shapely(geometry, **kwargs):
...
-        # used, converting the Geometry to its coordinates
-        arr = np.empty(1, dtype=object)
-        arr[0] = geometry
-        arr.shape = ()
-    elif not isinstance(geometry, np.ndarray) and isinstance(geometry, Sized):
-        # geometry is a list/array-like
-        arr = np.empty(len(geometry), dtype=object)
-        arr[:] = geometry
+    if shapely_compatible:
+        if isinstance(geometry, ShapelyGeometry):
```
caspervdw, line 310: Have you tested what happens if `geometry`  is a shapely `PreparedGeometry` ? See https://github.com/pygeos/pygeos/issues/119

pygeos/io.py
```diff
@@ -7,26 +8,48 @@
...
+shapely_compatible = None
+shapely_lgeos = None
+shapely_geom_factory = None
+shapely_wkb_loads = None
 _shapely_checked = False
 
 def check_shapely_version():
-    global shapely_geos_version
     global ShapelyGeometry
+    global shapely_compatible
```
caspervdw, line 20: Can you document here that `shapely_compatible` can be `None`, `True` and `False`?

caspervdw (changes requested):
Thanks @0phoff  for this addition and you thorough analysis of the performance of the different options.

I had to think about this one for a while, because the efforts (tagging @jorisvandenbossche) of merging pygeos and shapely are under way for some time now. As soon as the repo's are merged, these functions have no reason to exist anymore. But until that time, I think this is a good addition.

Given the almost complete state of this PR, I think we could merge this still. What is your opinion @jorisvandenbossche ?

pygeos/io.py
0phoff: I can rewrite this part as a normal for loop. It will probably take a small performance hit as we would need to append to the list, and python cannot pre-allocate as it does with list comprehensions, but I don't think it will be that noticeable.

I tried using a try/except, but shapely somehow throws an `IllegalArgumentException` that I cannot catch:
``` python
>>> import shapely
>>> from shapely import wkt
>>> empty = wkt.loads('POINT EMPTY')
>>> try:
...   empty.wkb
... except:
...   print('caught exception')
IllegalArgumentException: Empty Points cannot be represented in WKB
```

pygeos/io.py
0phoff: I added some documentation to the `check_shapely_version()` function

pygeos/io.py
0phoff: I just rewrote this to a normal loop, as I felt it was necessary in order to work with prepared geometries.

pygeos/io.py
0phoff: I got it working for the case when the GEOS versions do not match, but I did not change anything to the existing C code.

I can look into this, but wanted some pointers on your end.
- in [ufunc.c](https://github.com/pygeos/pygeos/blob/master/src/ufuncs.c#L2454), I can check whether the object has a `context` attribute and use `obj.context.__geom__` in that case.
- Is there a better approach in C ? (eg. like `isinstance(obj, PreparedGeometry)`)

caspervdw (approved):
Thanks @0phoff  this looks good, I really like the thorough testing although I am starting to think it it might be a better idea to run the tests with shapely installed. I did that locally (removing the mocks) and everything checks out. So this is good to merge. Can you add a line to the changelog?

caspervdw:
@0phoff  I took the liberty of finishing this PR, thanks again for your contribution. I put your github username in the attribution list, if you want me to replace it with your full name, please tell me and I'll change it after.

jorisvandenbossche:
> Given the almost complete state of this PR, I think we could merge this still. What is your opinion @jorisvandenbossche ?

Sorry for the slow reply here. Yes, fine with this being merged on the short term, but I *hope* we will be able to remove this addition again sooner rather than later! ;)

@0phoff thank a lot for your contribution

PR #314, opened 2021-03-16
ENH: add oriented_envelope (minimum rotated rectangle)

martinfleis:
Exposing `GEOSMinimumRotatedRectangle_r`. 

I also have one interesting observation. The `minimum_rotated_rectangle` function in shapely is custom Python code, not GEOS and it actually returns different geometry, often smaller than what GEOS does.

```py
>>> import geopandas as gpd
>>> import pygeos
>>> pygeos.geos_version
(3, 9, 1)

>>> df = gpd.read_file(gpd.datasets.get_path('nybb'))
>>> geoms = df.geometry.values.data
>>> pygeos.area(pygeos.minimum_rotated_rectangle(geoms))
array([2.99661872e+09, 7.12609927e+09, 3.41384061e+09, 1.37383388e+09,
       2.19130492e+09])

>>> df.geometry.apply(lambda x: x.minimum_rotated_rectangle).area
0    2.941075e+09
1    6.623816e+09
2    3.351064e+09
3    1.373834e+09
4    2.074161e+09
dtype: float64
```
Shapely code: https://github.com/Toblerity/Shapely/blob/18fe1023822dc55d48f68e11ec79346f1deebd69/shapely/geometry/base.py#L484

And apparently, older GEOS returns slightly different geometries, which fails the docstring test. Should I just skip those?
```
Expected:
    <pygeos.Geometry LINESTRING (1 1, 10 1)>
Got:
    <pygeos.Geometry LINESTRING Z (1 1 0, 10 1 0)>
```

xref #126


ljwolf:
This'd be great; we use the [opencv2](https://github.com/pysal/pointpats/blob/master/pointpats/centrography.py#L79) implementation currently (as well as our [own Numba implementation of skyum's algorithm for the minimum bounding circle](https://github.com/pysal/pointpats/blob/master/pointpats/centrography.py#L323) to ref #315).

jorisvandenbossche:
The original Shapely PR: https://github.com/Toblerity/Shapely/pull/361

@martinfleis did you inspect the produced rectangles visually to see what's the difference?

martinfleis:
It is not a marginal difference.

```py
import pygeos
import geopandas as gpd
df = gpd.read_file(gpd.datasets.get_path('nybb'))
geoms = df.geometry.values.data
pyg = gpd.GeoSeries(pygeos.minimum_rotated_rectangle(geoms))
shp = df.geometry.apply(lambda x: x.minimum_rotated_rectangle)
ax = pyg.boundary.plot(color='red')
shp.boundary.plot(ax=ax,)
```
Screenshot 2021-03-16 at 14 55 01

GEOS produces a larger polygon in all these cases.

With `naturalearth_lowres`, the plot looks like this:

Screenshot 2021-03-16 at 14 57 17

In this case, pygeos produces smaller polygon in 60 cases and shapely in 116 cases.

ljwolf:
What about the opencv2?

martinfleis:
OpenCV2 is almost the same as shapely. Still, mildly different areas but polygons essentially overlap.

pygeos/constructive.py
```diff
@@ -703,3 +704,33 @@ def voronoi_polygons(
     <pygeos.Geometry GEOMETRYCOLLECTION EMPTY>
     """
     return lib.voronoi_polygons(geometry, tolerance, extend_to, only_edges, **kwargs)
+
+
+@requires_geos("3.6.0")
+@multithreading_enabled
+def minimum_rotated_rectangle(geometry, **kwargs):
```
caspervdw, line 711: In PyGEOS we try to stick to the PostGIS namings of the functions. This one is named `ST_OrientedEnvelope` in postgis, so I'd like to propose `oriented_envelope` here.

pygeos/test/test_constructive.py
```diff
@@ -589,3 +589,14 @@ def test_segmentize_none():
 def test_segmentize(geometry, tolerance, expected):
     actual = pygeos.segmentize(geometry, tolerance)
     assert pygeos.equals(actual, geometry).all()
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
+@pytest.mark.parametrize("geometry", all_types)
+def test_minimum_rotated_rectangle(geometry):
```
caspervdw, line 596: Can you add a test containing the situations that are in the docstring? We don't doctest anymore in all environments.

pygeos/constructive.py
```diff
@@ -703,3 +704,33 @@ def voronoi_polygons(
...
+    Examples
+    --------
+    >>> minimum_rotated_rectangle(Geometry("MULTIPOINT (0 0, 10 0, 10 10)"))
+    <pygeos.Geometry POLYGON ((0 0, 5 -5, 15 5, 10 10, 0 0))>
+    >>> minimum_rotated_rectangle(Geometry("LINESTRING (1 1, 5 1, 10 10)"))
+    <pygeos.Geometry POLYGON ((1 1, 3 -1, 12 8, 10 10, 1 1))>
+    >>> minimum_rotated_rectangle(Geometry("POLYGON ((1 1, 15 1, 5 10, 1 1))"))
+    <pygeos.Geometry POLYGON ((15 1, 15 10, 1 10, 1 1, 15 1))>
+    >>> minimum_rotated_rectangle(Geometry("LINESTRING (1 1, 10 1)"))
+    <pygeos.Geometry LINESTRING (1 1, 10 1)>
```
caspervdw, line 730: Test failures on Py35 and Py36 will resolve after merging master (#325 )

caspervdw (changes requested):
Thanks @martinfleis  for this addition!

I left some remarks, most important is the naming of the function.

pygeos/constructive.py
martinfleis: I've changed it to `oriented_envelope`. Just out of curiosity, how will this play together with shapely's (different) names once it is merged together?

jorisvandenbossche: Yeah, I just wanted to comment here: Shapely already had this as `minimum_rotated_rectangle`, so we will need to keep that as an alias anyway, if we want to have the Postgis-compatible name as well

jorisvandenbossche: Also, another argument for `minmum_rotated_rectangle` name is the consistentcy with `minimum_bounding_circle`

martinfleis: I can switch back to `minimum_rotated_rectangle` as a default function, which is consistent with GEOS terminology (hence easier to follow through the code) and add `oriented_envelope` as an alias. Unless you want to do this once this will go to shapely.

caspervdw: > Also, another argument for minmum_rotated_rectangle name is the consistentcy with minimum_bounding_circle

And an argument for `oriented_envelope` is consistency with `envelope`. Neither solution is objectively better than the other. 

Naming discussions always take a ridiculous amount of time away from the development process. I intended to prevent this from happening in pygeos by blindly sticking to an already present corpus, namely, postgis. I don't see a reason for deviating from this now.

Because of the future merge with shapely, I think an alias is a good idea. But then `minimum_rotated_rectangle` can be the alias.

martinfleis: > Because of the future merge with shapely, I think an alias is a good idea. But then minimum_rotated_rectangle can be the alias.

Okay, I've done `minimum_rotated_rectangle` as an alias.

caspervdw:
With an “alias” I had something like `minimum_rotated_rectangle = oriented_envelope` in mind, not a complete copy. Not sure what is better. @jorisvandenbossche  do you have any preference? Maybe you encountered such a situation in `pandas`?

caspervdw:
If there are no strong opinions about how to do "aliases", I'll merge this. We can always do "aliases" different later.

Thanks @martinfleis for contributing and bearing with us for the requested changes!

PR #315, opened 2021-03-16
ENH: add minimum_bounding_circle and minimum_bounding_radius

martinfleis:
Exposing `GEOSMinimumBoundingCircle_r`. 

xref #126


martinfleis:
Hmm, this seems to work only on macOS and segfaults elsewhere. I am afraid that is out of my C capabilities to debug this.

jorisvandenbossche:
GEOSMinimumBoundingCircle_r takes an additional (`radius`) argument that is missing here? (not sure why it would sometimes succeed then, if that's the case)

jorisvandenbossche:
Ah, so it's not an additional "argument" to the function in a python-sense, but it actually is an additional return value (I wanted to say: have a look at the "Yd_Y" signatures, but so that won't work either)

jorisvandenbossche:
Using the following wrapper around the actual GEOS function to align the signature avoids the segfault for me on linux:

```
static void* GEOSMinimumBoundingCircleWithReturn(void* context, void* geom) {
  GEOSGeometry* center = NULL;
  double radius;
  void* ret = GEOSMinimumBoundingCircle_r(context, geom, &radius, &center);
  if (ret == NULL) {
    return NULL;
  }
  GEOSGeom_destroy_r(context, center);
  return ret;
}
static void* minimum_bounding_circle_data[1] = {GEOSMinimumBoundingCircleWithReturn};
```

Now, of course, that simply ignores the additional center/radius results. 

If we want to add those to the actual python return value, we need to think about the desired API for this (different return values? a tuple of arrays or an array of tuples?) I have done something similar for `polygonize_full` in https://github.com/pygeos/pygeos/pull/298

jorisvandenbossche:
PostGIS has a separate [`ST_MinimumBoundingRadius`](https://postgis.net/docs/ST_MinimumBoundingRadius.html), so [`ST_MinimumBoundingCircle`](https://postgis.net/docs/ST_MinimumBoundingCircle.html) only returns the geometry (which I think I like, since often you might want one or the other?)

martinfleis:
Thanks a lot for the wrapper!

> PostGIS has a separate ST_MinimumBoundingRadius, so ST_MinimumBoundingCircle only returns the geometry (which I think I like, since often you might want one or the other?)

I also like this approach. It feels natural to get a polygon from `minimum_bounding_circle`. We can do the same here and then mirror this behaviour for `maximum_inscribed_circle` (which I am also playing with now) which returns a LineString from the centre to a point on the boundary.

src/ufuncs.c
```diff
@@ -414,6 +414,17 @@ static void* polygons_without_holes_data[1] = {GEOSLinearRingToPolygon};
 static void* build_area_data[1] = {GEOSBuildArea_r};
 static void* make_valid_data[1] = {GEOSMakeValid_r};
 static void* coverage_union_data[1] = {GEOSCoverageUnion_r};
+static void* GEOSMinimumBoundingCircleWithReturn(void* context, void* geom) {
```
caspervdw, line 417: We could also have the `Y_d` function that returns the radius `minimum_bounding_radius` with this wrapper:

```
static int GEOSMinimumBoundingRadius(void* context, GEOSGeometry* geom, double* radius) {
  GEOSGeometry* center = NULL;
  GEOSGeometry* ret = GEOSMinimumBoundingCircle_r(context, geom, radius, &center);
  if (ret == NULL) {
    return 0;  // exception code
  }
  GEOSGeom_destroy_r(context, center);
  GEOSGeom_destroy_r(context, ret);
  return 1;  // success code
}
```

src/ufuncs.c
```diff
@@ -414,6 +414,17 @@ static void* polygons_without_holes_data[1] = {GEOSLinearRingToPolygon};
...
+  GEOSGeometry* center = NULL;
+  double radius;
+  void* ret = GEOSMinimumBoundingCircle_r(context, geom, &radius, &center);
```
caspervdw, line 420:
```suggestion
  GEOSGeometry* ret = GEOSMinimumBoundingCircle_r(context, geom, &radius, &center);
```

pygeos/constructive.py
```diff
@@ -703,3 +704,28 @@ def voronoi_polygons(
...
+    >>> minimum_bounding_circle(Geometry("POLYGON ((0 0, 0 10, 10 10, 10 0, 0 0))"))
+    <pygeos.Geometry POLYGON ((12.1 5, 11.9 3.62, 11.5 2.29, 10.9 1.07, 10 3.55e...>
+    >>> minimum_bounding_circle(Geometry("LINESTRING (1 1, 10 10)"))
+    <pygeos.Geometry POLYGON ((11.9 5.5, 11.7 4.26, 11.4 3.06, 10.8 1.96, 10 1, ...>
+    >>> minimum_bounding_circle(Geometry("MULTIPOINT (2 2, 4 2)"))
+    <pygeos.Geometry POLYGON ((4 2, 3.98 1.8, 3.92 1.62, 3.83 1.44, 3.71 1.29, 3...>
+    >>> minimum_bounding_circle(Geometry("POINT (0 1)"))
+    <pygeos.Geometry POINT (0 1)>
+    >>> minimum_bounding_circle(Geometry("GEOMETRYCOLLECTION EMPTY"))
+    <pygeos.Geometry POLYGON EMPTY>
```
caspervdw, line 729: Thanks; same comment as the other PR, can you include these as unittests as well?

caspervdw (changes requested):
Thanks @martinfleis , some small comments, maybe you can also add the `radius` version?

pygeos/measurement.py
```diff
@@ -277,3 +278,32 @@ def minimum_clearance(geometry, **kwargs):
...
+
+@requires_geos("3.8.0")
+@multithreading_enabled
+def minimum_bounding_radius(geometry, **kwargs):
+    """Computes the radius of tje minimum bounding circle that encloses an input geometry.
+
+    Parameters
+    ----------
+    geometry : Geometry or array_like
+
```
jorisvandenbossche, line 291: Same here for `**kwargs`

pygeos/measurement.py
```diff
@@ -277,3 +278,32 @@ def minimum_clearance(geometry, **kwargs):
     nan
     """
     return lib.minimum_clearance(geometry, **kwargs)
+
+
+@requires_geos("3.8.0")
+@multithreading_enabled
+def minimum_bounding_radius(geometry, **kwargs):
+    """Computes the radius of tje minimum bounding circle that encloses an input geometry.
```
jorisvandenbossche, line 286:
```suggestion
    """Computes the radius of the minimum bounding circle that encloses an input geometry.
```

src/ufuncs.c
```diff
@@ -399,6 +399,29 @@ static void* polygons_without_holes_data[1] = {GEOSLinearRingToPolygon};
...
+  double radius;
+  GEOSGeometry* ret = GEOSMinimumBoundingCircle_r(context, geom, &radius, &center);
+  if (ret == NULL) {
+    return NULL;
+  }
+  GEOSGeom_destroy_r(context, center);
+  return ret;
+}
+static void* minimum_bounding_circle_data[1] = {GEOSMinimumBoundingCircleWithReturn};
+static int GEOSMinimumBoundingRadius(void* context, GEOSGeometry* geom, double* radius) {
```
jorisvandenbossche, line 413: Can you move this to where the other Y_d functions are defined? (this is the Y_Y section)

caspervdw (approved):
Thanks! All looks good now for merge. Can you add a line to the CHANGELOG under "added GEOS functions" ?

martinfleis:
> Can you add a line to the CHANGELOG under "added GEOS functions" ?

@caspervdw All done. Thanks for the review to both of you.

PR #316, opened 2021-03-17
TST: Allow s390 CI fail

brendan-ward:
This allows the `s390x` arch to fail in Travis-CI, since it is failing due to errors at the GEOS level.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ dist: focal
 language: python
 python: "3.8"
 
+allow_failures:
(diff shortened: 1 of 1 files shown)


brendan-ward:
Unfortunately that didn't work as documented.  I'm going to try and enable Travis-CI testing for PRs temporarily to try and get this fixed (provided that we can set those in `.travis.yml`).  Otherwise will have to try to do this for named branches in the main repo.

PR #318, opened 2021-03-17
Enable "return_index" in get_coordinates

caspervdw:
This adds the option to return indices of coordinates in ``pygeos.get_coordinates``. 

In general I don't like the pattern of returning a different type depending on input parameters, but here it is such a simple solution to implement a large chunk of functionality, that I went for it anyway. `np.unique` does this too.

### DIFF
--- a/pygeos/coordinates.py
+++ b/pygeos/coordinates.py
@@ -103,7 +103,8 @@ def get_coordinates(geometry, include_z=False, return_index=False):
         geometry has no third dimension, the z-coordinates will be NaN.
     return_index : bool, default False
         Whether to return the index of each returned geometry as a separate
-        ndarray of integers.
+        ndarray of integers. For multidimensional arrays, this indexes into the
+        flattened array (in C contiguous order).
 
     Examples
     --------
@@ -120,8 +121,16 @@ def get_coordinates(geometry, include_z=False, return_index=False):
     [[0.0, 0.0]]
     >>> get_coordinates(Geometry("POINT Z (0 0 0)"), include_z=True).tolist()
     [[0.0, 0.0, 0.0]]
+
+    When return_index=True, indexes are returned also:
+    >>> geometries = [Geometry("LINESTRING (2 2, 4 4)"), Geometry("POINT (0 0)")]
+    >>> coordinates, index = get_coordinates(geometries, return_index=True)
+    >>> coordinates.tolist(), index.tolist()
+    ([[2.0, 2.0], [4.0, 4.0], [0.0, 0.0]], [0, 0, 1])
     """
-    return lib.get_coordinates(np.asarray(geometry, dtype=np.object_), include_z, return_index)
+    return lib.get_coordinates(
+        np.asarray(geometry, dtype=np.object_), include_z, return_index
+    )
 
 
 def set_coordinates(geometry, coordinates):
@@ -162,9 +171,8 @@ def set_coordinates(geometry, coordinates):
             "The coordinate array should have dimension of 2 "
             "(has {})".format(coordinates.ndim)
         )
-    if (
-        (coordinates.shape[0] != lib.count_coordinates(geometry_arr))
-        or (coordinates.shape[1] not in {2, 3})
+    if (coordinates.shape[0] != lib.count_coordinates(geometry_arr)) or (
+        coordinates.shape[1] not in {2, 3}
     ):
         raise ValueError(
             "The coordinate array has an invalid shape {}".format(coordinates.shape)
--- a/pygeos/coordinates.py
+++ b/pygeos/coordinates.py
@@ -20,8 +20,8 @@ def apply(geometry, transformation, include_z=False):
         A function that transforms a (N, 2) or (N, 3) ndarray of float64 to
         another (N, 2) or (N, 3) ndarray of float64.
     include_z : bool, default False
-        Whether to include the third dimension in the coordinates array
-        that is passed to the `transformation` function. If True, and a
+        If True, include the third dimension in the coordinates array
+        that is passed to the `transformation` function. If a
         geometry has no third dimension, the z-coordinates passed to the
         function will be NaN.
 
@@ -99,10 +99,10 @@ def get_coordinates(geometry, include_z=False, return_index=False):
     ----------
     geometry : Geometry or array_like
     include_z : bool, default False
-        Whether to include the third dimension in the output. If True, and a
-        geometry has no third dimension, the z-coordinates will be NaN.
+        If, True include the third dimension in the output. If a geometry
+        has no third dimension, the z-coordinates will be NaN.
     return_index : bool, default False
-        Whether to return the index of each returned geometry as a separate
+        If True, also return the index of each returned geometry as a separate
         ndarray of integers. For multidimensional arrays, this indexes into the
         flattened array (in C contiguous order).
 
--- a/pygeos/coordinates.py
+++ b/pygeos/coordinates.py
@@ -44,7 +44,7 @@ def apply(geometry, transformation, include_z=False):
     <pygeos.Geometry POINT Z (1 1 1)>
     """
     geometry_arr = np.array(geometry, dtype=np.object_)  # makes a copy
-    coordinates = lib.get_coordinates(geometry_arr, include_z)
+    coordinates = lib.get_coordinates(geometry_arr, include_z, False)
     new_coordinates = transformation(coordinates)
     # check the array to yield understandable error messages
     if not isinstance(new_coordinates, np.ndarray):
@@ -87,7 +87,7 @@ def count_coordinates(geometry):
     return lib.count_coordinates(np.asarray(geometry, dtype=np.object_))
 
 
-def get_coordinates(geometry, include_z=False):
+def get_coordinates(geometry, include_z=False, return_index=False):
     """Gets coordinates from a geometry array as an array of floats.
 
     The shape of the returned array is (N, 2), with N being the number of
@@ -101,6 +101,9 @@ def get_coordinates(geometry, include_z=False):
     include_z : bool, default False
         Whether to include the third dimension in the output. If True, and a
         geometry has no third dimension, the z-coordinates will be NaN.
+    return_index : bool, default False
+        Whether to return the index of each returned geometry as a separate
+        ndarray of integers.
 
     Examples
     --------
@@ -118,7 +121,7 @@ def get_coordinates(geometry, include_z=False):
     >>> get_coordinates(Geometry("POINT Z (0 0 0)"), include_z=True).tolist()
     [[0.0, 0.0, 0.0]]
     """
-    return lib.get_coordinates(np.asarray(geometry, dtype=np.object_), include_z)
+    return lib.get_coordinates(np.asarray(geometry, dtype=np.object_), include_z, return_index)
 
 
 def set_coordinates(geometry, coordinates):
--- a/src/coords.c
+++ b/src/coords.c
@@ -364,7 +364,7 @@ PyObject* GetCoords(PyArrayObject* arr, int include_z, int return_index) {
   NpyIter* iter;
   NpyIter_IterNextFunc* iternext;
   char** dataptr;
-  npy_intp cursor;
+  npy_intp cursor, i, geom_i;
   GeometryObject* obj;
   GEOSGeometry* geom;
   PyArrayObject* index = NULL;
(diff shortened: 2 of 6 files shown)


pygeos/coordinates.py
```diff
@@ -101,6 +101,10 @@ def get_coordinates(geometry, include_z=False):
     include_z : bool, default False
         Whether to include the third dimension in the output. If True, and a
         geometry has no third dimension, the z-coordinates will be NaN.
+    return_index : bool, default False
+        Whether to return the index of each returned geometry as a separate
```
brendan-ward, line 105:
```suggestion
        If True will return the index of each returned geometry as a separate
```

src/coords.c
```diff
@@ -384,24 +384,42 @@ PyObject* GetCoords(PyArrayObject* arr, int include_z) {
   if (result == NULL) {
     return NULL;
   }
+  if (return_index) {
+    npy_intp dims_ind[1] = {size};
+    index = (PyArrayObject*)PyArray_SimpleNew(1, dims_ind, NPY_INTP);
+    if (index == NULL) {
+      Py_DECREF(result);
```
brendan-ward, line 391: It seems like this should raise an exception similar to `CHECK_ALLOC`?

src/coords.c
```diff
@@ -384,24 +384,42 @@ PyObject* GetCoords(PyArrayObject* arr, int include_z) {
...
 
   /* We use the Numpy iterator C-API here.
   The iterator exposes an "iternext" function which updates a "dataptr"
   see also: https://docs.scipy.org/doc/numpy/reference/c-api.iterator.html */
-  iter = NpyIter_New(arr, NPY_ITER_READONLY | NPY_ITER_REFS_OK, NPY_KEEPORDER,
+  iter = NpyIter_New(arr, NPY_ITER_READONLY | NPY_ITER_REFS_OK, NPY_CORDER,
                      NPY_NO_CASTING, NULL);
   if (iter == NULL) {
     Py_DECREF(result);
+    Py_DECREF(index);
```
brendan-ward, line 415: It seems like this and following block (line 423) should raise exceptions?

src/coords.c
```diff
@@ -411,9 +429,11 @@ PyObject* GetCoords(PyArrayObject* arr, int include_z) {
   /* We work with a "cursor" that tells the get_coordinates function where
   to write the coordinate data into the output array "result" */
   cursor = 0;
+  geom_i = -1;
```
brendan-ward, line 432: This seems a bit awkward; what about initializing to 0, and then moving the update of it from line 436 to the end of the `do` loop (after line 455, before line 456)

src/coords.c
```diff
@@ -423,10 +443,16 @@ PyObject* GetCoords(PyArrayObject* arr, int include_z) {
       continue;
     }
     /* get the coordinates */
+    i = cursor;  // initialize the index to use in the return_index loop later
```
brendan-ward, line 446: This took me a little bit to figure out: this is initialized here because `cursor` is then incremented by `get_coordinates` below.  Perhaps expanding on the comment a little bit would make that more clear?

Otherwise it is confusing when you then look at `i <= cursor` in the loop below.

brendan-ward (commented):
Thanks for working on this, it will be a helpful addition and having it as a parameter seems reasonable to me.  A few comments to consider.

One thing to also consider is how this expands to nested coordinate geometry types (Polygon, GeometryCollection, Multi*).

The user can easily deal with GeometryCollection / Multi* by using `get_parts` first and keeping track of the outer indexes (is this worth a mention in the docstring?).  

However, for Polygons with inner rings, it is more awkward.  Often in my usage of something like this, I need to keep the coordinate arrays of each ring separate.  For example, to calculate the area of the outer ring (using geodesic methods) and subtract the area of the inner rings.

One approach (future PR, but good to consider how that impacts the API in this PR) would be to add an additional parameter `return_ring_index` that if `True` would add another integer array to the output tuple.  Would need to have a little validation: `return_index` must also be `True` and behavior for non-polygon types in a mixed collection is undefined (0 seems incorrect, -1 could produce bad indexing behavior later, etc).

src/coords.c
caspervdw: I did that first but then I got in trouble with the `continue` on line 443. So the alternative is doing `geom_1++` there also.

caspervdw:
> One thing to also consider is how this expands to nested coordinate geometry types (Polygon, GeometryCollection, Multi*).
> 
> The user can easily deal with GeometryCollection / Multi* by using `get_parts` first and keeping track of the outer indexes (is this worth a mention in the docstring?).
> 
> However, for Polygons with inner rings, it is more awkward. Often in my usage of something like this, I need to keep the coordinate arrays of each ring separate. For example, to calculate the area of the outer ring (using geodesic methods) and subtract the area of the inner rings.
> 
> One approach (future PR, but good to consider how that impacts the API in this PR) would be to add an additional parameter `return_ring_index` that if `True` would add another integer array to the output tuple. Would need to have a little validation: `return_index` must also be `True` and behavior for non-polygon types in a mixed collection is undefined (0 seems incorrect, -1 could produce bad indexing behavior later, etc).

I was thinking along these lines as well. Returning the index of each coordinate inside a geometrycollection would be great also from a serialization/deserialization perspective. Polygons could be treated here as a "collection of linearrings".

I do wonder if this is strictly necessary to implement this in C. I think with clever usage of `pygeos.get_num_coordinates` /  `pygeos.get_num_geometries`, `pygeos.get_num_interior_rings` one could construct the arrays without any python inner loop.

caspervdw:
And I am having a hard time to reproduce the test failures locally. I set up a conda env exactly as it is done in the tests, but still, tests pass... any ideas?

brendan-ward:
I created a brand new conda env that matches the MacOS / Python 3.8 / GEOS and numpy latest env on CI.  I was also able to reproduce in my default environment (MacOS, GEOS `master` only after manually removing `build`, `.cache`, `*.egg-info`.

I am able to reproduce a segfault only by running entire test suite; running test modules individually does not produce the exception.  Might be helpful to check if there is a memory leak; testing that now...

brendan-ward:
There is a memory leak (via valgrid Docker container):
```

Valgrind has detected an error:

**1** 
**1** **********************************************************************
**1** pygeos/test/test_coordinates.py::test_get_coords_index[geoms2-index2]
**1** **********************************************************************
==1== Invalid write of size 8
==1==    at 0xE21EB33: GetCoords (coords.c:453)
==1==    by 0xE21EE58: PyGetCoords (coords.c:590)
==1==    by 0x4F6CB3C: _PyCFunction_FastCallDict (methodobject.c:234)
==1==    by 0x4FB1529: call_function (ceval.c:4851)
==1==    by 0x4FAA92E: _PyEval_EvalFrameDefault (ceval.c:3335)
==1==    by 0x4FB1B37: _PyEval_EvalCodeWithName (ceval.c:4166)
==1==    by 0x4FB1B37: fast_function (ceval.c:4992)
==1==    by 0x4FB165A: call_function (ceval.c:4872)
==1==    by 0x4FAB435: _PyEval_EvalFrameDefault (ceval.c:3351)
==1==    by 0x4FA9BBA: _PyEval_EvalCodeWithName (ceval.c:4166)
==1==    by 0x4FA9BBA: PyEval_EvalCodeEx (ceval.c:4187)
==1==    by 0x4F4EEC2: function_call (funcobject.c:604)
==1==    by 0x4F3A43A: PyObject_Call (abstract.c:2261)
==1==    by 0x4FAC655: do_call_core (ceval.c:5120)
==1==    by 0x4FAC655: _PyEval_EvalFrameDefault (ceval.c:3404)
==1==  Address 0x7f18008 is 0 bytes after a block of size 8 alloc'd
==1==    at 0x4C2BBAF: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1==    by 0xEFCF41D: npy_alloc_cache (in /usr/local/lib/python3.6/site-packages/numpy/core/_multiarray_umath.cpython-36m-x86_64-linux-gnu.so)
==1==    by 0xF01FC2A: PyArray_NewFromDescr_int (in /usr/local/lib/python3.6/site-packages/numpy/core/_multiarray_umath.cpython-36m-x86_64-linux-gnu.so)
==1==    by 0xF02023F: PyArray_New (in /usr/local/lib/python3.6/site-packages/numpy/core/_multiarray_umath.cpython-36m-x86_64-linux-gnu.so)
==1==    by 0xE21E98A: GetCoords (coords.c:389)
==1==    by 0xE21EE58: PyGetCoords (coords.c:590)
==1==    by 0x4F6CB3C: _PyCFunction_FastCallDict (methodobject.c:234)
==1==    by 0x4FB1529: call_function (ceval.c:4851)
==1==    by 0x4FAA92E: _PyEval_EvalFrameDefault (ceval.c:3335)
==1==    by 0x4FB1B37: _PyEval_EvalCodeWithName (ceval.c:4166)
==1==    by 0x4FB1B37: fast_function (ceval.c:4992)
==1==    by 0x4FB165A: call_function (ceval.c:4872)
==1==    by 0x4FAB435: _PyEval_EvalFrameDefault (ceval.c:3351)
==1== 
==1== LEAK SUMMARY:
==1==    definitely lost: 2,720 (+0) bytes in 30 (+0) blocks
==1==    indirectly lost: 881 (+0) bytes in 13 (+0) blocks
==1==      possibly lost: 6,930,139 (+1,040) bytes in 67,167 (+2) blocks
==1==    still reachable: 16,335,430 (+1,229) bytes in 114,012 (+25) blocks
==1==                       of which reachable via heuristic:
==1==                         newarray           : 722 (+0) bytes in 39 (+0) blocks
==1==                         multipleinheritance: 2,712 (+0) bytes in 47 (+0) blocks
==1==         suppressed: 0 (+0) bytes in 0 (+0) blocks
==1== Reachable blocks (those to which a pointer was found) are not shown.
==1== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==1== 
```

jorisvandenbossche:
Just as reference for the "need ring indices as well" discussion: that's basically what I implemented in https://github.com/pygeos/pygeos/pull/93 
It's true that you can achieve something similar with combination of `get_num_geometries`/`get_num_interior_rings`/`get_num_coordinates` (see the python code snippet in the top post of https://github.com/pygeos/pygeos/issues/75, which basically does this. And the "flatten" parts in that snippet can now be replaced with the new `get_parts`). 
Before doing this in C, we should probably update that Python snippet to have a  baseline to compare with (it will still be faster in C as you need less "passes" over the data, but difficult to guess how much faster).

caspervdw:
Thanks @brendan-ward  for investigating, I'll look further into this

src/coords.c
caspervdw: In my experience, the python C API will have set an exception in case it returns NULL. I don't really no a way of testing it here though.

src/coords.c
brendan-ward: It looks like this line is leading to the segfault, for reasons I don't understand.  If I hardcode it to 0, no segfault.

src/coords.c
```diff
@@ -360,14 +359,15 @@ npy_intp CountCoords(PyArrayObject* arr) {
...
 }
 
-PyObject* GetCoords(PyArrayObject* arr, int include_z) {
+PyObject* GetCoords(PyArrayObject* arr, int include_z, int return_index) {
   npy_intp coord_dim;
   NpyIter* iter;
   NpyIter_IterNextFunc* iternext;
   char** dataptr;
-  npy_intp cursor;
+  npy_intp cursor, i, geom_i;
```
brendan-ward, line 367: To be safe, probably should initialize all to 0, especially `cursor`.

brendan-ward: Oops, I see that is done a little further down.

caspervdw:
@brendan-ward  I solved the segfault, `GetCoords` tried to write outside the bounds of the `index` array.

pygeos/coordinates.py
caspervdw: What is wrong with the "whether to"? we used it also in the `include_z`

brendan-ward: Either is fine, I just thought "If True" was a bit more explicit.  Could be updated for `include_z` too.  😄 

Elsewhere we seem to generally use "If True..." or a somewhat more verbose variant ("If <parameter> is set to True")

caspervdw: Ok, I normalized these docstrings to "If True, ..."

brendan-ward (approved):
Thanks for the changes and for normalizing boolean parameter docstrings!

PR #321, opened 2021-03-18
Raise warning and return None on invalid WKB/WKT instead of raising exception

### LINKED ISSUE #320: GEOS errors decoding WKB prevent reading from datasets with invalid geometries
Various tools are using WKB as a serialization format or transport format between different binary packages (e.g., reading from GDAL as WKB, then converting to geometries here).  The same is probably true for WKT.

If any of the WKB geometries are malformed, GEOS throws an error and the entire conversion fails.  This makes it challenging to work with 3rd-party datasets that have invalid geometries, since these first need to be converted to geometries before they can be filtered.

We can potentially get around this in one of 2 ways:
1. add a parameter to `from_wkb` that would allow invalid geometries to be returned as `None`, and raise a warning
2. always return invalid geometries as `None` and raise a warning.  This is simpler and the only API change is that it doesn't fail outright (which is likely undesirable here).

The user can then filter out the `None` geometries and continue with the remainder.

To be clear: for other areas where the user can actually control how geometries are constructed, it is totally appropriate to raise an exception.  It is only in cases of deserializing WKB / WKT that it would be better to gracefully handle such errors.

brendan-ward:
Resolves #320

For reasons stated in #320, it is problematic to raise exceptions when deserializing invalid geometries from formats such as WKB.  This provides a way to gracefully read those in, raise a warning, and allow the user to filter them out instead of failing hard.

This adds a new enum value `PGWARN_GEOS_EXCEPTION` which is used to raise the GEOS exception as a warning instead of an exception.

### DIFF
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -181,31 +181,36 @@ def from_wkt(geometry, **kwargs):
     return lib.from_wkt(geometry, **kwargs)
 
 
-def from_wkb(geometry, **kwargs):
+def from_wkb(geometry, ignore_invalid=False, **kwargs):
     r"""
     Creates geometries from the Well-Known Binary (WKB) representation.
 
     The Well-Known Binary format is defined in the `OGC Simple Features
     Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
 
-    Invalid geometries will be returned as None and will raise a warning.
-    These can be filtered out using pygeos.is_geometry.
 
     Parameters
     ----------
     geometry : str or array_like
         The WKB byte object(s) to convert.
+    ignore_invalid : bool (default: False)
+        If True, each invalid WKB geometry will be returned as None and a
+        warning will be raised.
 
     Examples
     --------
     >>> from_wkb(b'\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\xf0?')
     <pygeos.Geometry POINT (1 1)>
     """
+
+    if not np.isscalar(ignore_invalid):
+        raise TypeError("ignore_invalid only accepts scalar values")
+
     # ensure the input has object dtype, to avoid numpy inferring it as a
     # fixed-length string dtype (which removes trailing null bytes upon access
     # of array elements)
     geometry = np.asarray(geometry, dtype=object)
-    return lib.from_wkb(geometry, **kwargs)
+    return lib.from_wkb(geometry, ignore_invalid, **kwargs)
 
 
 def from_shapely(geometry, **kwargs):
--- a/pygeos/io.py
+++ b/pygeos/io.py
@@ -188,6 +188,9 @@ def from_wkb(geometry, **kwargs):
     The Well-Known Binary format is defined in the `OGC Simple Features
     Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
 
+    Invalid geometries will be returned as None and will raise a warning.
+    These can be filtered out using pygeos.is_geometry.
+
     Parameters
     ----------
     geometry : str or array_like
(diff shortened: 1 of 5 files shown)


src/geos.h
```diff
@@ -85,6 +86,9 @@ enum {
     case PGERR_EMPTY_GEOMETRY:                                                           \
       PyErr_SetString(PyExc_ValueError, "One of the Geometry inputs is empty.");         \
       break;                                                                             \
+    case PGWARN_GEOS_EXCEPTION:                                                          \
+      PyErr_WarnEx(PyExc_Warning, last_error, 0);                                        \
```
jorisvandenbossche, line 90: We should probably prepend that last_error message to make the warning message clearer? (saying something about some values being invalid WKB and being replaced with None in the result)

brendan-ward:
A couple things to consider:
* I'm not sure that `ignore_invalid` is the best name for this parameter.  Is there something better?
* WKT and WKB warning types are separate enum values now but this doesn't add much value.  These could be combined into a single enum value if we don't think these need to be kept separate.

caspervdw (changes requested):
This looks good. Some thoughts about the keyword `ignore_invalid`:

I think it should be a string (like numpy does in `np.seterr` or pandas in `to_datetime`). In that way we can later on add more functionality without breaking the API. This can be done now easily by only allowing `raise` and `ignore`, and converting that to a bool to pass into the `lib.from_wkt`. Then our external API remains extensible. Then it is reasonable to rename it to:

- `errors` (like in pandas.to_datetime https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html
- `on_invalid`

The separation between WKT and WKB warnings is not necessary, but I am also not against it. We can always change it back.

jorisvandenbossche:
+1 on making it a string option

> This can be done now easily by only allowing `raise` and `ignore`

We could maybe also further use the terminology of `np.seterr` and use "raise" and "warn", which keeps the possibility open to add a "ignore" (which would not show the warning)? 

About naming, I think I find `on_invalid` a bit more descriptive than `errors`.

brendan-ward:
Latest changes make this a string parameter and add an `ignore` option.

pygeos/io.py
```diff
@@ -172,37 +181,59 @@ def from_wkt(geometry, **kwargs):
     ----------
     geometry : str or array_like
         The WKT string(s) to convert.
+    on_invalid : str, one of {"raise", "warn", "ignore"} (default: "raise")
```
caspervdw, line 184: From https://numpydoc.readthedocs.io/en/latest/format.html:

> When a parameter can only assume one of a fixed set of values, those values can be listed in braces, with the default appearing first:
> order : {'C', 'F', 'A'}
    Description of `order`.

```suggestion
    on_invalid : {"raise", "warn", "ignore"}
```

The default is also listed in the call signature, which is visible in the docs.

Also the markup of the list below is not ok, see https://pygeos--321.org.readthedocs.build/en/321/io.html#pygeos.io.from_wkb. You might want each line with a hyphens (not sure)

caspervdw (approved):
Looks good, one nitpick about the docstrings and then good to merge!

pygeos/io.py
```diff
@@ -172,37 +181,59 @@ def from_wkt(geometry, **kwargs):
...
+        raise: an exception will be raised if WKT input geometries are invalid.
+        warn: a warning will be raised and invalid WKT geometries will be
+        returned as `None`.
+        ignore: invalid WKT geometries will be returned as `None` without a warning.
```
jorisvandenbossche, line 188:
```suggestion
        - raise: an exception will be raised if WKT input geometries are invalid.
        - warn: a warning will be raised and invalid WKT geometries will be
          returned as `None`.
        - ignore: invalid WKT geometries will be returned as `None` without a warning.
```

Issue #138, opened 2020-04-27
Create geometries with different number of points in a vectorized way


TonyXiang8787:
I have found this open-source project by looking for a vectorized geospatial python library. I am impressed about how the library can handle geospatial processing with full vectorization. However, I could not figure out how to create geometries with different number of points in a vectorized way. A very simple example.

I can create two polygons in one call, if the polygons have the same number of points.

```python
pygeos.polygons(
    [
        [[0.0, 0.0], [1.0, 0.0], [1.0, 1.0]],
        [[0.0, 0.0], [1.0, 0.0], [1.0, 1.0]]
    ]
)
```

However, if I put two polygons with different number of points, it will not work.

```python
pygeos.polygons(
    [
        [[0.0, 0.0], [1.0, 0.0], [1.0, 1.0]],
        [[0.0, 0.0], [1.0, 0.0], [1.0, 1.0], [0.0, 1.0]]
    ]
)
```

I will get a  type error:
`TypeError: One of the arguments is of incorrect type. Please provide only Geometry objects.`

Is this a design choice? Or does it have to do with the limitations of `ufunc` in `numpy`? Or maybe I have not found the correct usage of this library. In that case I apologise for not being able to find the relevant information in the documentation.

caspervdw:
Hi @TonyXiang8787, thanks for your question. The fact that your example gives a TypeError is caused by that the functions expects an ndarray. What you supply can't be supported as it is not 'rectangular'. So it fails. This is not by design, this is just because you can't do it differently if you want to use numpy directly. 

The issue you are raising is in fact the reverse of https://github.com/pygeos/pygeos/issues/128 : how can you deal with vectorizing over arrays of  geometries  with different numbers of elements? We just reached a conclusion for the API of extracting the points, I guess we could also implement the reverse of that. 

Keep in mind that vectorizing over coordinate *input* only makes sense if coordinates are already in a numpy ndarray. In your example, the input coordinates are in a (nested) list so you will always need a python for loop to loop over individual coordinate lists. You'll never get the speedup because you need to extract the data from a 'slow' datastructure. Given your example, I'd say you can't be faster than:

```
np.array([pygeos.polygons(x) for x in
    [
        [[0.0, 0.0], [1.0, 0.0], [1.0, 1.0]],
        [[0.0, 0.0], [1.0, 0.0], [1.0, 1.0], [0.0, 1.0]]
    ]
])
```

brendan-ward:
[...]

Issue #253, opened 2020-11-15
API change: equals(None, None) should be True


brendan-ward:
Current behavior is that `equals(None, None)` is `False`.  It is listed this way in the docstring.  But this doesn't make sense (unless I'm misunderstanding), it seems like for this special case, it should always be `True'.

jorisvandenbossche:
I am personally not sure `True` is necessarily the better / more expected answer. For example, NaN is also not equal to itself. So at least there is also reason to have it return False, but there are probably motivations to give for both.

I think it probably depends on your usecase of `equals` whether you would prefer/expect True vs False. 
For example, if you want to check if two arrays are fully (spatially) equal with `pygeos.equals(arr1, arr2).all()`, then you typically want to consider missing values at the same location in the arrays as equal as well (eg the `np.allclose` function has a `equal_nan` keyword, with default False. The `Seris.equals()` method in pandas (but which already returns a boolean scalar!) will also consider missing values as equal. 
But, if you are using `equals` for example in a boolean indexing operation (like `arr[arr == arr2]`, but then using `arr[pygeos.equals(arr, arr2)]`), then I think you would expect missing values not to be equal to each other (at least that is the behaviour in numpy/pandas for boolean indexing).

brendan-ward:
I was thinking about it in the context of testing, it came up in #254.  For instance, it seems counterintuitive in this context that `None` is `False`:

```
>>> pygeos.equals([None,pygeos.box(0,0,1,1)],[None, pygeos.box(0,0,1,1)])
array([False,  True])
```

because for object-level equality
```
>>> None == None
True
```

Whereas empty geometries remain equal:
```
>>> pygeos.equals(pygeos.Geometry('POINT EMPTY'),pygeos.Geometry('POINT EMPTY'))
True
```

It seems reasonable and follows numpy for `nan` to be non-equal, but diverges from numpy for `None`:
```
>>> np.array_equal([np.nan], [np.nan])
False
>>> np.array_equal([None], [None])
True
```

(allowing user to define that `nan` are equal based on a parameter might be good here too).

jorisvandenbossche:
> I was thinking about it in the context of testing, 

Yep, so more in line with my first example usecase. 
In general, I think we should actually have a specialized `assert_equal`-like method that works on arrays, and deals with this issue (in the same way as numpy has a way to test if two arrays are equal with considering NaN as equal, and pandas has methods to assert series/dataframe equality ignoring certain aspects, etc)

[...]

PR #322, opened 2021-03-22
Create simple geometries from coordinates and indices

caspervdw:
This solves https://github.com/pygeos/pygeos/issues/138 , API was already described in #241  and implemented for collections in #290

Polygons are still a todo.


pygeos/_geos.pxd
```diff
@@ -8,30 +8,40 @@ Example:
...
+    # Types
     ctypedef void *GEOSContextHandle_t
+    ctypedef struct GEOSGeometry
+    ctypedef struct GEOSCoordSequence
+    ctypedef void (*GEOSMessageHandler_r)(const char *message, void *userdata)
 
+    # GEOS Context & Messaging
     GEOSContextHandle_t GEOS_init_r() nogil
     void GEOS_finish_r(GEOSContextHandle_t handle) nogil
+    void GEOSContext_setErrorMessageHandler_r(GEOSContextHandle_t handle, GEOSMessageHandler_r ef, void* userData) nogil
```
caspervdw, line 20: See here on why I removed the `except` on each line:

https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#checking-return-values-of-non-cython-functions

brendan-ward:
Thanks for working on this!

I'm a bit torn on the handling of empty geometries / `None` / exceptions here.  Ideally, it would also be good for this to be consistent with #290, and either always return `None` when there are no indices for that output record or always return empty of that particular geometry type.  I do think that `None` seems more reasonable in this case, so I agree with the first part of the proposal (and apply that back to the geometry collections too).  An exception doesn't seem appropriate in this case.

For invalid geometries, I think in this case, raising an exception seems better than a warning; this is more consistent with the non-indexed geometry constructors.  I think one key question to ask here is the degree to which the exceptions are preventable or recoverable: upon encountering the exception, can the user go back and easily modify the inputs to prevent the exception?  And do we want to force the user to inspect those inputs more carefully?  I think yes: invalid inputs should not be allowed to pass through the constructors in this case.

Returning `None` certainly makes it easy to filter these out afterward, but I think consistency may be important here: if I cannot construct the singular case with those coordinates, I shouldn't be able to do so within a larger array.

I don't think that geometries should revert to a lower dimension if insufficient coordinates are supplied (i.e., linearring -> point).

pygeos/_geometry.pyx
```diff
@@ -27,6 +35,102 @@ from pygeos._pygeos_api cimport (
...
+                idx += 1
+
+            if geometry_type == 0:
+                geom = GEOSGeom_createPoint_r(geos_handle, seq)
+            elif geometry_type == 1:
+                geom = GEOSGeom_createLineString_r(geos_handle, seq)
+            elif geometry_type == 2:
+                if ring_closure == 1:
+                    if GEOSCoordSeq_setX_r(geos_handle, seq, geom_size, coord_view[idx - geom_size, 0]) == 0:
+                        return
```
brendan-ward, line 118: Does the coordinate sequence need to be cleaned up if there are errors on line 118, 120, or 123?  Once `GEOSGeom_createLinearRing_r` is called, it takes ownership and would get cleaned up properly.

brendan-ward (commented):
I didn't have a chance to fully review everything but wanted to get some feedback in sooner.

Overall I like where this is going; see prior comments around `None` and exceptions.

Thanks for catching the erroneous `except` statements on the wrapping of C functions; that was a misinterpretation of the Cython docs on my part.

caspervdw:
Thanks for your review @brendan-ward . I agree with your considerations and now the simple constructors return `None` if there are no coordinates provided. To test this, I implemented `assert_geometries_equal` (ref https://github.com/pygeos/pygeos/issues/253)

caspervdw:
@brendan-ward   I'd like to merge this to be able to continue with the polygons (and get the whole indices thing done for a release). Do you want to take a last look?

pygeos/_geometry.pyx
```diff
@@ -69,9 +183,13 @@ def get_parts(object[:] array):
             for part_idx in range(counts_view[geom_idx]):
                 index_view[idx] = geom_idx
                 part = GEOSGetGeometryN_r(geos_handle, geom, part_idx)
+                if part == NULL:
```
brendan-ward, line 186: This `return` doesn't seem right; the expected return shape is a tuple of ndarrays.

This should raise an exception instead: we expected a part and we didn't get one back.  The GEOS docs state that a `NULL` return here is an exception.

pygeos/_geometry.pyx
```diff
@@ -81,6 +199,14 @@ def get_parts(object[:] array):
     return parts, index
 
 
+cdef _deallocate_arr(void* handle, np.intp_t[:] arr, Py_ssize_t last_geom_i):
+    """Deallocate a temporary geometry array to prevent memory leaks"""
+    cdef Py_ssize_t i = 0
+
+    for i in range(last_geom_i):
+        GEOSGeom_destroy_r(handle, <GEOSGeometry *>arr[i])
```
brendan-ward, line 207: I know this was just centralized and not otherwise new, but given recent other changes to handle freeing things we shouldn't so as to avoid segfaults, I wonder if we should first check that `arr[i]` is != `NULL` before calling `GEOSGeom_destroy_r` on it - just to be sure?

pygeos/_geometry.pyx
```diff
@@ -162,16 +286,21 @@ def collections_1d(object geometries, object indices, int geometry_type = 7):
...
                             f"One of the arguments has unexpected geometry type {curr_type}."
                         )
 
-                # assign to the temporary geometry array                   
-                temp_geoms_view[coll_size] = <np.intp_t>GEOSGeom_clone_r(geos_handle, geom)
+                # assign to the temporary geometry array  
+                geom = GEOSGeom_clone_r(geos_handle, geom)
+                if geom == NULL:
+                    _deallocate_arr(geos_handle, temp_geoms_view, coll_size)
+                    return
```
brendan-ward, line 302: Shouldn't this be an exception?  GEOS returns `NULL` on exception here.

pygeos/_geometry.pyx
```diff
@@ -181,6 +310,8 @@ def collections_1d(object geometries, object indices, int geometry_type = 7):
                 <GEOSGeometry**> &temp_geoms_view[0],
                 coll_size
             )
+            if coll == NULL:
+                return
```
brendan-ward, line 314: Shouldn't this be an exception?  GEOS returns `NULL` on exception here.

pygeos/test/test_creation_indices.py
```diff
@@ -1,36 +1,120 @@
...
+    "indices", [np.array([point]), " hello", [0, 1]]  # wrong length
+)
+def test_invalid_indices_simple(func, indices):
+    with pytest.raises((TypeError, ValueError)):
+        func([[0.2, 0.3]], indices=indices)
+
+
+def test_points_invalid():
+    with pytest.raises(pygeos.GEOSException):
+        pygeos.points([[1, 1], [2, 2]], indices=[0, 0])
```
brendan-ward, line 59: It's not immediately obvious why this fails, perhaps adding either a comment or adding a `matches='Point coordinate list must contain a single element'` to `pytest.raises` would make that a little bit more obvious?

brendan-ward (changes requested):
@caspervdw apologies for the delay on review.  There are a few returns here that I think should be exceptions instead, but otherwise the implementation looks good.

Feel free to merge once the exceptions are in place.

pygeos/_geometry.pyx
caspervdw: If you do `return NULL` from C, you tell python to raise an exception. It will raise an exception reading something like "function returned NULL without an exception message set".

However, we do set an exception message. We basically trust GEOS here to have called the `setErrorMessageHandler` which result in a `GEOSException` in the `get_geos_handle().__exit__`. It seems a bit convolved, but it is the only way I know to be able to catch the GEOS exception message. I'll make a comment on the `return NULL`.

brendan-ward (approved):
Thanks for the changes!  The bare return was definitely not obvious; the comments help.

PR #325, opened 2021-03-28
Only run doctests on conda-forge, latest Py38

### LINKED ISSUE #317: TST: run doctest only in a single environment (with a single GEOS version) ?
The doctest regularly give problems in failing doctests across different CI builds. This can generally be solved by either adding a `normalize` call, or by skipping (`# doctest: +SKIP`) some parts, or some other workaround.

I would propose to run the doctests only in a single build (eg with the latest released packages on conda-forge), making it easier to get them passing. Arguments:

- The doctests' purpose is to ensure the documentation is correct, not as an actual test of the functionality (for that we have the other tests). So for this reason, it's not necessarily needed to run them on all CI builds IMO
- The workarounds to deal with varying behaviour across versions IMO only adds clutter to the docstring, while the main purpose of the docstring is to give a clear, nice example.

caspervdw:
Closes #317

### DIFF
--- a/.github/workflows/test-conda.yml
+++ b/.github/workflows/test-conda.yml
@@ -16,6 +16,7 @@ jobs:
           - os: ubuntu-latest
             python: 3.8
             channel: conda-forge
+            pytest_args: "--doctest-modules"
           - os: windows-latest
             python: 3.8
             channel: conda-forge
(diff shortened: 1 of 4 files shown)


caspervdw:
I going to self-merge is, I believe we already agreed on this in the issue.

jorisvandenbossche:
Yes, thanks!

PR #327, opened 2021-04-01
[Done] Add kwargs where necessary and document it

caspervdw:
This adds `**kwargs` forwarding to ufuncs where missing.

Also I wanted to document this on each function. I did one (`boundary`), what do you think? 

https://pygeos--327.org.readthedocs.build/en/327/constructive.html#pygeos.constructive.boundary

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -51,7 +51,7 @@ def boundary(geometry, **kwargs):
     geometry : Geometry or array_like
         This function will return None for geometrycollections.
     **kwargs : dict
-        For other keyword-only arguments, see the
+        For the keyword-only arguments, see the
         `Numpy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
 
     Examples
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -52,7 +52,7 @@ def boundary(geometry, **kwargs):
         This function will return None for geometrycollections.
     **kwargs
         For other keyword-only arguments, see the
-        `Numpy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+        `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
 
     Examples
     --------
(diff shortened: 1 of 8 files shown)


brendan-ward (commented):
Thanks for working on this.  Adding them everywhere appropriate will be nice for consistency.  

Adding them to the docstring is also helpful; I wasn't familiar with these until you pointed them out in a prior PR.

pygeos/constructive.py
```diff
@@ -50,6 +50,9 @@ def boundary(geometry, **kwargs):
     ----------
     geometry : Geometry or array_like
         This function will return None for geometrycollections.
+    **kwargs : dict
```
jorisvandenbossche, line 53: I would remove the "dict", since you typically don't pass it as a dict

pygeos/constructive.py
```diff
@@ -50,6 +50,9 @@ def boundary(geometry, **kwargs):
...
+        For the keyword-only arguments, see the
```
jorisvandenbossche, line 54:
```suggestion
        For the additional keyword-only arguments, see the
```

jorisvandenbossche (commented):
Indeed a good idea to add and document this consistently. Added 2 comments on the docstring

pygeos/constructive.py
caspervdw: I made it "For other keyword-only", exactly how numpy does it.

pygeos/constructive.py
```diff
@@ -50,6 +50,9 @@ def boundary(geometry, **kwargs):
     ----------
     geometry : Geometry or array_like
         This function will return None for geometrycollections.
+    **kwargs
+        For other keyword-only arguments, see the
+        `Numpy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
```
jorisvandenbossche, line 55:
```suggestion
        `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
```

(very small nitpick, but with a search/replace using the full line it might be easy to change)

I think in general it's either all lowercase (the package), or if referrring the project they capitalize the P as well

caspervdw:
@jorisvandenbossche  OK to merge this to stay clear of the merge conflicts? Or do you want to do #298  first?

jorisvandenbossche:
It shouldn't give merge conflicts with #298, but yet, let's merge this

PR #328, opened 2021-04-01
Add flake8 linter and format repo

caspervdw:
I used the tools `black` and `isort` to format the repo. The `isort` required some settings to match `black`.

`flake8` captured one thing that will lead to a performance improvement in `set_coordinates` and `apply`.

### DIFF
--- a/pygeos/_geometry.pyx
+++ b/pygeos/_geometry.pyx
@@ -1,36 +1,38 @@
 # distutils: define_macros=GEOS_USE_ONLY_R_API
 
-from cpython cimport PyObject
 cimport cython
+from cpython cimport PyObject
 from cython cimport view
 
 import numpy as np
+
 cimport numpy as np
+
 import pygeos
 
 from pygeos._geos cimport (
-    GEOSGeometry,
-    GEOSCoordSequence,
     GEOSContextHandle_t,
-    GEOSGeom_clone_r,
-    GEOSGetGeometryN_r,
-    get_geos_handle,
-    GEOSGeom_destroy_r,
-    GEOSGeom_createCollection_r,
-    GEOSGeomTypeId_r,
     GEOSCoordSeq_create_r,
     GEOSCoordSeq_destroy_r,
     GEOSCoordSeq_setX_r,
     GEOSCoordSeq_setY_r,
     GEOSCoordSeq_setZ_r,
-    GEOSGeom_createPoint_r,
+    GEOSCoordSequence,
+    GEOSGeom_clone_r,
+    GEOSGeom_createCollection_r,
+    GEOSGeom_createLinearRing_r,
     GEOSGeom_createLineString_r,
-    GEOSGeom_createLinearRing_r
+    GEOSGeom_createPoint_r,
+    GEOSGeom_destroy_r,
+    GEOSGeometry,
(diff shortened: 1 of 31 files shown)


setup.cfg
```diff
@@ -4,3 +4,19 @@ style = pep440
 versionfile_source = pygeos/_version.py
 versionfile_build = pygeos/_version.py
 tag_prefix =
+
+[flake8]
+ignore = E203, E266, E501, W503, "pygeos/_version.py"
+
+[isort]
+atomic = true
+force_alphabetical_sort = true
```
brendan-ward, line 13: I think you should add: `sections=FUTURE,STDLIB,FIRSTPARTY,THIRDPARTY,LOCALFOLDER`
([source](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering/))

alphebetical sort within those categories would be good though.

caspervdw: Thanks, I just copied these settings from another project. In the meantime, isort added `profile=black` which fixes most of the things. Also I changed the alphabetic sorting to alphabetic sorting *per section* (as you suggested)

jorisvandenbossche:
Looks good!

I think we all three already have been using black (to some extent) locally, and you now reformatted some remaining code with black. But shall we then directly enforce black as well in the linting step? (as you do for isort now)

jorisvandenbossche:
Also: have you been using pre-commit?  

Personally, if things like isort are enforced on CI, having a pre-commit hook to check this for you locally is really handy, I have found with other projects. 

(And if we have a pre-commit config, that could also be used for the CI linting step)

caspervdw:
I added black as well.

> Also: have you been using pre-commit?

Yes I have been using it on several projects, just didn't set it up here. Good idea.

jorisvandenbossche (approved):

jorisvandenbossche:
Is it OK that I merge this? Then I can update https://github.com/pygeos/pygeos/pull/298, as otherwise merging that PR will probably require to rerun things here.

pygeos/_geometry.pyx
```diff
@@ -1,36 +1,38 @@
 # distutils: define_macros=GEOS_USE_ONLY_R_API
 
-from cpython cimport PyObject
 cimport cython
+from cpython cimport PyObject
```
brendan-ward, line 4: @caspervdw I'm getting different ordering on `master` when I run `pre-commit run --files pygeos/_geometry.pyx`

```
from cpython cimport PyObject

cimport cython
from cython cimport view
```

So it fails the linting check...
(using latest version of `isort`, 5.8.0 but also happened on 4.3 prior to upgrading it)

Not seeing what would cause this to be different locally; CI seems to be passing OK.  Any ideas other than wiping and reinstalling env?

jorisvandenbossche: If you are using pre-commit, it should already use a separate isolated env (so wiping the env shouldn't help normally)

For me that is passing:

```
(geo-dev) joris@joris-XPS-13-9380:~/scipy/repos/pygeos$ pre-commit run --files pygeos/_geometry.pyx
black................................................(no files to check)Skipped
flake8...............................................(no files to check)Skipped
isort....................................................................Passed
```

PR #330, opened 2021-04-02
Add pre-commit configuration

jorisvandenbossche:
Follow-up on https://github.com/pygeos/pygeos/pull/328

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,15 +13,5 @@ jobs:
         with:
           python-version: 3.8
 
-      - name: Install flake8, isort and black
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@caspervdw this adds a pre-commit configuration to use locally, so you can ensure your commits will pass our CI.

I *also* updated the lint.yml to use this instead of manually installing/running black/flake8/isort. This part is not necessary for this PR, but I think one big advantage of doing so is to ensure your local linting check (with pre-commit) exactly matches what happens on CI.

caspervdw:
Looks good, thanks!

PR #326, opened 2021-03-29
[Done] Create polygons from indices

caspervdw:
This build onto #322

PR #331, opened 2021-04-02
Release the GIL in shared_paths (missing decorator)

### LINKED ISSUE #155: ENH: release the GIL in spatial operations (intersection, union, etc) that return geometries

caspervdw:
This closes #155 (along with #326 which releases the GIL for geometry-creating functions)

### DIFF
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -113,6 +113,7 @@ def line_merge(line, **kwargs):
     return lib.line_merge(line, **kwargs)
 
 
+@multithreading_enabled
 def shared_paths(a, b, **kwargs):
     """Returns the shared paths between geom1 and geom2.
 


jorisvandenbossche (approved):
Thanks!

PR #334, opened 2021-04-05
ENH: Add shortest_line (nearest_points) ufunc

jorisvandenbossche:
This is initial work to expose the `GEOSNearestPoints_r` function. There are however some design questions to decide on.

- The GEOS function returns a lenght-2 `CoordSequence`, so we need to wrap/unwrap that in some way for the actual return value
- Shapely exposes it as [`shapely.ops.nearest_points`](https://shapely.readthedocs.io/en/stable/manual.html#nearest-points) which returns a tuple of 2 Point geometries.
- The R spatial packages (`sf`) have a [`st_nearest_points`](https://r-spatial.github.io/sf/reference/st_nearest_points.html) to expose this, which returns this as a two-point LineString. 
- PostGIS doesn't expose this GEOS function, but they do have a [`ST_ShortestLine`](https://postgis.net/docs/ST_ShortestLine.html) with similar functionality, also returning a line of two points. 

In general, I think returning the two-point LineString geometries for the ufunc would make sense. You can then still easily get the first or second point using `get_points` (to mimic the current shapely return value), but having a single return value makes it easier to implement and use, I think.

If we decide to return LineStrings (and not arrays of points), I think the name "shortest_line" actually makes more sense as "nearest_points". It would be consistent in naming with PostGIS, but since they have their own custom implementation (it's not exposing `GEOSNearestPoints`), I am not sure it behaves exactly the same as "nearest_points", though. On the other hand "shortest_line" would be inconsistent with R `sf`. 
If we we go with "shortest_line" for the ufunc, we can still easily have a "nearest_points" in python,as a small wrapper around the ufunc, which directly returns the tuple/array of points, to be consistent with Shapely.

---

Given those open discussion points, I didn't yet add the python wrapper or tests, but so the actual ufunc already works:

```
In [12]: pygeos.lib.nearest_points(pygeos.box(0, 0, 1, 1), pygeos.box(2, 2, 3, 3))
Out[12]: <pygeos.Geometry LINESTRING (1 1, 2 2)>
```

### DIFF
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -4,7 +4,13 @@ from . import Geometry  # NOQA
 from . import lib
 from .decorators import multithreading_enabled
 
-__all__ = ["line_interpolate_point", "line_locate_point", "line_merge", "shared_paths"]
+__all__ = [
+    "line_interpolate_point",
+    "line_locate_point",
+    "line_merge",
+    "shared_paths",
+    "shortest_line",
+]
 
 
 @multithreading_enabled
@@ -140,3 +146,25 @@ def shared_paths(a, b, **kwargs):
     <pygeos.Geometry GEOMETRYCOLLECTION (MULTILINESTRING EMPTY, MULTILINESTRING ...>
     """
     return lib.shared_paths(a, b, **kwargs)
+
+
+def shortest_line(a, b, **kwargs):
+    """
+    Returns the shortest line between geom1 and geom2.
+
+    Parameters
+    ----------
+    a : Geometry or array_like
+    b : Geometry or array_like
+    **kwargs
+        For other keyword-only arguments, see the
+        `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+
+    Examples
+    --------
+    >>> geom1 = Geometry("LINESTRING (0 0, 1 0, 1 1, 0 1, 0 0)")
+    >>> geom2 = Geometry("LINESTRING (0 3, 3 0, 5 3)")
+    >>> shortest_line(geom1, geom2)
+    <pygeos.Geometry LINESTRING (1 1, 1.5 1.5)>
+    """
+    return lib.nearest_points(a, b, **kwargs)
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -148,15 +148,16 @@ def shared_paths(a, b, **kwargs):
     return lib.shared_paths(a, b, **kwargs)
 
 
+@multithreading_enabled
 def shortest_line(a, b, **kwargs):
     """
-    Returns the shortest line between geom1 and geom2.
+    Returns the shortest line between two geometries.
 
     The resulting line consists of two points, representing the nearest
-    points between the geometry pair. The line always starts in geom1 and
-    ends in geom2. The endpoints of the line will not necessarily be existing
-    vertices of geom1 and geom2, but can be any point along a line segment
-    (if the input geometries are lines or polygons).
+    points between the geometry pair. The line always starts in the first
+    geometry `a` and ends in he second geometry `b`. The endpoints of the
+    line will not necessarily be existing vertices of the input geometries
+    `a` and `b`, but can also be a point along a line segment.
 
     Parameters
     ----------
@@ -165,6 +166,10 @@ def shortest_line(a, b, **kwargs):
     **kwargs
         For other keyword-only arguments, see the
         `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+
+    See also
+    --------
+    prepare : improve performance by preparing ``a`` (the first argument) (for GEOS>=3.9)
 
     Examples
     --------
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -152,6 +152,12 @@ def shortest_line(a, b, **kwargs):
     """
     Returns the shortest line between geom1 and geom2.
 
+    The resulting line consists of two points, representing the nearest
+    points between the geometry pair. The line always starts in geom1 and
+    ends in geom2. The endpoints of the line will not necessarily be existing
+    vertices of geom1 and geom2, but can be any point along a line segment
+    (if the input geometries are lines or polygons).
+
     Parameters
     ----------
     a : Geometry or array_like
--- a/pygeos/linear.py
+++ b/pygeos/linear.py
@@ -178,4 +178,4 @@ def shortest_line(a, b, **kwargs):
     >>> shortest_line(geom1, geom2)
     <pygeos.Geometry LINESTRING (1 1, 1.5 1.5)>
     """
-    return lib.nearest_points(a, b, **kwargs)
+    return lib.shortest_line(a, b, **kwargs)
--- a/pygeos/test/test_linear.py
+++ b/pygeos/test/test_linear.py
@@ -165,9 +165,9 @@ def test_shortest_line():
 
 
 def test_shortest_line_none():
-    # assert pygeos.shortest_line(line_string, None) is None
+    assert pygeos.shortest_line(line_string, None) is None
     assert pygeos.shortest_line(None, line_string) is None
-    # assert pygeos.shortest_line(None, None) is None
+    assert pygeos.shortest_line(None, None) is None
 
 
 def test_shortest_line_empty():
--- a/pygeos/test/test_linear.py
+++ b/pygeos/test/test_linear.py
@@ -156,21 +156,36 @@ def test_shared_paths_non_linestring():
         pygeos.shared_paths(g1, g2)
 
 
-def test_shortest_line():
+def _prepare_input(geometry, prepare):
+    """Prepare without modifying inplace"""
+    if prepare:
+        geometry = pygeos.apply(geometry, lambda x: x)  # makes a copy
+        pygeos.prepare(geometry)
+        return geometry
+    else:
+        return geometry
+
+
+@pytest.mark.parametrize("prepare", [True, False])
+def test_shortest_line(prepare):
     g1 = pygeos.linestrings([(0, 0), (1, 0), (1, 1)])
     g2 = pygeos.linestrings([(0, 3), (3, 0)])
-    actual = pygeos.shortest_line(g1, g2)
+    actual = pygeos.shortest_line(_prepare_input(g1, prepare), g2)
     expected = pygeos.linestrings([(1, 1), (1.5, 1.5)])
     assert pygeos.equals(actual, expected)
 
 
-def test_shortest_line_none():
-    assert pygeos.shortest_line(line_string, None) is None
+@pytest.mark.parametrize("prepare", [True, False])
+def test_shortest_line_none(prepare):
+    assert pygeos.shortest_line(_prepare_input(line_string, prepare), None) is None
     assert pygeos.shortest_line(None, line_string) is None
     assert pygeos.shortest_line(None, None) is None
 
 
-def test_shortest_line_empty():
-    assert pygeos.shortest_line(line_string, empty_line_string) is None
-    assert pygeos.shortest_line(empty_line_string, line_string) is None
-    assert pygeos.shortest_line(empty_line_string, empty_line_string) is None
+@pytest.mark.parametrize("prepare", [True, False])
+def test_shortest_line_empty(prepare):
+    g1 = _prepare_input(line_string, prepare)
+    assert pygeos.shortest_line(g1, empty_line_string) is None
+    g1_empty = _prepare_input(empty_line_string, prepare)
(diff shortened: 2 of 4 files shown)


src/ufuncs.c
```diff
@@ -1942,6 +1942,70 @@ static void polygonize_full_func(char** args, npy_intp* dimensions, npy_intp* st
...
+      break;
+    }
+
+    if ((in1 == NULL) | (in2 == NULL)) {
+      // in case of a missing value: return NULL (None)
+      geom_arr[i] = NULL;
+    } else if (GEOSisEmpty_r(ctx, in1) | GEOSisEmpty_r(ctx, in2)) {
+      // GEOSNearestPoints_r would otherwise return a NULL not distinguishable
+      // from an actual error
+      geom_arr[i] = NULL;
```
jorisvandenbossche, line 1980: An implementation aspect: if we decide to return a LineString (as I did here), this is actually a `YY_Y` type of function, and I could write a small wrapper around `GEOSNearestPoints_r` that returns the LineString, so we can use it with the `YY_Y_func` template. 

But, one difference with the other YY_Y functions (intersection, difference, union, etc) is that `GEOSNearestPoints_r` also returns NULL if any of the inputs are empty geometries (it checks for that explicitly in the C API wrapper: https://github.com/libgeos/geos/blob/cd0ade60ab050cf70c66f8c555701717e4486461/capi/geos_ts_c.cpp#L865-L866), while it in general uses a NULL return value to indicate an error (as the other functions used in YY_Y). 

Personally, I think it would be nice to not error for empty input, but handle it like we already handle None input. In addition, the GEOS exception is also completely unclear that it comes from empty geometries (although this is something that could be fixed in a future GEOS version)

brendan-ward:
`GEOSPreparedNearestPoints_r` (GEOS >= 3.9) could be good to add here as well, for geometries that are already prepared (or in a later PR).

Is it possible to craft the ufunc to add an inner dimension of size 2 and return these as points?

A LineString return also seems OK and in keeping with the precedent set elsewhere. `shortest_line` seems most appropriate for a function that returns lines, otherwise it is confusing if it has points in the name but returns a line.

Wrapping to return tuples as an additional optional return type via another function (`nearest_points`) exposed on the python side seems reasonable, but my hunch is it would be good to do this in addition to exposing the ufunc above as well.

jorisvandenbossche:
> Is it possible to craft the ufunc to add an inner dimension of size 2 and return these as points?

That should be possible yes. We already have the `bounds` ufunc which is a generalized ufunc with signature `"()->(n)"` (only with a float return type, but I suppose this will work for object dtype as well). 

But so the question is: what do we prefer as the main API? (two-point linestrings vs points with added dimension with length 2)

If we would also want a "shortest_line" version, then it might be easier to implement that as the ufunc, and layer the point returning version on top of it in Python. But it could also be the other way around. 
(I don't think it is needed to implement both in C as ufunc)

> `GEOSPreparedNearestPoints_r` (GEOS >= 3.9) could be good to add here as well, for geometries that are already prepared (or in a later PR).

Yes, that's a good idea

jorisvandenbossche:
@caspervdw are you OK with implementing the ufunc as "shortest_line" (so returning a 2-point linestring)?

src/ufuncs.c
```diff
@@ -1942,6 +1942,79 @@ static void polygonize_full_func(char** args, npy_intp* dimensions, npy_intp* st
...
+    if (!get_geom(*(GeometryObject**)ip2, &in2)) {
+      errstate = PGERR_NOT_A_GEOMETRY;
+      destroy_geom_arr(ctx, geom_arr, i - 1);
+      break;
+    }
+
+    if ((in1 == NULL) | (in2 == NULL)) {
+      // in case of a missing value: return NULL (None)
+      geom_arr[i] = NULL;
+    } else if (GEOSisEmpty_r(ctx, in1) | GEOSisEmpty_r(ctx, in2)) {
```
brendan-ward, line 1978: This could be simplified to something like:
```
if ((in1 == NULL) | (in2 == NULL) | GEOSisEmpty_r(ctx, in1) | GEOSisEmpty_r(ctx, in2)) {
    // in case of a missing value or empty geometry: return NULL (None)
    // GEOSNearestPoints_r returns NULL for empty geometries
    // but this is not distinguishable from an actual error, so we handle this ourselves
    geom_arr[i] = NULL;
    continue;
}
#if GEOS_SINCE_3_9_0
```
(note: drops the corresponding `else` since the first conditional has a `continue`)

brendan-ward (commented):
Agreed that only one of `shortest_line` or `nearest_points` needs to be a ufunc, and adding the other side (if needed) in Python.  Another option would be to return a 2-point `MultiPoint` to keep it as a singular geometry object rather than tuple of geometry objects.  I don't have a personal preference for which of these would make the best API; I haven't needed this functionality previously, and all of them are easy to transform into other representations for a particular analysis.

Minor suggestion to flatten simplify the conditional logic a little bit.

jorisvandenbossche:
> Another option would be to return a 2-point `MultiPoint` to keep it as a singular geometry object rather than tuple of geometry objects.

The GEOS function returns a ``GEOSCoordSequence``, and it's easier / cheaper to create a LineString from this than a MultiPoint, I think.

src/ufuncs.c
```diff
@@ -1945,6 +1945,78 @@ static void polygonize_full_func(char** args, npy_intp* dimensions, npy_intp* st
...
+      destroy_geom_arr(ctx, geom_arr, i - 1);
+      break;
+    }
+    if (!get_geom(*(GeometryObject**)ip2, &in2)) {
+      errstate = PGERR_NOT_A_GEOMETRY;
+      destroy_geom_arr(ctx, geom_arr, i - 1);
+      break;
+    }
+
+    if ((in1 == NULL) | (in2 == NULL) | GEOSisEmpty_r(ctx, in1) |
```
jorisvandenbossche, line 1978: Does somebody see something obviously wrong here?  The tests with `None` geometry input are segfaulting, and it's crashing inside `GEOSisEmpty`, so for some reason it's getting passed the ``(in1 == NULL) | (in2 == NULL)`` check.

jorisvandenbossche: @brendan-ward if I write this again as two separate steps (checking for NULL and checking for empty), then it passes. So C does not short-circuit? 

-> Ah, I was using the bitwise `|` and not the logical `||` ... (and not only here, we do it in other places as well!). Properly using the logical one does short-circuit.

jorisvandenbossche: Made a general PR to fix this -> https://github.com/pygeos/pygeos/pull/348

pygeos/linear.py
```diff
@@ -140,3 +146,25 @@ def shared_paths(a, b, **kwargs):
     <pygeos.Geometry GEOMETRYCOLLECTION (MULTILINESTRING EMPTY, MULTILINESTRING ...>
     """
     return lib.shared_paths(a, b, **kwargs)
+
+
+def shortest_line(a, b, **kwargs):
+    """
+    Returns the shortest line between geom1 and geom2.
+
```
brendan-ward, line 154: It would likely be helpful to clarify if the endpoints of the lines created here (output of `GEOSNearestPoints_r`) are interpolated on the nearest lines between the two geometries, if the inputs are lines or polygons.  At least, that's my interpretation of the dostring for [JTS](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/operation/distance/DistanceOp.html)
> The distance computation also finds a pair of points in the input geometries which have the minimum distance between them. If a point lies in the interior of a line segment, the coordinate computed is a close approximation to the exact point.

As opposed to finding the nearest _existing_ vertices on each geometry that are the closest of other existing vertices between the geometries.

jorisvandenbossche: I am not fully sure how to interpret "the coordinate computed is a close approximation to the exact point". 

Would be it be sufficient to indicate that the resulting endpoints aren't necessarily existing vertices, but can be any point along the line segment? (that seems a simpler version as saying that the points "are interpolated on the nearest lines")

jorisvandenbossche: > to meant that there exists a theoretical exact nearest point within the nearest line segment facing the other geometry, and that the method should get very close but not necessarily exactly equal to that theoretical point.

What I am wondering is whether this "not necessarily exactly equal" is just due to inherent floating point precision limitations, or if there is something more specific to this algorithm (as the floating point precision of coordinates is something that can come up in many GEOS operations)

pygeos/linear.py
```diff
@@ -140,3 +146,31 @@ def shared_paths(a, b, **kwargs):
     <pygeos.Geometry GEOMETRYCOLLECTION (MULTILINESTRING EMPTY, MULTILINESTRING ...>
     """
     return lib.shared_paths(a, b, **kwargs)
+
+
+def shortest_line(a, b, **kwargs):
+    """
+    Returns the shortest line between geom1 and geom2.
```
caspervdw, line 153:
```suggestion
    Returns the shortest line between two geometries.
```

pygeos/linear.py
```diff
@@ -140,3 +146,31 @@ def shared_paths(a, b, **kwargs):
...
+
+    The resulting line consists of two points, representing the nearest
+    points between the geometry pair. The line always starts in geom1 and
```
caspervdw, line 156:
```suggestion
    points between the geometry pair. The line always starts in `a` and
```

pygeos/linear.py
```diff
@@ -140,3 +146,31 @@ def shared_paths(a, b, **kwargs):
...
+    ends in geom2. The endpoints of the line will not necessarily be existing
```
caspervdw, line 157:
```suggestion
    ends in `b`. The endpoints of the line will not necessarily be existing
```

pygeos/linear.py
```diff
@@ -140,3 +146,31 @@ def shared_paths(a, b, **kwargs):
...
+    vertices of geom1 and geom2, but can be any point along a line segment
```
caspervdw, line 158:
```suggestion
    vertices of geom1 and geom2, but can also be a point along a line segment.
```

The latest remark "(if the input geometries are lines or polygons)." is a bit confusing, I think it doesn't add anything and should be removed

src/ufuncs.c
```diff
@@ -1945,6 +1945,78 @@ static void polygonize_full_func(char** args, npy_intp* dimensions, npy_intp* st
 }
 static PyUFuncGenericFunction polygonize_full_funcs[1] = {&polygonize_full_func};
 
+static char nearest_points_dtypes[3] = {NPY_OBJECT, NPY_OBJECT, NPY_OBJECT};
```
caspervdw, line 1948: The "nearest_points" function here does not return points, but a line. So I think it better be named `shortest_line` in the ufuncs.c (just like in the python function).

pygeos/linear.py
```diff
@@ -140,3 +146,31 @@ def shared_paths(a, b, **kwargs):
...
+
+    Parameters
+    ----------
+    a : Geometry or array_like
+    b : Geometry or array_like
+    **kwargs
+        For other keyword-only arguments, see the
+        `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
+
+    Examples
```
caspervdw, line 169: Can you add a "See also" section before the Examples that refers to `prepare`? The predicate functions have that, I think it is good to mention that preparing geometries will help here (under the condition that GEOS>=3.9).

pygeos/test/test_linear.py
```diff
@@ -154,3 +154,23 @@ def test_shared_paths_non_linestring():
     g2 = pygeos.points(0, 1)
     with pytest.raises(pygeos.GEOSException):
         pygeos.shared_paths(g1, g2)
+
+
+def test_shortest_line():
```
caspervdw, line 159: Can you also test with a prepared geometry?

pygeos/linear.py
```diff
@@ -140,3 +146,31 @@ def shared_paths(a, b, **kwargs):
     <pygeos.Geometry GEOMETRYCOLLECTION (MULTILINESTRING EMPTY, MULTILINESTRING ...>
     """
     return lib.shared_paths(a, b, **kwargs)
+
+
+def shortest_line(a, b, **kwargs):
```
caspervdw, line 151: This releases the GIL so it requires the `@multithreading_enabled`

caspervdw (changes requested):
Looks good, thanks!

Some minor remarks, this should be good to merge after you address those. And please also append the changelog.

src/ufuncs.c
jorisvandenbossche: Indee, I didn't yet rename the actual ufunc after renaming the public function

pygeos/test/test_linear.py
jorisvandenbossche: Ah, good point, I forgot to test with prepared geometries

caspervdw (approved):

jorisvandenbossche:
Thanks for the reviews!

PR #341, opened 2021-04-13
Show docstrings for functions decorated with requires_geos

### LINKED ISSUE #339: Functions decorated with @requires_geos have no proper documentation
See: https://pygeos.readthedocs.io/en/latest/constructive.html?highlight=segmentize#pygeos.constructive.segmentize

I guess this only happens when the GEOS version in the docbuilder is too low. We should be able to propagate the documentation in the decorator.

caspervdw:
Closes #339

### DIFF
--- a/.github/workflows/test-conda.yml
+++ b/.github/workflows/test-conda.yml
@@ -75,4 +75,4 @@ jobs:
         shell: bash
         run: |
           source activate test
-          pytest ${{ matrix.pytest_args }}
+          pytest pygeos ${{ matrix.pytest_args }}
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -697,8 +697,8 @@ def segmentize(geometry, tolerance, **kwargs):
     Examples
     --------
     >>> line = Geometry("LINESTRING (0 0, 0 10)")
-    >>> segmentize(line, tolerance=5)
-    <pygeos.Geometry LINESTRING (0 0, 0 5, 0 10)>  # doctest: +SKIP
+    >>> segmentize(line, tolerance=5)  # doctest: +SKIP
+    <pygeos.Geometry LINESTRING (0 0, 0 5, 0 10)>
     >>> poly = Geometry("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))")
     >>> segmentize(poly, tolerance=5)  # doctest: +SKIP
     <pygeos.Geometry POLYGON ((0 0, 5 0, 10 0, 10 5, 10 10, 5 10, 0 10, 0 5, 0 0))>
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -698,11 +698,11 @@ def segmentize(geometry, tolerance, **kwargs):
     --------
     >>> line = Geometry("LINESTRING (0 0, 0 10)")
     >>> segmentize(line, tolerance=5)
-    <pygeos.Geometry LINESTRING (0 0, 0 5, 0 10)>
+    <pygeos.Geometry LINESTRING (0 0, 0 5, 0 10)>  # doctest: +SKIP
     >>> poly = Geometry("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))")
-    >>> segmentize(poly, tolerance=5)
+    >>> segmentize(poly, tolerance=5)  # doctest: +SKIP
     <pygeos.Geometry POLYGON ((0 0, 5 0, 10 0, 10 5, 10 10, 5 10, 0 10, 0 5, 0 0))>
-    >>> segmentize(None, tolerance=5) is None
+    >>> segmentize(None, tolerance=5) is None  # doctest: +SKIP
     True
     """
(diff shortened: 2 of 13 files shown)


pygeos/test/test_misc.py
```diff
@@ -71,6 +71,32 @@ def foo():
         foo()
 
 
+@pytest.mark.parametrize("version", ["3.6.0", "3.8.0"])
+def test_requires_geos_adapts_docstring(version, mocked_geos_version):
+    """The requires_geos decorator adapts the docstring."""
+
+    @requires_geos(version)
+    def foo():
```
jorisvandenbossche, line 79: You could also put this function just before the test function so it's not nested, and then you don't have that strange indentation issue to comment about.

jorisvandenbossche (approved):
Nice! 

If we want to solve that it also works for non-top-level functions, I think you would need to add some `textwrap.dedent` use (we do that in pandas, eg https://github.com/pandas-dev/pandas/blob/7d4757b4deb851bb44ab6bb20cdc404fa13fffcf/pandas/util/_decorators.py#L82). 
But since we don't yet have such functions, that's not really important for now of course ;) (just noticed it based on your comment in the test)

caspervdw:
See for example:

https://pygeos--341.org.readthedocs.build/en/341/constructive.html#pygeos.constructive.segmentize

pygeos/decorators.py
```diff
@@ -16,19 +16,23 @@ def __init__(self, version):
...
 
-            wrapped.__doc__ = msg
-            return wrapped
         else:
-            return func
+            wrapped = func
+
+        # Add a note about the version constraint after the first line of
+        # the docstring.
+        if wrapped.__doc__:
```
brendan-ward, line 31: It seems like this would be more useful where `lib.geos_version < self.version`; if the current install of pygeos / GEOS already satisfies this requirement, this is not necessary info unless needed for the docs on readthedocs.

brendan-ward (commented):
Thanks for working on this @caspervdw.  I think my main question is around how helpful it is to know the GEOS version requirement when your install already satisfies it.  However, it _is_ useful when you don't have that version, or for the docs out on readthedocs.

caspervdw:
I added some changes. Now the note is added *only* if there is a docbuild. A docbuild is identified via an env variable set in `docs/conf.py`. 

If it is not a docbuild, the current behaviour is maintained (so: if the version does not match, then the docstrings contains the message, else, it remains unchanged).

caspervdw:
Removed Python 3.5 compat as decided in #125.

pygeos/decorators.py
```diff
@@ -16,20 +17,33 @@ def __init__(self, version):
...
+        elif lib.geos_version < self.version:
 
             @wraps(func)
             def wrapped(*args, **kwargs):
                 raise UnsupportedGEOSOperation(msg)
 
             wrapped.__doc__ = msg
-            return wrapped
-        else:
+        else:  # return directly, do not change the docstring
```
brendan-ward, line 42: Suggestion: return `wrapped` before the else, then you can dropped the `else` and simply `return func`.

.github/workflows/test-conda.yml
```diff
@@ -75,4 +75,4 @@ jobs:
         shell: bash
         run: |
           source activate test
-          pytest ${{ matrix.pytest_args }}
+          pytest pygeos ${{ matrix.pytest_args }}
```
brendan-ward, line 78: Note: you can also use the `setup.cfg` file for this: https://docs.pytest.org/en/stable/customize.html#setup-cfg

brendan-ward (approved):
A few minor comments but otherwise this looks good.

pygeos/decorators.py
```diff
@@ -16,19 +17,33 @@ def __init__(self, version):
...
+        if os.environ.get("SPHINX_DOC_BUILD") == "1":  # set in docs/conf.py
+            # Doc build: just add a note about the version constraint after
+            # the first line of the docstring.
+            @wraps(func)
+            def wrapped(*args, **kwargs):
+                return func(*args, **kwargs)
+
+            if wrapped.__doc__:
+                wrapped.__doc__ = wrapped.__doc__.replace(
+                    "\n\n", "\n\n    .. note:: {}\n\n".format(msg), 1
```
jorisvandenbossche, line 32: I would personally do this for the case of `lib.geos_version < self.version` as well. Also if your geos version is too old, it can be useful to be able to read the docstring

jorisvandenbossche (approved):

pygeos/constructive.py
```diff
@@ -702,11 +698,11 @@ def segmentize(geometry, tolerance, **kwargs):
     --------
     >>> line = Geometry("LINESTRING (0 0, 0 10)")
     >>> segmentize(line, tolerance=5)
-    <pygeos.Geometry LINESTRING (0 0, 0 5, 0 10)>
+    <pygeos.Geometry LINESTRING (0 0, 0 5, 0 10)>  # doctest: +SKIP
```
jorisvandenbossche, line 701: Reason for red CI; the skip needs to be on the line above (you did it correctly below)

PR #342, opened 2021-04-14
Explode polygons into rings: get_rings

jorisvandenbossche:
Related to the comment from @brendan-ward at https://github.com/pygeos/pygeos/pull/318#pullrequestreview-615323018 about the indices returned by `get_coordinates`, which is fine for Multi-part geometries (through `get_parts`), but harder if you also want to know the indices of rings:

> One approach (future PR, but good to consider how that impacts the API in this PR) would be to add an additional parameter `return_ring_index` that if `True` would add another integer array to the output tuple.  Would need to have a little validation: `return_index` must also be `True` and behavior for non-polygon types in a mixed collection is undefined (0 seems incorrect, -1 could produce bad indexing behavior later, etc).

An alternative to returning additional ring indices from `get_coordinates`, would be to have something similar as `get_parts` but then for the rings. This PR does a quick draft of such a function. 

The main downside is that if you only need to indices, exploding into rings creates a lot of intermediate objects (and memory usage) just for that. Although the same can be said about `get_parts` for this use case.


caspervdw:
This looks good! Is it conceivable to actually integrate the polygons into `get_parts`? Or is it too different in your opinion?

For just returning the indexes, I think we can try to also add a `collection_index` to `get_coordinates`. With that, we will have a complete representation of the geometries (except for the types and the mixed geometrycollections). That will be very useful for serialization.

jorisvandenbossche:
> Is it conceivable to actually integrate the polygons into get_parts? Or is it too different in your opinion?

Do you mean that `get_parts` would also automatically explode a polygon into its rings, like a MultiPolygon gets exploded in Polygons?

If we add it to `get_parts`, I think that should certainly be behind a flag (to enable with a keyword), because it is typically not want you want. For example when you have a mixture of Polygons/MultiPolygons, you want to get all as Polygons, and `get_parts` handles that perfectly. 
Also, an important difference with rings is that those no longer are a simple composition. A MultiPolygon is semantically equal to its set of Polygons, but a Polygon is not just a combination of its rings, i.e. the different rings have a different interpretation (exterior/shell vs holes). I think that's an important distinction compared to what `get_parts` does for all types of geometries, and therefore probably worth to be a separate function?

jorisvandenbossche:
If we would like such a `get_rings` as a separate function, I can certainly clean-up / finalize this PR.

caspervdw:
I thought it might give us less code to maintain. Semantically there should indeed be a difference, I agree.

An option is to add `explode_polygons=True/False` to `get_parts`, but if it is easier implementation-wise to stick to the `get_rings`, please go for it.

jorisvandenbossche:
Even with different public functions, the implementation in cython could still be shared. 

Now, given that for rings, you need to get exterior ring and interior rings separately (it's not a single inner loop with ``GEOSGetGeometryN_r``), it won't necessarily result in more readable code by putting that inside `get_parts` behind a keyword, I think (although it would certainly would reduce some duplication and thus less lines of code)

caspervdw:
> Now, given that for rings, you need to get exterior ring and interior rings separately (it's not a single inner loop with `GEOSGetGeometryN_r`), it won't necessarily result in more readable code by putting that inside `get_parts` behind a keyword, I think (although it would certainly would reduce some duplication and thus less lines of code)

I think we could do something on these lines:

```
cdef GetRingN(handle, polygon, n):
    if n == 0:
        return GEOSGetExteriorRing_r(handle, polygon)
    else:
        return GEOSGetInteriorRingN_r(handle, polygon, n - 1)
 ```

Then it looks much more like `GEOSGetGeometryN_r`. I think the extra overhead of the if-then will not be noticable.

**Comment by jorisvandenbossche** [comment]
> I think we could do something on these lines:

Good idea, that's indeed a clean way to combine them (and reduce the cython code). Doing some timings now to ensure it doesn't give any slowdown (but I also don't expect that)

**Comment by jorisvandenbossche** [comment]
As expected, no noticeable speed difference by combining both.

Further updated the PR to include docs and tests. Should be ready for review now.

**Review by brendan-ward — COMMENTED** [review]
@jorisvandenbossche thanks for working on this!

This seems like a case where it would be useful to also know the index or some other role identifier within the original geometry, since exterior shells and holes have different interpretations.  For example, one of my uses of something like this is to calculate the geodesic area (using external lib) of the shells and then subtract the sum of geodesic areas of the holes.

It seems there are a few ways to get at this:
- also return inner index within each polygon, where 0 always indicates the shell and 1...N always indicates holes
- also return a boolean array to indicate if it is a hole
- in addition to `get_rings`, provide this via 2 Cythonized functions: `get_exterior_rings` and `get_interior_rings`, each of which returns a structure similar to here (array of geometries and array of indices toggled via `return_index` parameter)

**Comment by jorisvandenbossche** [comment]
> This seems like a case where it would be useful to also know the index or some other role identifier within the original geometry, 

Yes. It will probably be possible to calculate most things from the current return value. Eg a mask for exterior vs interior rings could be created with `np.insert(np.diff(index), 0, 1) == 0` for holes. And for example in https://github.com/pygeos/pygeos/pull/93#issuecomment-821103456 I needed the output in yet another form, and did a `np.insert(np.bincount(index).cumsum(), 0, 0)` to get offsets.

There might be several possible outputs people might need, which might be a reason to keep the function more simple (and document some examples, as those numpy code snippets I gave are not that trivial to found yourself). But of course, if there is a typical output you often need, it would be a good convenience to provide this.

**Comment by jorisvandenbossche** [comment]
> also return inner index within each polygon, where 0 always indicates the shell and 1...N always indicates holes

This one would be similar as the "cumulative count" we did in the geopandas' explode implementation:

```
run_start = np.r_[True, outer_idx[:-1] != outer_idx[1:]]
counts = np.diff(np.r_[np.nonzero(run_start)[0], len(outer_idx)])
inner_index = (~run_start).cumsum()
inner_index -= np.repeat(inner_index[run_start], counts)
```

caspervdw (approved):
Looks good to me!

jorisvandenbossche:
@brendan-ward are you OK with going forward with this PR as is? Your comments above (https://github.com/pygeos/pygeos/pull/342#pullrequestreview-644160823) are certainly heard, but I think the current PR gives functionality that is consistent with the API of `get_parts` (and is compatible with doing a reverse operation with `polygons` after https://github.com/pygeos/pygeos/pull/346/). And further enhancements / options in what kind of information is returned (or examples to achieve this) can be left as a follow-up I think.

brendan-ward (approved):
@jorisvandenbossche sorry for the slow follow-up!  This looks good, and like you said, other things can get added at a later time and are more specific to how folks will use this, so fine to wait and see where the common needs are.

For now, I'd suggest expanding the docstring a little bit to note that that exterior ring is always the first returned for each polygon.

jorisvandenbossche:
> For now, I'd suggest expanding the docstring a little bit to note that that exterior ring is always the first returned for each polygon.

Good idea, added a sentence about this.

caspervdw:
Thanks @jorisvandenbossche  !

Issue #343, opened 2021-04-14
BUG? polygons(..) with indices gives "buffer source array is read-only" error


jorisvandenbossche:
Assume I have an array of rings, and an array of indices mapping each ring to the polygon they belong:

```python
import pygeos

# 1D array of rings with shape (10, )
rings = pygeos.linearrings(np.random.randn(10, 3, 2))
polygon_indices = np.array([0,1,1,2,2,2,3,4,4,5])
```

with some manipulation of the indices, I can get split the rings into shells and holes, and get the hole indices:

```python
polygon_indices_shell = np.insert(np.diff(polygon_indices), 0, 1)
shells_mask = polygon_indices == 1
holes_mask = polygon_indices == 0

shells = rings[shells_mask].copy()
holes = rings[holes_mask].copy()
hole_indices = polygon_indices[holes_mask].copy()
```

but passing those arguments (I even added `copy()` to be sure) gives the following error:
```
In [5]: pygeos.polygons(shells, holes, indices=hole_indices)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-76bad8a0d07f> in <module>
----> 1 pygeos.polygons(rings[shells_mask].copy(), rings[holes_mask].copy(), indices=polygon_indices[holes_mask].copy())

~/scipy/repos/pygeos/pygeos/decorators.py in wrapped(*args, **kwargs)
     54             for arr in array_args:
     55                 arr.flags.writeable = False
---> 56             return func(*args, **kwargs)
     57         finally:
     58             for arr, old_flag in zip(array_args, old_flags):

~/scipy/repos/pygeos/pygeos/creation.py in polygons(shells, holes, indices, **kwargs)
    164         return lib.polygons(shells, holes, **kwargs)
    165     else:
--> 166         return polygons_1d(shells, holes, indices)
    167 
    168 

~/scipy/repos/pygeos/pygeos/_geometry.pyx in pygeos._geometry.polygons_1d()

~/scipy/repos/pygeos/pygeos/_geometry.cpython-38-x86_64-linux-gnu.so in View.MemoryView.memoryview_cwrapper()

~/scipy/repos/pygeos/pygeos/_geometry.cpython-38-x86_64-linux-gnu.so in View.MemoryView.memoryview.__cinit__()

ValueError: buffer source array is read-only
```

Is there something that I am missing that I am doing wrong?

jorisvandenbossche:
[...]

Issue #345, opened 2021-04-15
Handle missing indices in geometry+index constructors


caspervdw:
We currently have mixed behaviour on the handling of missing indices in the geometry creation functions.

I'd like to decide on this and make it consistent. This already arised in #322.

Do we want this?

```
>>> pygeos.linestrings([[0.1, 0.1], [0.2, 0.2]], indices=[1, 1])
[None, Geometry("LINESTRING (0.1 0.1, 0.2 0.2)")]
```

Or this?

```
pygeos.multipoints([pygeos.Geometry("POINT (1 1)")] * 2, indices=[1, 1])
>>> array([<pygeos.Geometry MULTIPOINT EMPTY>,
                 <pygeos.Geometry MULTIPOINT (1 1, 1 1)>], dtype=object)
```

We basically have 3 options:
- raise an exception (like what is done when you supply 1 coordinate for a linestring)
- insert an EMPTY geometry (but; LINEARRING EMPTY does not exist, so what to do there?)
- insert a None

I am tending towards the first option. It is always better to make it more strict, then you can keep backwards compatibility when changing it later. What do you think @jorisvandenbossche @brendan-ward ?

jorisvandenbossche:
I agree that for now being strict (raising an error) is fine, as we can always relax later on, while the other way around is more difficult. 

Long term, I think it would be nice to be able to create arrays this way that *do* contain empty and missing geometries, though (eg if you want to use those functionalities to roundtrip / serialize purposes). 
Now, both probably won't be possible, since we only have a missing index to signal this. When we have to choose one, I think it might be most logical to go for an empty geometry (so the roundtripping with the combo of those functions with `get_parts`/`get_coordinates` and `return_index=True` works for all geometries). If you want to support missing values in your roundtripping, you would then need to keep track of that separately (eg with a mask).

jorisvandenbossche:
Relevant discussion from @brendan-ward at https://github.com/pygeos/pygeos/pull/322#issuecomment-805043013:

> I'm a bit torn on the handling of empty geometries / `None` / exceptions here.  Ideally, it would also be good for this to be consistent with #290, and either always return `None` when there are no indices for that output record or always return empty of that particular geometry type.  I do think that `None` seems more reasonable in this case, so I agree with the first part of the proposal (and apply that back to the geometry collections too).  An exception doesn't seem appropriate in this case.

jorisvandenbossche:
[...]

PR #346, opened 2021-04-15
Merge polygons_1d into collections_1d + finish implementation for rls 0.10

caspervdw:
Follow up from #326 .

Now `polygons` only takes 1 array of rings if used in combination with `indices`.

Also now the collection constructor raise errors if there are any indices missing (see #345). It was easier to implement the Polygons in that way.

### DIFF
--- a/pygeos/_geometry.pyx
+++ b/pygeos/_geometry.pyx
@@ -331,7 +331,7 @@ def collections_1d(object geometries, object indices, int geometry_type = 7):
                 coll = GEOSGeom_createPolygon_r(
                     geos_handle,
                     <GEOSGeometry*> temp_geoms_view[0],
-                    <GEOSGeometry**> &temp_geoms_view[1 if coll_size > 1 else 0],
+                    NULL if coll_size <= 1 else <GEOSGeometry**> &temp_geoms_view[1],
                     coll_size - 1
                 )
             else:
--- a/pygeos/_geometry.pyx
+++ b/pygeos/_geometry.pyx
@@ -217,6 +217,15 @@ cdef _deallocate_arr(void* handle, np.intp_t[:] arr, Py_ssize_t last_geom_i):
 @cython.boundscheck(False)
 @cython.wraparound(False)
 def collections_1d(object geometries, object indices, int geometry_type = 7):
+    """Converts geometries + indices to collections
+
+    Allowed geometry type conversions are:
+    - linearrings to polygons
+    - points to multipoints
+    - linestrings/linearrings to multilinestrings
+    - polygons to multipolygons
+    - any to geometrycollections
+    """
     cdef Py_ssize_t geom_idx_1 = 0
     cdef Py_ssize_t coll_idx = 0
     cdef unsigned int coll_size = 0
@@ -227,7 +236,9 @@ def collections_1d(object geometries, object indices, int geometry_type = 7):
     cdef int expected_type_alt = -1
     cdef int curr_type = -1
 
-    if geometry_type == 4:  # MULTIPOINT
+    if geometry_type == 3:  # POLYGON
+        expected_type = 2
+    elif geometry_type == 4:  # MULTIPOINT
         expected_type = 0
     elif geometry_type == 5:  # MULTILINESTRING
         expected_type = 1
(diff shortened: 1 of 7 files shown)


pygeos/creation.py
```diff
@@ -125,45 +125,50 @@ def linearrings(coords, y=None, z=None, indices=None, **kwargs):
...
-        shells = linearrings(shells)
+    geometries = np.asarray(geometries)
+    if not isinstance(geometries, Geometry) and np.issubdtype(
+        geometries.dtype, np.number
+    ):
+        geometries = linearrings(geometries)
+
+    if indices is not None:
+        if holes is not None:
+            raise TypeError("Indices provided with a holes array.")
```
jorisvandenbossche, line 158:
```suggestion
            raise TypeError("Cannot specify separate holes array when using indices.")
```

(might be slightly clearer message)

pygeos/test/test_creation_indices.py
```diff
@@ -179,13 +187,7 @@ def test_invalid_indices_collections(func, indices):
     [
         ([point, line_string], [0, 0], [geom_coll([point, line_string])]),
         ([point, line_string], [0, 1], [geom_coll([point]), geom_coll([line_string])]),
-        (
-            [point, line_string],
-            [1, 1],
-            [geom_coll([]), geom_coll([point, line_string])],
-        ),
         ([point, None], [0, 0], [geom_coll([point])]),
-        ([point, None], [0, 1], [geom_coll([point]), geom_coll([])]),
```
jorisvandenbossche, line 188: When actually passing `None` (so it's not a missing index), it might still be unambiguous to create an empty collection? The None is also skipped when it's not the only value for a given collection (the test case on the line just above)

jorisvandenbossche (commented):
Thanks! This looks good to me (and tested locally, seems to work nicely for the example case I was testing out)

pygeos/test/test_creation_indices.py
caspervdw: I think we will make things more complex if we handle the case of "no index" and "geometry is None" differently. But let's discuss this further in #345

jorisvandenbossche: That's true, but you are still handling `None` in generally, if it's not the only geometry for a given collection, so that also gives some inconsistency ..

caspervdw:
Closes #343

caspervdw:
This also adds a temporary solution for #345 (raise exceptions on missing indices / only-None inputs)

pygeos/_geometry.pyx
```diff
@@ -217,6 +217,15 @@ cdef _deallocate_arr(void* handle, np.intp_t[:] arr, Py_ssize_t last_geom_i):
 @cython.boundscheck(False)
 @cython.wraparound(False)
 def collections_1d(object geometries, object indices, int geometry_type = 7):
```
jorisvandenbossche, line 219:
```suggestion
def collections_1d(np.ndarray[object, ndim=1] geometries, object indices, int geometry_type = 7):
```

(I assume telling cython that it's an numpy array will still give better indexing performance, compared to specifying it as a generic python object)

jorisvandenbossche: Ah, there is still the `geometries = np.asarray(geometries, dtype=object)` below, so you can also type it there. 

But I think adding the type in the signature is fine, for the low level cython function it is fine to only accept arrays, I think)

pygeos/test/test_creation_indices.py
caspervdw: As described in #345, I adapted the behaviour as per your suggestion. Missing indices give errors, but indices pointing to `None` are ignored (which may lead to an EMPTY geometry creation)

pygeos/_geometry.pyx
caspervdw: I went a bit down this road. It does give you more readable code, but also, you can get segfaults from the low-level functions. 

As everything is nicely inplace now (with the `asarray` inside the cython function + checks on dimensionality) I think I rather leave the type coersion inside the function.

pygeos/creation.py
```diff
@@ -125,45 +125,51 @@ def linearrings(coords, y=None, z=None, indices=None, **kwargs):
...
+        and the holes. If not, the geometries only include the shells and the
+        ``holes`` argument should be used to create polygons with holes.
     holes : array_like or None
         An array of lists of linearrings that constitute holes for each shell.
+        Not to be used in combination with ``indices``.
     indices : array_like or None
-        Indices into the shells array where input holes belong. If
-        provided, shells, holes, and indices should all be 1D and the size
-        of holes must equal the size of indices.
+        Indices into the target array where input geometries belong. If
```
brendan-ward, line 142: A couple examples in the docstring could be very helpful for demonstrating the difference between using `indices` vs `holes`.

pygeos/test/test_creation_indices.py
```diff
@@ -61,16 +61,26 @@ def test_points_invalid():
...
+        indices=np.array([0, 1], dtype=np.intp),
+    )
+    assert_geometries_equal(actual, [point, point])
+
+
+def test_points_no_index_raises():
+    with pytest.raises(ValueError):
+        pygeos.points(
+            np.array([[2, 3], [2, 3]], dtype=float),
+            indices=np.array([0, 2], dtype=np.intp),
```
brendan-ward, line 75:
```suggestion
            indices=np.array([0], dtype=np.intp),
```

(to keep it focused on only the missing `1` index; having an index out of bounds is separate problem)

Also, shouldn't this have failed because the arrays are of different sizes (i.e., this fails for `geometrycollections` and `polygons`)?

pygeos/creation.py
```diff
@@ -125,45 +125,51 @@ def linearrings(coords, y=None, z=None, indices=None, **kwargs):
...
 
     Parameters
     ----------
-    shell : array_like
-        An array of linearrings that constitute the out shell of the polygons.
-        Coordinates can also be passed, see linearrings.
+    geometries : array_like
+        An array of linearrings or coordinates (see linearrings).
+        If ``indices`` are given, the geometries include both the outer shell
+        and the holes. If not, the geometries only include the shells and the
```
brendan-ward, line 136:
```suggestion
        and the holes; the first geometry for each index is the outer shell and all subsequent geometries in that index are the holes...
```

brendan-ward (commented):
Apologies for the long delay in review!

Provided I'm following #345 correctly, it seems like the handling of `None` should be consistent with and without indices.  I'm currently seeing this:

```python
>>> pygeos.polygons([None])
array([None], dtype=object)
>>> pygeos.polygons([None], indices=[0])
array([<pygeos.Geometry POLYGON EMPTY>], dtype=object)
```

Handling that would be an API change for the non-indices version of `polygons`, but it seems challenging to communicate that this behavior is toggled by the use of `indices`.

Otherwise, it looks very reasonable to merge collections and polygons here.  👍

pygeos/_geometry.pyx
```diff
@@ -258,9 +270,6 @@ def collections_1d(object geometries, object indices, int geometry_type = 7):
     if np.any(indices[1:] < indices[:indices.shape[0] - 1]):
         raise ValueError("The indices should be sorted.")  
 
-    cdef object[:] geometries_view = geometries
-    cdef int[:] indices_view = indices
-
```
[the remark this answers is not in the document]
jorisvandenbossche: See the discussion in #343, the memoryviews gave problems for object dtype arrays

pygeos/test/test_creation_indices.py
caspervdw: No the idea of this test is that index `1` is missing here. If I would make the indices array `[0]` I would get a different error (that indices and coordinates are not of equal length)

pygeos/creation.py
caspervdw: Thanks ,I added your suggestion, at the `indices` description

pygeos/creation.py
caspervdw: Good point, the whole module has only few examples. I added some.

caspervdw:
> Apologies for the long delay in review!
> 
> Provided I'm following #345 correctly, it seems like the handling of `None` should be consistent with and without indices. I'm currently seeing this:
> 
> ```python
> >>> pygeos.polygons([None])
> array([None], dtype=object)
> >>> pygeos.polygons([None], indices=[0])
> array([<pygeos.Geometry POLYGON EMPTY>], dtype=object)
> ```
> 
> Handling that would be an API change for the non-indices version of `polygons`, but it seems challenging to communicate that this behavior is toggled by the use of `indices`.
> 
> Otherwise, it looks very reasonable to merge collections and polygons here. +1

Thanks for the review. I adapted the "normal" `polygons` to be more consistent with the collection constructors. Now it is:

```
>>> pygeos.polygons(None)
<pygeos.Geometry POLYGON EMPTY>
>>> pygeos.multipoints([None])
<pygeos.Geometry MULTIPOINT EMPTY>
```

caspervdw:
@jorisvandenbossche @brendan-ward  is this good to merge? I'd like 0.10 to be out soon.

brendan-ward (approved):
Thanks for the changes!  The expanded docstrings help clarify the behavior.

pygeos/creation.py
```diff
@@ -125,45 +150,94 @@ def linearrings(coords, y=None, z=None, indices=None, **kwargs):
...
-        An array of lists of linearrings that constitute holes for each shell.
-    indices : array_like or None
-        Indices into the shells array where input holes belong. If
-        provided, shells, holes, and indices should all be 1D and the size
-        of holes must equal the size of indices.
-    **kwargs
-        For other keyword-only arguments, see the
-        `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
-        Ignored if ``indices`` is provided.
+    and the holes; the first geometry for each index is the outer shell and all subsequent geometries in that index are the holes...
```
jorisvandenbossche, line 155: This seems a small copy/paste or merge error. I think the line can just be removed? (it's content from the `indices` keyword explanation)

jorisvandenbossche:
> Thanks for the review. I adapted the "normal" `polygons` to be more consistent with the collection constructors. Now it is:

Indeed, given that our `pygeos.multi...(..)` constructors also return empty geometries for a list of None, that seems appropriate (although it's still a bit different, because for a polygon you don't necessarily pass a collection, but a single linearring for the outer shell)

PR #348, opened 2021-05-06
Fix use of C logical operator

jorisvandenbossche:
As noticed in https://github.com/pygeos/pygeos/pull/334/files#r627371656, we were mostly using the bitwise operators (`&`, `|`) instead of the logical operators (`&&`, `||`). See https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B#Logical_operators 
In most cases this probably won't make much difference in practice, but apart from the logical operators being more correct usage, an important difference is that those short-circuit. And `hausdorff_distance` was one case were we relied on short-circuiting (to avoid passing NULL to `GEOSisEmpty`; the test case I added currently segfaults)

### DIFF
--- a/pygeos/test/test_measurement.py
+++ b/pygeos/test/test_measurement.py
@@ -158,6 +158,8 @@ def test_hausdorff_distance_densify():
 def test_hausdorff_distance_missing():
     actual = pygeos.hausdorff_distance(point, None)
     assert np.isnan(actual)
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
There are a few cases where flags are combined (eg ` NPY_ITER_READONLY | NPY_ITER_REFS_OK`) where I left the bitwise operator as is (which I think is the correct usage there?)

PR #350, opened 2021-05-18
Remove minimum_rotated_rectangle duplicate + fix docstrings

caspervdw:
In #314 we decided to add `minimum_rotated_rectangle` as an alias for `oriented_envelope`. Now this is implemented as a duplicate (with duplicate docstrings & code).

I think it is better to not have "two things to do something". We do not have to consider legacy so there is no reason for having the duplicate there.

So instead I do `minimum_rotated_rectangle = oriented_envelope`, and hide the duplicate docstring.

Also it is not necessary (and we currently do it wrong) to show the `default` in the docstring under numpydoc convention. From https://numpydoc.readthedocs.io/en/latest/format.html#parameters: 

> Optional keyword parameters have default values, which are displayed as part of the function signature. They can also be detailed in the description: (...) or as part of the type, instead of optional.

### DIFF
--- a/pygeos/constructive.py
+++ b/pygeos/constructive.py
@@ -104,23 +104,23 @@ def buffer(
     geometry : Geometry or array_like
     width : float or array_like
         Specifies the circle radius in the Minkowski sum (or difference).
-    quadsegs : int
+    quadsegs : int, default 8
         Specifies the number of linear segments in a quarter circle in the
         approximation of circular arcs.
-    cap_style : {'round', 'square', 'flat'}
+    cap_style : {'round', 'square', 'flat'}, default 'round'
         Specifies the shape of buffered line endings. 'round' results in
         circular line endings (see ``quadsegs``). Both 'square' and 'flat'
         result in rectangular line endings, only 'flat' will end at the
         original vertex, while 'square' involves adding the buffer width.
-    join_style : {'round', 'bevel', 'mitre'}
+    join_style : {'round', 'bevel', 'mitre'}, default 'round'
         Specifies the shape of buffered line midpoints. 'round' results in
         rounded shapes. 'bevel' results in a beveled edge that touches the
         original vertex. 'mitre' results in a single vertex that is beveled
         depending on the ``mitre_limit`` parameter.
-    mitre_limit : float
+    mitre_limit : float, default 5.0
         Crops of 'mitre'-style joins if the point is displaced from the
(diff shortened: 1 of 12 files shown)


pygeos/coordinates.py
```diff
@@ -21,9 +21,9 @@ def apply(geometry, transformation, include_z=False):
     transformation : function
         A function that transforms a (N, 2) or (N, 3) ndarray of float64 to
         another (N, 2) or (N, 3) ndarray of float64.
-    include_z : bool, default False
+    include_z : bool
```
jorisvandenbossche, line 24: Personally, I find this very useful to include the default in the docstring

(yes, that information is also in the signature, but IMO it is useful to have it close to the explanation of the parameter as well)

caspervdw: Yes you're right. I changed it back and made it uniform across the rest of the files.

caspervdw:
I found a setting in sphinx to hide the module names in the docs. Now `pygeos.constructive.buffer` becomes just `buffer`.

jorisvandenbossche:
> I found a setting in sphinx to hide the module names in the docs. Now `pygeos.constructive.buffer` becomes just `buffer`.

Related, personally I think we should document our functions as how we want users to use it, so eg `pygeos.buffer` instead of `pygeos.constructive.buffer` (the `.constructive.` part is only an implementation detail I would say). But that's for another issue / PR ;) 
(I though I brought this up before, but can't directly find any discussion related to it)

pygeos/constructive.py
```diff
@@ -540,8 +540,8 @@ def polygonize(geometries, **kwargs):
     ignored.
 
     This function returns the polygons within a GeometryCollection.
-    Individual Polygons can be obtained using `get_geometry` to get
-    a single polygon or `get_parts` to get an array of polygons.
+    Individual Polygons can be obtained using ``get_geometry`` to get
+    a single polygon or ``get_parts`` to get an array of polygons.
```
jorisvandenbossche, line 544: I think in theory we could instruct sphinx to automatically generate links when using single backticks (but we are not doing that right now, so the change here is certainly an improvement)

jorisvandenbossche: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role

jorisvandenbossche (approved):

PR #360, opened 2021-06-08
TST: rename head branch for GEOS from 'master' to 'main'

mwtoews:
GEOS renamed their git development head branch from 'master' to 'main' in March 2021 ([ref](https://lists.osgeo.org/pipermail/geos-devel/2021-March/010193.html)).

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -28,7 +28,7 @@ jobs:
             numpy: 1.19.5
           # dev
           - python: 3.9
-            geos: master
+            geos: main
 
     env:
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

And also the same discussion to have here: shall we also rename our master branch to main? (which is now the default on github for new repos)  cc @caspervdw @brendan-ward

mwtoews:
NumPy (the other dependency) also renamed their head branch to 'main' in March 2021 ([ref](https://mail.python.org/pipermail/numpy-discussion/2021-March/081575.html))

PR #362, opened 2021-06-17
Fix multi-threaded STRtree query by ensuring it is built on creation

### LINKED ISSUE #361: Using STRtree in multithreading causes segfaults
While implementing and testing a spatial join in dask-geopandas (https://github.com/geopandas/dask-geopandas/pull/54), I ran into a dying kernel (segfault) once in a while. And I was now able to boil it down into a small script that for me segfaults in at least 1 out of 2 runs.

I further reduced that script to not use dask but a plain ThreadPoolExecutor, which still reproduces the issue, see below (probably I could even further boil it down to just use pygeos.STRtree instead of calling geopandas.sjoin).

Running the script gives (circa 1 out of 2 runs):

```
$ python test_parallel_sjoin.py 
double free or corruption (out)
Aborted (core dumped)
```

Running it inside gdb gives the following backtrace:

<details>

```
(gdb) run
Starting program: /home/joris/miniconda3/envs/geo-dev/bin/python test_parallel_sjoin.py
...
double free or corruption (out)

Thread 9 "python" received signal SIGABRT, Aborted.
[Switching to Thread 0x7f75e9ffb700 (LWP 1371190)]
0x00007f761a32618b in raise () from /lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007f761a32618b in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f761a305859 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f761a3703ee in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007f761a37847c in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007f761a37a120 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007f7603cdc78b in void std::vector<geos::index::strtree::AbstractNode*, std::allocator<geos::index::strtree::AbstractNode*> >::_M_realloc_insert<geos::index::strtree::AbstractNode* const&>(__gnu_cxx::__normal_iterator<geos::index::strtree::AbstractNode**, std::vector<geos::index::strtree::AbstractNode*, std::allocator<geos::index::strtree::AbstractNode*> > >, geos::index::strtree::AbstractNode* const&) ()
   from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#6  0x00007f7603cdddd1 in geos::index::strtree::STRtree::createNode(int) () from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#7  0x00007f7603cd735a in geos::index::strtree::AbstractSTRtree::createParentBoundables(std::vector<geos::index::strtree::Boundable*, std::allocator<geos::index::strtree::Boundable*> >*, int) ()
   from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#8  0x00007f7603cdd79e in geos::index::strtree::STRtree::createParentBoundablesFromVerticalSlice(std::vector<geos::index::strtree::Boundable*, std::allocator<geos::index::strtree::Boundable*> >*, int) ()
   from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#9  0x00007f7603cde1bf in geos::index::strtree::STRtree::createParentBoundablesFromVerticalSlices(std::vector<std::vector<geos::index::strtree::Boundable*, std::allocator<geos::index::strtree::Boundable*> >*, std::allocator<std::vector<geos::index::strtree::Boundable*, std::allocator<geos::index::strtree::Boundable*> >*> >*, int) () from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#10 0x00007f7603cde426 in geos::index::strtree::STRtree::createParentBoundables(std::vector<geos::index::strtree::Boundable*, std::allocator<geos::index::strtree::Boundable*> >*, int) ()
   from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#11 0x00007f7603cd68e0 in geos::index::strtree::AbstractSTRtree::createHigherLevels(std::vector<geos::index::strtree::Boundable*, std::allocator<geos::index::strtree::Boundable*> >*, int) ()
   from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#12 0x00007f7603cd6881 in geos::index::strtree::AbstractSTRtree::build() () from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#13 0x00007f7603cd6de3 in geos::index::strtree::AbstractSTRtree::query(void const*, geos::index::ItemVisitor&) () from /home/joris/miniconda3/envs/geo-dev/lib/./libgeos-3.8.1.so
#14 0x00007f7603dc712a in GEOSSTRtree_query_r () from /home/joris/miniconda3/envs/geo-dev/lib/libgeos_c.so
#15 0x00007f7603b95edf in STRtree_query_bulk (self=0x7f75fd7ee180, args=<optimized out>) at src/strtree.c:488
#16 0x0000562ba1a993bc in method_vectorcall_VARARGS (func=<optimized out>, args=0x7f75f9ba3a88, nargsf=<optimized out>, kwnames=<optimized out>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Objects/descrobject.c:300
#17 0x0000562ba1b06e09 in _PyObject_Vectorcall (kwnames=0x0, nargsf=<optimized out>, args=0x7f75f9ba3a88, callable=<method_descriptor at remote 0x7f7603bd8310>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Include/cpython/abstract.h:127
#18 call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>, tstate=0x562ba6193f30) at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Python/ceval.c:4987
#19 _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Python/ceval.c:3486
#20 0x0000562ba1ab1cd4 in _PyEval_EvalCodeWithName (_co=<code at remote 0x7f7603b08d40>, globals=<optimized out>, locals=<optimized out>, args=<optimized out>, argcount=<optimized out>, kwnames=0x0, 
    kwargs=0x7f75f9ba4bc0, kwcount=<optimized out>, kwstep=1, defs=0x7f7603b28268, defcount=1, kwdefs=0x0, closure=0x0, name='query_bulk', qualname='STRtree.query_bulk')
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Python/ceval.c:4298
#21 0x0000562ba1ab2bb5 in _PyFunction_Vectorcall (func=<optimized out>, stack=0x7f75f9ba4ba8, nargsf=<optimized out>, kwnames=<optimized out>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Objects/call.c:435
#22 0x0000562ba1a8ed65 in _PyObject_Vectorcall (kwnames=<optimized out>, nargsf=<optimized out>, args=0x7f75f9ba4ba8, callable=<function at remote 0x7f7603b2b280>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Include/cpython/abstract.h:127
#23 method_vectorcall (method=<optimized out>, args=0x7f75f9ba4bb0, nargsf=<optimized out>, kwnames=<optimized out>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Objects/classobject.c:60
#24 0x0000562ba1b0b518 in _PyObject_Vectorcall (kwnames=0x0, nargsf=<optimized out>, args=0x7f75f9ba4bb0, callable=<method at remote 0x7f75fda7d600>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Include/cpython/abstract.h:127
#25 call_function (kwnames=0x0, oparg=<optimized out>, pp_stack=<synthetic pointer>, tstate=0x562ba6193f30) at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Python/ceval.c:4987
#26 _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Python/ceval.c:3469
#27 0x0000562ba1ab2049 in _PyEval_EvalCodeWithName (_co=<code at remote 0x7f76038fe2f0>, globals=<optimized out>, locals=<optimized out>, args=<optimized out>, argcount=<optimized out>, kwnames=0x7f75fdadf3d8, 
    kwargs=0x7f75d8001440, kwcount=<optimized out>, kwstep=1, defs=0x7f76038fb518, defcount=2, kwdefs=0x0, closure=(<cell at remote 0x7f76038f6a00>,), name='query_bulk', 
    qualname='PyGEOSSTRTreeIndex.query_bulk') at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Python/ceval.c:4298
#28 0x0000562ba1ab2c58 in _PyFunction_Vectorcall (func=<optimized out>, stack=0x7f75d8001430, nargsf=<optimized out>, kwnames=<optimized out>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Objects/call.c:435
#29 0x0000562ba1a8ed65 in _PyObject_Vectorcall (kwnames=<optimized out>, nargsf=<optimized out>, args=0x7f75d8001430, callable=<function at remote 0x7f7603900700>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Include/cpython/abstract.h:127
#30 method_vectorcall (method=<optimized out>, args=0x7f75d8001438, nargsf=<optimized out>, kwnames=<optimized out>)
    at /home/conda/feedstock_root/build_artifacts/python_1587713251558/work/Objects/classobject.c:60
#31 0x0000562ba1b07cce in _PyObject_Vectorcall (kwnames=('predicate', 'sort'), nargsf=<optimized out>, args=<optimized out>, callable=<method at remote 0x7f75fda7d300>)
```

</details>

where it seems that the memory corruption comes from inside creating the STRtree (`AbstractSTRtree::build()`), but I didn't yet further investigate (probably could use better tools, or a debug build of GEOS, etc).

Currently I am reproducing this with an environment based on conda-forge with Python 3.8, GEOS 3.8.1 and pygeos master (but also see it in an environment with GEOS 3.9.1 on Python 3.9, but I don't have `gdb` installed there, and the segfault message there is "free(): invalid pointer").

The `test_parallel_sjoin.py` script (it is using the "states and provinces" from [NaturalEarth](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/); I am not able to reproduce it with the smaller countries dataset included with geopandas):

```python
from concurrent.futures import ThreadPoolExecutor
import itertools

import numpy as np
import pandas as pd
import geopandas


## Create data

# polygon data

gdf = geopandas.read_file("ne_10m_admin_1_states_provinces.zip")
# gdf = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
# for purposes of the demo, take a subset of the columns
gdf = gdf[["name", "adm0_a3", "type_en", "geometry"]]
# and drop Antartica
gdf = gdf[gdf["adm0_a3"] != "ATA"]
# and convert to Mercator (don't do this at home for global area calculation!)
gdf = gdf.to_crs("EPSG:3857")

# point data (random points in bbox of polygon layer)

minx, miny, maxx, maxy = gdf.total_bounds
N = 1_000_000
df_points = geopandas.GeoDataFrame(
    {"id": np.arange(N)},
    geometry=geopandas.points_from_xy(minx + maxx * np.random.random(N), miny + maxy * np.random.random(N)),
    crs=gdf.crs
)

## Slice parts of the dataframes -> 4x4 => 16 combinations

n = int(len(gdf) / 4)
gdf_parts = [gdf[:n], gdf[n:2*n], gdf[2*n:3*n], gdf[3*n:]]

n = int(len(df_points) / 4)
df_points_parts = [df_points[:n], df_points[n:2*n], df_points[2*n:3*n], df_points[3*n:]]


## The function calling geopandas.sjoin / pygeos.STRtree in parallel

def thread_func(idxs):
    i, j = idxs
    left = df_points_parts[i]
    right = gdf_parts[j]
    return geopandas.sjoin(left, right, op="within")


def main():
    with ThreadPoolExecutor() as pool:
        res = list(pool.map(thread_func, itertools.product(range(4), range(4))))
        print(pd.concat(res))


if __name__ == "__main__":
    main()
```

jorisvandenbossche:
Closes #361


src/strtree.c
```diff
@@ -197,6 +199,38 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
...
+      return NULL;
+    }
+  }
+  dummy = GEOSGeom_createPoint_r(ctx, coord_seq);
+  if (dummy == NULL) {
+    GEOSCoordSeq_destroy_r(ctx, coord_seq);
+    GEOSSTRtree_destroy_r(ctx, tree);
+    GEOS_FINISH;
+    return NULL;
+  }
```
jorisvandenbossche, line 229: @caspervdw do you know of a less verbose way to create a simple point?

brendan-ward: `GEOSGeom_createPointFromXY` was introduced in GEOS 3.8.  Though it adds a bit of code, you could use this in a `#if GEOS_SINCE_3_8_0` block with a fallback to the coordinate sequence varient above, so that eventually when that is the only supported version, this could be simplified.

src/strtree.c
```diff
@@ -121,6 +121,8 @@ static void STRtree_dealloc(STRtreeObject* self) {
   Py_TYPE(self)->tp_free((PyObject*)self);
 }
 
+void query_callback(void*, void*);
```
brendan-ward, line 124:
```suggestion
void dummy_query_callback(void* item, void* user_data) {}
```

src/strtree.c
```diff
@@ -197,6 +199,38 @@ static PyObject* STRtree_new(PyTypeObject* type, PyObject* args, PyObject* kwds)
...
+  }
+  dummy = GEOSGeom_createPoint_r(ctx, coord_seq);
+  if (dummy == NULL) {
+    GEOSCoordSeq_destroy_r(ctx, coord_seq);
+    GEOSSTRtree_destroy_r(ctx, tree);
+    GEOS_FINISH;
+    return NULL;
+  }
+
+  GEOSSTRtree_query_r(ctx, tree, dummy, query_callback, &query_geoms);
```
brendan-ward, line 231:
```suggestion
  GEOSSTRtree_query_r(ctx, tree, dummy, dummy_query_callback, NULL);
```

This then allows you to remove all the management required for `query_geoms` here.

brendan-ward (changes requested):
Thanks for working on this and for adding the test to prove it.  It looks like you can use a dummy query callback function that does nothing, so that you can sidestep needing to manage the `query_geoms` vector altogether.

jorisvandenbossche:
@brendan-ward thanks for the review! Good idea for the dummy callback. And moved the verbose code to create a Point to a helper function in geos.c (so the code in strtree itself is more clean to follow) + added the use of GEOSGeom_createPointFromXY_r for recent versions in that helper.

jorisvandenbossche:
So this change gives a segfault specifically on Mac with GEOS 3.6 (general segfault with STRtree, I think, not just in the threaded test). Other builds with Mac work, and a Linux build with GEOS 3.6 also works ..

brendan-ward:
The 3.6 Mac-specific failure is odd; I'd expect it to fail or pass consistently on 3.6 with Linux.

Possibly related [fix to segfault for empty tree](https://github.com/libgeos/geos/commit/ae8a2d31dcfe5c7a60c9e48bdd66886f95aa5772#diff-d510dcdc7b7a18de966b1af07c78ad3666e903f58ce69a753d02866b0e478153)

I see that our first test case is against an empty tree, so this might be it.  We specifically first check that the tree is non-empty before querying it in our main wrappers for query (`static PyObject* STRtree_query(STRtreeObject* self, PyObject* args)`):

```python
if (self->count == 0) {...return empty array...}
```

and probably need to do that here before our initial dummy query against the tree.  The result is that the tree is never built, but it is empty, so if we are always guarding for that before query / nearest, then we shouldn't run into issues in a multi-threaded environment (i.e., the threads never ultimately query the unbuilt tree).

src/strtree.c
caspervdw: Maybe use an empty point for the query? Creating one of those is certainly less verbose, but I am not sure whether a query with an empty point would trigger the tree to be built.

caspervdw:
Thanks @jorisvandenbossche for addressing this! All looks fine, I am however not a big fan of using a script plus subprocess.call in a unittest. A unittest should be simple and fast.

Why the subprocess anyway? If the test passes then you don’t need protection against a kernel abort right?

jorisvandenbossche:
@brendan-ward good catch! And indeed, not doing a dummy query for the an empty tree seems to have fixed the segfault. 

@caspervdw yeah, I used the script+subprocess approach so that you actually get a normal failing test if the script crashes, instead of crashing pytest itself. I think that's more friendly for when using the tests. But of course, once this is fixed, we don't expect to typically run into this again, so it might not be worth the extra complexity. 
No strong opinion either way, also fine with changing it to a normal unit test.

src/strtree.c
jorisvandenbossche: Looking in the GEOS code, I think it currently doesn't check if the envelope is empty before checking if the tree is built (https://github.com/libgeos/geos/blob/edd209ae551bda9cdb6188230a1861937153ce79/include/geos/index/strtree/TemplateSTRtree.h#L243-L253, it's only the `root->boundsIntersect(queryEnv)` that would never return true for an empty point), but not sure to what extent this should be considered an implementation detail.

jorisvandenbossche:
@caspervdw do you prefer me to change the tests?

caspervdw:
Maybe you can remove the subprocess-wrap? Many tests test for segfaults, it is ok to crash pytest if it happens.

Removing the test altogether is also fine.

caspervdw (approved):

jorisvandenbossche:
OK, I moved the test into a normal test function. Without the use of subprocess, it now takes ~1.5s. That's of course still considerable slower as the other tests, but that seems acceptable to keep the test, I think? (we could also have a "slow" marker that is only run on CI by default it it turns out to be annoying for quick local testing)

jorisvandenbossche:
Failures are unrelated, I think -> https://github.com/pygeos/pygeos/issues/366

caspervdw (approved):
Thanks @jorisvandenbossche for redoing the test. 

It is good to keep it, but maybe we could bundle these kind of high level tests (integration tests could be the word, not sure) at some point and also resurrect the test I did when implementing the GIL release.

jorisvandenbossche:
> maybe we could bundle these kind of high level tests (integration tests could be the word, not sure) at some point and also resurrect the test I did when implementing the GIL release.

That sounds as a good idea, yes.

Issue #354, opened 2021-05-19
Investigate cibuildwheel


caspervdw:
I recently tried out https://cibuildwheel.readthedocs.io/ for building wheels on github actions, and I must say, it is really convenient to use. I only did a small part (only the 64-bit Linux builds, and without a custom library like GEOS) but I think it must be easy to setup here as well.

@jorisvandenbossche @brendan-ward   Do you have any preference on this?

brendan-ward:
This looks promising.  I haven't built more complex binary wheels against C / C++ libs on CI services yet, so I don't have a strong preference or recommendation here.

jorisvandenbossche:
My first reaction was: my preference is to use the tools that other packages like numpy and pandas etc in the ecosystem are also using (for familiarity with those, ensuring it gets maintained etc). But I see that cibuildwheel was recently "adopted" by the PyPA (https://twitter.com/RomanYurchak/status/1395100234435502088) and for example scikit-learn started to use it as well (they switched from multibuild): https://github.com/scikit-learn/scikit-learn/pull/17921

So +1 on trying that, and since out other wheel infrastructure (travis ..) is down at the moment, I started experimenting a bit with your WIP PR.

PR #365, opened 2021-06-26
Build wheels on CI with cibuildwheel

caspervdw:
Addresses #354


jorisvandenbossche:
I experimented a bit with getting the linux build working (many commits, but an overview of the limited overall changes I made: https://github.com/pygeos/pygeos/pull/365/files/038c36089af018177f69f4f5a19f63048c86e73d..1875b77e60b6db2530b535ff2606f2b8a51ab5c3). Summary:

* Fixed the `GEOS_CONFIG` env variable (needed to point to the install location and not build location)
* Added `LD_LIBRARY_PATH` to point to the installed GEOS lib, needed (I think) for auditwheel repair to find GEOS
* Fixed the pytest command
* The tests are still failing .. (see https://github.com/pygeos/pygeos/runs/3009117547, importing pygeos fails with "ModuleNotFoundError: No module named 'pygeos.lib'"). Not directly sure what's the cause (is it not finding GEOS? Or where the C extensions not correctly built/included? )
* If I disable the tests, the first build for `manylinux2010_x86_64` passed, but then installing GEOS for `manylinux2010_i686` fails because of failing tests (https://github.com/pygeos/pygeos/runs/3009261895)

jorisvandenbossche:
Additional observation: in the last commit I disabled the 32bit linux (we also don't have that currently), so the 64bit manylinux wheel could finish and get uploaded as artifact (https://github.com/pygeos/pygeos/runs/3009573811). I downloaded that and installed it locally, and that seems to be working fine. So not fully sure why the tests are failing here on CI (something to do with path fiddling / importing from the project instead of the installed package / ... with pytest?)

jorisvandenbossche:
Based on the full traceback, it seems that it is trying to import pygeos from the /project directory, and so not the installed wheel. 

```
  ImportError while importing test module '/project/pygeos/test/test_coordinates.py'.
  Hint: make sure your test modules/packages have valid Python names.
  Traceback:
  /opt/python/cp38-cp38/lib/python3.8/importlib/__init__.py:127: in import_module
      return _bootstrap._gcd_import(name[level:], package, level)
  /project/pygeos/__init__.py:1: in <module>
      from .lib import GEOSException  # NOQA
  E   ModuleNotFoundError: No module named 'pygeos.lib'
```

caspervdw:
Thanks for working on this @jorisvandenbossche . I encountered the issue with pytest before. The challenge is to test the _installed_ package using tests from a source directory. Pytest manipulates the PATH in such a way that the (uncompiled) source files are found before the (compiled) package that was installed from the binary wheel.

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v1.12.0
+        env:
+          CIBW_BUILD: cp38-manylinux_x86_64
+          CIBW_ENVIRONMENT_LINUX: GEOS_VERSION=${{ env.GEOS_VERSION }} GEOS_INSTALL=${{ github.workspace }}/geos GEOS_BUILD=${{ github.workspace }}/geosbuild GEOS_CONFIG=${{ github.workspace }}/geos/bin/geos-config LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${{ github.workspace }}/geos/lib
+          CIBW_ENVIRONMENT_MACOS: GEOS_VERSION=${{ env.GEOS_VERSION }} GEOS_INSTALL=${{ github.workspace }}/geos GEOS_BUILD=${{ github.workspace }}/geosbuild GEOS_CONFIG=${{ github.workspace }}/geosbuild/geos-${{ env.GEOS_VERSION }}/build/bin/geos-config
+          CIBW_ENVIRONMENT_WINDOWS: GEOS_VERSION=${{ env.GEOS_VERSION }} GEOS_INSTALL=${{ github.workspace }}\geos-${{ env.GEOS_VERSION }} GEOS_LIBRARY_PATH=${{ github.workspace }}\\geos-${{ env.GEOS_VERSION }}\\lib GEOS_INCLUDE_PATH=${{ github.workspace }}\\geos-${{ env.GEOS_VERSION }}\\include
```
mwtoews, line 26: You can (probably?) split these over a few lines to improve their readability, i.e.:
```
          CIBW_ENVIRONMENT_LINUX:
            GEOS_VERSION=${{ env.GEOS_VERSION }}
            GEOS_INSTALL=${{ github.workspace }}/geos
            GEOS_BUILD=${{ github.workspace }}/geosbuild
            GEOS_CONFIG=${{ github.workspace }}/geos/bin/geos-config
            LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${{ github.workspace }}/geos/lib
```
And while I see it's not currently running, `GEOS_CONFIG` in `CIBW_ENVIRONMENT_MACOS` will need to be updated. Does `GEOS_BUILD` need to be specified?

caspervdw: Yes it works indeed. Thanks. I also enabled macos and windows, let's see what happens.

caspervdw:
@mwtoews Could you maybe take a look at the windows build? I am at a loss on how to do the slashes in the paths correctly. The current yaml leads to the following log, which seems correct:

```
environment: ParsedEnvironment(['GEOS_VERSION=3.9.1', 'GEOS_INSTALL=D:\\a\\pygeos\\pygeos\\geos-3.9.1', 'GEOS_LIBRARY_PATH=D:\\a\\pygeos\\pygeos\\geos-3.9.1\\lib', 'GEOS_INCLUDE_PATH=D:\\a\\pygeos\\pygeos\\geos-3.9.1\\include', 'PATH=$PATH:D:\\a\\pygeos\\pygeos\\geos-3.9.1\\lib'])
```

But then cmake gives the following, which is weird:
```
2021-07-14T19:26:05.1053277Z -- Installing: D:apygeospygeosgeos-3.9.1/lib/libgeos.dll.a
2021-07-14T19:26:05.1126650Z -- Installing: D:apygeospygeosgeos-3.9.1/bin/libgeos.dll
```

And ld.exe is called with `-ID:apygeospygeosgeos-3.9.1include` which definitely isn't good. It fails because it can't find the GEOS headers.

caspervdw:
I went on tinkering and what I found out is: the backslashes require escaping because the CIBW_ENVIRONMENT gets evaluated like in a bash shell. Sadly the github.workspace doesn’t have double backslashes and GitHub actions does not support inline string replacement functions.

mwtoews:
yes, it does look ridiculous, but have you tried four backslashes for each sep? `D:\\\\a\\\\pygeos`? It's possible there are two layers of escaping.

mwtoews:
nevermind, it looks like the single quotes work fine. There are two concerning things:

1. The CMake build configures a shared build with GCC, with installed `D:/a/pygeos/pygeos/geos-3.9.1/bin/libgeos_c.dll`. The normal Windows method is to MSVC, which creates `geos_c.dll` or `geos_c.lib` installed in `bin`. I'll take a look why GCC is preferred over MSVC.
2. The main error message is "LINK : fatal error LNK1181: cannot open input file 'geos_c.lib'", which implies static linking. Shouldn't this be dynamic, looking for `geos_c.dll`? Also `GEOS_LIBRARY_PATH` should be changed to the `bin` directory for Windows, not `lib`.

mwtoews (commented):
Somewhere, [MSVC2019](https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-160) needs to be initialized using:
```
call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat" x64
```
If you are planning to also support 32-bit platforms, the last parameter would need to be `x86`.

This should probably go in `ci/install_geos.cmd` (around the `pip install ...` commands). It seems that the wheelbuilding is already using `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30037\bin\HostX86\x86\cl.exe`. The cibuildwheel docs don't really say if the compiler has been activated or not.

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,61 @@
...
+            GEOS_INSTALL=${{ github.workspace }}/geos-${{ env.GEOS_VERSION }}
+            GEOS_CONFIG=${{ github.workspace }}/geos-${{ env.GEOS_VERSION }}/bin/geos-config
+            DYLD_FALLBACK_LIBRARY_PATH=${{ github.workspace }}/geos-${{ env.GEOS_VERSION }}/lib
+            LDFLAGS=-Wl,-rpath,${{ github.workspace }}/geos-${{ env.GEOS_VERSION }}/lib
+          CIBW_ENVIRONMENT_WINDOWS:
+            GEOS_VERSION=${{ env.GEOS_VERSION }}
+            GEOS_INSTALL='${{ github.workspace }}\geos-${{ env.GEOS_VERSION }}'
+            GEOS_LIBRARY_PATH='${{ github.workspace }}\geos-${{ env.GEOS_VERSION }}\lib'
+            GEOS_INCLUDE_PATH='${{ github.workspace }}\geos-${{ env.GEOS_VERSION }}\include'
+            PATH=$PATH';${{ github.workspace }}\geos-${{ env.GEOS_VERSION }}\lib'
```
mwtoews, line 45: `geos_c.dll` is in the `bin` folder. The `lib` folder is not useful for PATH (I think it's for [linker inputs](https://docs.microsoft.com/en-us/cpp/build/reference/dot-lib-files-as-linker-input?view=msvc-160)).
```suggestion
            PATH=$PATH;${{ github.workspace }}\geos-${{ env.GEOS_VERSION }}\bin
```
also, I'm guessing single quotes as not necessary for PATH (but keep them for the other variables).

mwtoews (approved):
I just tested `pygeos-0+untagged.1.ge720fd7.dirty-cp37-cp37m-win_amd64.whl` in a local venv downloaded from the latest success run, and it worked splendidly. It's nice to see how delvewheel can simplify some of the processes.

caspervdw:
Thanks @mwtoews for the review. I sorted out the LICENSE inclusion and merged the build script with the release script that @jorisvandenbossche supplied earlier.

@jorisvandenbossche  If you have time, I'd like a review. Especially the "publish" part which is basically untested. I think we should do a 0.11a1 afterwards to test the whole process.

caspervdw:
@jorisvandenbossche Are you ok with merging this and doing a prerelease of pygeos 0.10.2a1 ? I’d like to have the bugfix in #374 released and it seems to be a nice opportunity for testing the new wheels.

jorisvandenbossche:
@caspervdw sorry for the slow reply here. Merging and testing this with a new (pre)release sounds good! (I am offline this weekend so not available to help the coming days)

I didn't yet look in detail at the latest changes, but if the wheel building is all working on CI, let's try I would say. 
I only don't really like the changes to the test files, but I know that this is tricky issue with pytest (so if you got it working this way .. I will shut up :)). https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-outside-application-code recommends using the /src layout for this reason (but have no direct experience with this).

caspervdw:
Thanks for the response. I can have another look at the tests and see if I can implement some best practice. Not sure if I like the src dir approach tough, maybe including it in the distributed package is best (also e.g. for conda-forge and users compiling pygeos themselves).

caspervdw:
@jorisvandenbossche I am happy with this approach of including the tests with the distribution. Agreed?

jorisvandenbossche:
Thanks for finishing this (and nice update to the installation docs as well!)

> including it in the distributed package is best

Yes, certainly fine with that for now. We do that eg in pandas and geopandas as well. For larger packages, it might make sense to not include them (at least in wheels) to limit the package size (that's a reason in pandas we are considering splitting/removing the tests), but for pygeos that doesn't seem directly a problem (yet).

PR #367, opened 2021-06-29
BUG: fix no inplace output check for box and set_precision

### LINKED ISSUE #366: Latest numpy 1.21.0 release causes "Unknown ufunc mode" for set_precision
[Seen on a PR](https://github.com/pygeos/pygeos/runs/2929895681?check_suite_focus=true), but I think it's not related to the changes in that PR but to the bump of the numpy version, which just had a 1.21.0 release (https://numpy.org/doc/stable/release/1.21.0-notes.html#compatibility-notes):

```
 =================================== FAILURES ===================================
______________________________ test_get_precision ______________________________

    @pytest.mark.skipif(pygeos.geos_version < (3, 6, 0), reason="GEOS < 3.6")
    def test_get_precision():
        geometries = all_types + (point_z, empty_point, empty_line_string, empty_polygon)
        # default is 0
        actual = pygeos.get_precision(geometries).tolist()
        assert actual == [0] * len(geometries)
    
>       geometry = pygeos.set_precision(geometries, 1)

pygeos/test/test_geometry.py:430: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pygeos/decorators.py:70: in wrapped
    return func(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

geometry = (<pygeos.Geometry POINT (2 3)>, <pygeos.Geometry LINESTRING (0 0, 1 0, 1 1)>, <pygeos.Geometry LINEARRING (0 0, 1 0, 1...0, 2 0, 2 2, 0 2, 0 0))>, <pygeos.Geometry MULTIPOINT (0 0, 1 2)>, <pygeos.Geometry MULTILINESTRING ((0 0, 1 2))>, ...)
grid_size = 1, preserve_topology = False, kwargs = {}

    @requires_geos("3.6.0")
    @multithreading_enabled
    def set_precision(geometry, grid_size, preserve_topology=False, **kwargs):
        """Returns geometry with the precision set to a precision grid size.
    
        By default, geometries use double precision coordinates (grid_size = 0).
    
        Coordinates will be rounded if a precision grid is less precise than the
        input geometry. Duplicated vertices will be dropped from lines and
        polygons for grid sizes greater than 0. Line and polygon geometries may
        collapse to empty geometries if all vertices are closer together than
        grid_size. Z values, if present, will not be modified.
    
        Note: subsequent operations will always be performed in the precision of
        the geometry with higher precision (smaller "grid_size"). That same
        precision will be attached to the operation outputs.
    
        Also note: input geometries should be geometrically valid; unexpected
        results may occur if input geometries are not.
    
        Returns None if geometry is None.
    
        Parameters
        ----------
        grid_size : float
            Precision grid size. If 0, will use double precision (will not modify
            geometry if precision grid size was not previously set). If this
            value is more precise than input geometry, the input geometry will
            not be modified.
        preserve_topology : bool, default False
            If True, will attempt to preserve the topology of a geometry after
            rounding coordinates.
        **kwargs
            For other keyword-only arguments, see the
            `NumPy ufunc docs <https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs>`_.
    
        See also
        --------
        get_precision
    
        Examples
        --------
        >>> set_precision(Geometry("POINT (0.9 0.9)"), 1.0)
        <pygeos.Geometry POINT (1 1)>
        >>> set_precision(Geometry("POINT (0.9 0.9 0.9)"), 1.0)
        <pygeos.Geometry POINT Z (1 1 0.9)>
        >>> set_precision(Geometry("LINESTRING (0 0, 0 0.1, 0 1, 1 1)"), 1.0)
        <pygeos.Geometry LINESTRING (0 0, 0 1, 1 1)>
        >>> set_precision(None, 1.0) is None
        True
        """
    
>       return lib.set_precision(geometry, grid_size, preserve_topology, **kwargs)
E       NotImplementedError: Unknown ufunc mode with args[0]=0x7f91f8161790, args[N]=0x1, steps[0]=8, steps[N]=0, dimensions[0]=13.

pygeos/geometry.py:737: NotImplementedError
```

jorisvandenbossche:
Closes #366

### DIFF
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -33,8 +33,8 @@
 #define CHECK_NO_INPLACE_OUTPUT(N)                                                \
   if ((steps[N] == 0) && (dimensions[0] > 1)) {                                   \
     PyErr_Format(PyExc_NotImplementedError,                                       \
-                 "Unknown ufunc mode with args[0]=%p, args[N]=%p, steps[0]=%ld, " \
-                 "steps[N]=%ld, dimensions[0]=%ld.",                              \
+                 "Zero-strided output detected. Ufunc mode with args[0]=%p, "     \
+                 "args[N]=%p, steps[0]=%ld, steps[N]=%ld, dimensions[0]=%ld.",    \
                  args[0], args[N], steps[0], steps[N], dimensions[0]);            \
     return;                                                                       \
   }
--- a/src/ufuncs.c
+++ b/src/ufuncs.c
@@ -1261,7 +1261,7 @@ static void box_func(char** args, npy_intp* dimensions, npy_intp* steps, void* d
   npy_intp i;
   GEOSGeometry** geom_arr;
 
-  CHECK_NO_INPLACE_OUTPUT(6);
+  CHECK_NO_INPLACE_OUTPUT(5);
 
   // allocate a temporary array to store output GEOSGeometry objects
   geom_arr = malloc(sizeof(void*) * n);
@@ -2024,7 +2024,7 @@ static void set_precision_func(char** args, npy_intp* dimensions, npy_intp* step
   GEOSGeometry* in1 = NULL;
   GEOSGeometry** geom_arr;
 
-  CHECK_NO_INPLACE_OUTPUT(4);
+  CHECK_NO_INPLACE_OUTPUT(3);
 
   // allocate a temporary array to store output GEOSGeometry objects
   geom_arr = malloc(sizeof(void*) * dimensions[0]);


jorisvandenbossche:
In addition to our CI, this also seems to be causing issues in the conda-forge packages: https://github.com/conda-forge/pygeos-feedstock/pull/39
We should maybe do a bug-fix release with this.

caspervdw (approved):
Thanks for the quick fix @jorisvandenbossche ! Good to merge and 👍 on a bug fix release.

caspervdw:
Btw, I like @seberg suggestion to make the error message more explicit like: “Zero-strided output detected”

PR #370, opened 2021-07-06
Release workflow: automate sdist creation / GitHub release

jorisvandenbossche:
@caspervdw what do you think of this?

We have been happily using this for GeoPandas and some related packages (https://github.com/geopandas/geopandas/blob/master/.github/workflows/release_to_pypi.yml). 
Of course, here the scope is more limited (I didn't include releasing to PyPI, as for that ideally you directly upload both sdist and wheels), making this less useful. But automating the sdist creation, and uploading it to a new GitHub release still could be useful,I think (not fully sure, the sdist creation is also not a lot of work, you only need to be careful about not including eg cythonized files, which is safer on a new checkout in the github action). 

The workflow would then be: 1) tag locally and push the tag (as before), 2) let the action run and create the github release, 3) update the github release notes 4) build wheels and upload those together with sdist to PyPI (same as before).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,60 @@
+name: Release and publish pygeos to GitHub
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  build-n-publish:
+    name: Build and publish pygeos to GitHub
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Build a source tarball
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools
+          python setup.py sdist
+
+      # - name: Publish distribution to PyPI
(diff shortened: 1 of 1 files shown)


caspervdw:
This looks handy, I am :+1: on automating this!
Why exactly don’t you want to include the sdist upload? It is perfectly fine to do the wheels afterwards (that is what happens now at least)

jorisvandenbossche:
> Why exactly don’t you want to include the sdist upload? It is perfectly fine to do the wheels afterwards (that is what happens now at least)

Ideally, I think you only upload the sdist if also the wheels are ready. Otherwise, you have a period in between where people doing `pip install pygeos` will end up building from source instead of installing a wheel because there is not yet a wheel available for the latest release.

caspervdw:
I see, good to know for the next release.

We may automate this then after #365 is worked out.

caspervdw (approved):

jorisvandenbossche:
This seems to have worked! -> https://github.com/pygeos/pygeos/releases/tag/0.10.1 
(after a small fix I pushed to master)

Indeed, if we also automate the wheel building further, that could be integrated with the release step.

PR #374, opened 2021-07-31
FIX: Handle NULL in object-dtype arrays

### LINKED ISSUE #373: Pygeos segfaults on arrays constructed with np.empty_like
Not sure what is happening here:

```
In [1]: import pygeos
In [2]: import numpy as np
In [3]: pygeos.is_missing(np.full((10,), None, dtype=object))
OK
In [4]: pygeos.is_missing(np.empty((10,), dtype=object))
OK
In [5]: pygeos.is_missing(np.empty_like(np.empty((100,)), dtype=object))
```

It doesn't matter which pygeos function you take, the behaviour is consistent: a geometry input constructed with `np.empty_like` results in a segfault.

caspervdw:
Closes #373 

Apparently, Numpy treats NULL the same as None in object-dtype arrays.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ Version 0.11 (unreleased)
 
 **Bug fixes**
 
+* Fixed operations on geometry arrays containing NULL instead of None.
+  These occur for instance by using ``numpy.empty_like`` (#371)
+
 **Acknowledgements**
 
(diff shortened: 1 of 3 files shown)


brendan-ward (approved):
Thanks for fixing this, @caspervdw

jorisvandenbossche (approved):
Thanks!

PR #378, opened 2021-08-19
[Done] Disallow linearrings with 3 coordinates in GEOS 3.10

caspervdw:


jorisvandenbossche:
See also my comments at https://github.com/pygeos/pygeos/issues/233#issuecomment-849483828. I don't know how much I looked at it in detail at that point, but if such linearrings are invalid, it might be nice to continue raise an error for it in the `linearrings` constructor

brendan-ward:
> continue raise an error for it in the linearrings constructor

I agree.  Since `linearrings` is invoked within the user's control to provide better coordinates, invalid rings should be caught.

In contrast, we should allow invalid linear rings that come in from WKB / WKT so that they can at least be deserialized and then repaired.

So as a general rule I'm thinking (outside scope of this PR, but just adding here for context):

* create invalid geometry manually => exception
* create invalid geometry from serialized form => warning

caspervdw:
The relevant issue is this one: https://github.com/libgeos/geos/issues/406

The amount of coordinates allowed in a linearring was deliberately reduced to 3. I think it is best to not build different logic in pygeos and allow construction of 3-coordinate rings.

caspervdw:
I was reading up on the issue and I actually think @brendan-ward 's idea is a good general guideline, which I also like to implement for NaN-valued coordinates which lead to invalid geometries also:

>    create invalid geometry manually => exception
>    create invalid geometry from serialized form => warning

This could be done in this PR.

caspervdw:
Done, I made it a `ValueError`. The NaN stuff can be addressed another time.

jorisvandenbossche:
Regarding the general rule of

> create invalid geometry manually => exception 
> create invalid geometry from serialized form => warning

I agree with it for the default behaviour, but just want to point out that also the "manual" creation functions can be used in a deserialization context (if the format is storing raw coordinates arrays instead of some custom binary/text format). So I think ideally even for those manual creation functions there would be an *option* to have it only raise a warning and not error.

jorisvandenbossche:
Now, for linearrings of 3 coordinates that's maybe less of an issue (I think for NaNs it might be more important to have an escape hatch)

jorisvandenbossche (approved):
Code changes look good!

PR #379, opened 2021-08-23
Handle linearrings in is_closed

### LINKED ISSUE #337: Consider LinearRing as "closed"?
I didn't directly find a previous discussion about this, but currently the `is_closed` ufunc will always return False for LinearRings:

```python
>>> ring = pygeos.linearrings(((0, 0), (0, 1), (1 ,1 ), (1, 0)))
>>> print(ring)
LINEARRING (0 0, 0 1, 1 1, 1 0, 0 0)
>>> pygeos.is_closed(ring)
False
```

This is following GEOS' behaviour where `GEOSisClosed` only works for LineStrings and not any other geometry (and thus also not LinearRings). 

However, in Shapely, a LinearRing is also considered as "closed" (by simply hardcoded always True for a LinearRing):

```python
>>> from shapely.geometry import LinearRing
>>> ring = LinearRing( ((0, 0), (0, 1), (1 ,1 ), (1, 0)))
>>> print(ring)
LINEARRING (0 0, 0 1, 1 1, 1 0, 0 0)
>>> ring.is_closed
True
```

PostGIS (https://postgis.net/docs/ST_IsClosed.html) seems to return True as well. It's not directly documented, but from a quick test (creating a polygon and extracting the exterior ring, as creating the LinearRing from WKT doesn't seem to be supported):

```
test_db=# SELECT ST_IsClosed(ST_ExteriorRing('POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))'::geometry));
 st_isclosed 
-------------
 t
(1 row)
```

Although PostGIS also returns True for Polygon

caspervdw:
Closes #337

Turns out GEOS is perfectly capable to determine whether a linearring is closed. Our code was just in the way.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Version 0.11 (unreleased)
 
 **Bug fixes**
 
-* ...
+* Return True instead of False for LINEARRING geometries in ``is_closed`` (#379).
 
 **Acknowledgments**
 
--- a/pygeos/tests/test_predicates.py
+++ b/pygeos/tests/test_predicates.py
@@ -4,7 +4,15 @@ import pytest
 import pygeos
 from pygeos import Geometry
 
-from .common import all_types, empty, geometry_collection, point, polygon
+from .common import (
+    all_types,
+    empty,
+    geometry_collection,
+    line_string,
+    linear_ring,
+    point,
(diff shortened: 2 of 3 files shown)


jorisvandenbossche (approved):
Looks good!

PR #380, opened 2021-08-23
Handle 'out' keyword argument in constructors if indices are given

### LINKED ISSUE #345: Handle missing indices in geometry+index constructors

caspervdw:
Closes #345

### DIFF
--- a/pygeos/_geometry.pyx
+++ b/pygeos/_geometry.pyx
@@ -67,7 +67,7 @@ def _check_out_array(object out, Py_ssize_t size):
     if out.ndim != 1:
         raise TypeError("out is not a one-dimensional array.") 
     if out.shape[0] < size:
-        raise ValueError("out array is to small ({out.shape[0]} < {size})") 
+        raise ValueError(f"out array is to small ({out.shape[0]} < {size})") 
     return out
 
  
@@ -116,6 +116,9 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type, 
 
     # The final target array
     cdef Py_ssize_t n_geoms = coord_counts.shape[0]
+    # Allow missing indices only if 'out' was given explicitly (if 'out' is not
+    # supplied by the user, we would have to come up with an output value ourselves).
+    cdef char allow_missing = out is not None
     out = _check_out_array(out, n_geoms)
     cdef object[:] out_view = out
 
@@ -123,11 +126,13 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type, 
         for geom_idx in range(n_geoms):
             geom_size = coord_counts[geom_idx]
 
-            # for now, raise if there are indices missing (decision on this in GH345)
             if geom_size == 0:
-                raise ValueError(
-                    f"Index {geom_idx} is missing from the input indices."
-                )
+                if allow_missing:
+                    continue
+                else:
+                    raise ValueError(
+                        f"Index {geom_idx} is missing from the input indices."
+                    )
 
             # check if we need to close a linearring
             if geometry_type == 2:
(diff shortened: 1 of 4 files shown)


pygeos/_geometry.pyx
```diff
@@ -54,10 +54,26 @@ cdef char _set_xyz(GEOSContextHandle_t geos_handle, GEOSCoordSequence *seq, unsi
...
+    if out is None:
+        return np.empty(shape=(size, ), dtype=object)
+    if not isinstance(out, np.ndarray):
+        raise TypeError("out array must be of numpy.ndarray type")
+    if not out.flags.writeable:
+        raise TypeError("out array must be writeable")
+    if out.dtype != object:
+        raise TypeError("out array dtype must be object")
+    if out.ndim != 1:
+        raise TypeError("out is not a one-dimensional array.")
```
brendan-ward, line 68:
```suggestion
        raise TypeError("out must be a one-dimensional array.") 
```

pygeos/creation.py
```diff
@@ -64,7 +64,9 @@ def points(coords, y=None, z=None, indices=None, **kwargs):
     if indices is None:
         return lib.points(coords, **kwargs)
     else:
-        return simple_geometries_1d(coords, indices, GeometryType.POINT)
+        return simple_geometries_1d(
+            coords, indices, GeometryType.POINT, out=kwargs.get("out")
```
brendan-ward, line 68: Is there are a reason for not making `out` a top-level parameter for `points` (or other top-level functions below)?

caspervdw: I added it and fixed the docstrings.

caspervdw:
Thanks for the review! I addressed your points.

> It seems like the dtype of `out` is intercepted by numpy instead of the validation you added here
> 
> ```
> >>> pg.points([[0,0]], out=np.empty(2))
> Traceback (most recent call last):
>   ...
> TypeError: ufunc 'points' output (typecode 'O') could not be coerced to provided output parameter (typecode 'd') according to the casting rule ''same_kind''
> ```
> 
> Whereas your validation does run when `indices` is provided:
> 
> ```
> >>> pg.points([[0,0]], indices=[0], out=np.empty(2))
> Traceback (most recent call last):
>   ...
>   File "pygeos/_geometry.pyx", line 66, in pygeos._geometry._check_out_array
>     raise TypeError("out array dtype must be object")
> TypeError: out array dtype must be object
> ```

This is what I would expect indeed. To be clear: the validation I added is only ran when providing `indices`. Only then the Cython code is touched. If `indices` are not provided, the code flows straight into the numpy ufunc in which case numpy validates the kwargs (`out` etc.)

PR #381, opened 2021-08-25
[Done] Add pygeos.empty

### LINKED ISSUE #149: Allow creating empty geometries in constructive functions
Currently, those functions like `polygons` etc raise an error if you try:

```
In [13]: pygeos.polygons(np.random.randn(3, 2))
Out[13]: <pygeos.Geometry POLYGON ((1.07 0.401, 0.386 -0.151, 0.133 0.257, 1.07 0.401))>

In [15]: pygeos.polygons(np.empty((0, 2)))  
<ipython-input-15-301ca2e69111> in <module>
----> 1 pygeos.polygons(np.empty((0, 2)))

~/scipy/repos/pygeos/pygeos/creation.py in polygons(shells, holes)
     85     shells = np.asarray(shells)
     86     if not isinstance(shells, Geometry) and np.issubdtype(shells.dtype, np.number):
---> 87         shells = linearrings(shells)
     88 
     89     if holes is None:

~/scipy/repos/pygeos/pygeos/creation.py in linearrings(coords, y, z)
     69     z : array_like
     70     """
---> 71     return _wrap_construct_ufunc(lib.linearrings, coords, y, z)
     72 
     73 

~/scipy/repos/pygeos/pygeos/creation.py in _wrap_construct_ufunc(func, coords, y, z)
     18 def _wrap_construct_ufunc(func, coords, y=None, z=None):
     19     if y is None:
---> 20         return func(coords)
     21     x = coords
     22     if z is None:

GEOSException: IllegalArgumentException: point array must contain 0 or >1 elements
```

This is a GEOS exception, so I am not fully this can be worked around (didn't check yet). This is from the linearrings function, so we might need a custom check if we want to allow this.

caspervdw:
Closes #149 

Some thoughts: I first wanted to make this a ufunc and use the GEOS API to construct the empty geometries with code.
But this approach is much simpler. It does have some ~10 microseconds overhead from WKT deserialization, but as it is O(1) (not scaling with shape) I think it is acceptable.

Some timings:
```
In [5]: %timeit pygeos.empty((1000,), geom_type=1)
18.7 µs ± 147 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [12]: fill_value = pygeos.Geometry("POINT EMPTY")

In [13]: %timeit np.full((1000,), fill_value=fill_value, dtype=object)
8.46 µs ± 40.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,8 @@ Version 0.11 (unreleased)
 
 **Major enhancements**
 
-* ...
+* Added ``pygeos.empty`` to create a geometry array pre-filled with None or
+  with empty geometries (#381).
 
 **API Changes**
 
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -16,6 +16,7 @@ __all__ = [
     "box",
     "prepare",
     "destroy_prepared",
+    "empty",
 ]
 
 
@@ -480,3 +481,26 @@ def destroy_prepared(geometry, **kwargs):
     prepare
     """
     lib.destroy_prepared(geometry, **kwargs)
+
+
+def empty(shape, geom_type=None, order="C"):
+    """Create an geometry array prefilled with None or empty geometries.
+
+    Parameters
+    ----------
+    shape : int or tuple of int
+        Shape of the empty array, e.g., ``(2, 3)`` or ``2``.
+    geom_type : pygeos.GeometryType, optional
+        The desired geometry type in case the array should be prefilled
+        with empty geometries. Default ``None``.
+    order : {'C', 'F'}, optional, default: 'C'
+        Whether to store multi-dimensional data in row-major
+        (C-style) or column-major (Fortran-style) order in
+        memory.
+    """
+    if geom_type is None:
+        return np.empty(shape, dtype=object, order=order)
+    else:
+        wkt = GeometryType(geom_type).name + " EMPTY"
+    fill_value = Geometry(wkt)
+    return np.full(shape, fill_value, dtype=object, order=order)
--- a/pygeos/tests/test_creation.py
+++ b/pygeos/tests/test_creation.py
@@ -468,3 +468,16 @@ def test_subclass_is_geometry(with_point_in_registry):
 
 def test_subclass_is_valid_input(with_point_in_registry):
     assert pygeos.is_valid_input(Point("POINT (1 1)"))
+
+
+def test_empty_none():
+    actual = pygeos.empty((2,))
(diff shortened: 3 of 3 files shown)


caspervdw:
> Thanks!
> 
> Two (late) thoughts:
> 
>     * Would it make sense to also allow to create a scalar geometry with this? For example when not passing a shape. In general I might find that useful, to avoid having to fallback to the "EMPTY" WKT parsing yourself (that's what I currently did in the Shapely geometry constructors, with a TODO to have a better way to create empty geometries with pygeos (eg https://github.com/Toblerity/Shapely/blob/6821fb899a6ca5de467b168f80a4ce57f1f8e293/shapely/geometry/point.py#L50-L51), but maybe there is no better way? `pygeos.empty(pygeos.GeometryType.POINT)` is not necessarily easier or cleaner than `pygeos.from_wkt("POINT EMPTY")`)
> 
>     * About the name, it's not really creating an empty array, but an array of empty geometries, so there might be some confusion with numpy's `np.empty` (it's actually more alike `np.ones` or `np.zeros`). So potentially a name like `pygeos.empty_geometries(..)` might be more explicit (although longer).
>       (of course, if we would go for the suggestion above allowing to create a scalar, the plural in the function name is also a bit weird for the scalar case. Although that's also something we have with `pygeos.points` and friends).

Thanks for sharing the thoughts! The name `empty` was actually deliberately chosen the same as `np.empty`. This because in the default behavior, `pygeos.empty` just calls `np.empty(shape, dtype=object, **kwargs)`. I wanted to keep the API close to what users already know to use from numpy.

So when a `geom_type` is chosen, the array is indeed not empty, but the individual elements are. That is more like `np.zeros` indeed. I am not sure if there is sufficient confusion here for us to worry about. The docstring is pretty clear what happens if you provide `geom_type`:

```
The desired geometry type in case the array should be prefilled
with empty geometries. Default ``None``.
```

And the scalar case gets a little more verbose:

```
>>> pygeos.empty((), 0).item()
<pygeos.Geometry POINT EMPTY>
```

One option to change the API is:
- drop the `np.empty` functionality (geom_type=None case) - users should just use numpy
- rename to `empty_geometries`. I don't think the long name or the plural form are an issue.
- default `shape=()` and in that case automatically convert to a scalar for the user. Then the user only needs to do:

```
>>> pygeos.empty_geometries(0)
<pygeos.Geometry POINT EMPTY>
```

Again, I like the current implementation also. Not sure what to do here. @brendan-ward  any thoughts?

PR #387, opened 2021-09-04
[Done] Write docs for runtime library finding + clean CI runners

### LINKED ISSUE #383: Issue installing PyGEOS on windows with Python 3.8+
I encountered this issue in #382 . For some reason I am getting this when importing `pygeos`:

```
 Traceback:
c:\hostedtoolcache\windows\python\3.8.10\x64\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
pygeos\__init__.py:1: in <module>
    from .lib import GEOSException  # NOQA
E   ImportError: DLL load failed while importing lib: The specified module could not be found.
```
Complete log: https://github.com/caspervdw/pygeos/runs/3455987095

This happens with any GEOS version, on Python 3.8 and 3.9 only. First I thought it was an issue with setuptools, but I can reproduce it locally (using OSGeo4W build of GEOS) with setuptools 47, 49, 56, and 57 (latest). One question here is: why does Appveyor work? Do you have any ideas @mwtoews ?

With a Dependency checker (This one: https://github.com/lucasg/Dependencies) I cannot find any broken dependency locally.

caspervdw:
For finding GEOS at runtime, we may use the `runtime_library_dirs` option of distutils/setuptools. Then you do not have to set `LD_LIBRARY_PATH` or its equivalent on OSX. I think it is a good idea to not ask users of setting `LD_LIBRARY_PATH` - it feels symptomatic of a bad installer.

However, there are some quirks here: distutils is not able to give the correct arguments to the linker in all cases (especially on OSX). On Windows such concept doesn’t even exist.  “runtime_library_dirs” effectively only works on Linux.

More importantly, pygeos links directly to geos_c.so and indirectly to geos.so. For the common GNU ld linker, runpaths only work for directly linked shared libs. So I ended up with errors “Cannot find geos.so”. There may be a way around but that would involve some hack or a serious addition of functionality on top of setuptools/distutils — something I don’t want to spend too much time on.

So we are left with advising the user on copying the shared libraries to a location findable by Python, or appending the dynamic linker paths.

This closes #383.


caspervdw:
@brendan-ward Sorry about the explorative work here while having requested a review already. I now made up my mind and adapted the PR description.

Could you or @jorisvandenbossche take a look?

caspervdw:
I’d like to merge this — the build on master fails due to an issue on OSX that this PR fixes.

ci/install_geos.sh
```diff
@@ -32,14 +32,18 @@ prepare_geos_build_dir(){
...
-    pip install cmake ninja
+    pip install cmake
 
     echo "Building geos-$GEOS_VERSION"
     mkdir build
     cd build
-    cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=$GEOS_INSTALL ..
-    cmake --build .
-    # ctest .
+    # Use Ninja on windows, otherwise, use the platform's default
```
jorisvandenbossche, line 40: Is there a reason for not doing this differently depending on the platform?

docs/installation.rst
```diff
@@ -98,27 +98,26 @@ GEOS discovery (runtime)
 ------------------------
 
 PyGEOS is dynamically linked to GEOS. This means that the same GEOS library that was used
-during PyGEOS compilation is required on your system. Make sure that the dynamic linker paths are
-set such that the libraries are found.
+during PyGEOS compilation is required on your system at runtime. When using pygeos that was distributed
+as a binary wheel or through conda,this is automatically the case and you can stop reading.
```
jorisvandenbossche, line 102:
```suggestion
as a binary wheel or through conda, this is automatically the case and you can stop reading.
```

docs/installation.rst
```diff
@@ -98,27 +98,26 @@ GEOS discovery (runtime)
...
+to each other. We only include some guidelines here to address this issue as this document is
+not intended as a general guide of shared library finding.
 
-On Linux::
-
-    $ export LD_LIBRARY_PATH=/path/to/geos/lib:$LD_LIBRARY_PATH
-    $ sudo ldconfig  # refresh dynamic linker cache
-
-On OSX::
+There are in general four ways of making Python aware of the location of shared library:
```
jorisvandenbossche, line 108: On linux, if GEOS was automatically found during compile time, it also won't be found during runtime?

(I am totally no expert here, but just thinking about what I do myself: for example for development, I am building pygeos from source (and thus not using a binary wheel or conda package), and I don't think I need to do any of the 4 options below to have it working)

jorisvandenbossche (commented):
(I am not very familiar with this topic, so can't really give an adequate review)

docs/installation.rst
caspervdw: On Linux the libraries must be on the “ld” path at runtime. You can check if that is the case using “ldconfig -p | grep geos”

If you don’t have to do any of the 4 options, probably one of them is already the case. Maybe geos’ cmake puts something in your /usr/local/lib?

ci/install_geos.sh
caspervdw: Cmake happens te pick a working default on the posix platforms. On windows, it picks the “Visual Studio 2019” generator, which fails.

mwtoews: you can set an environment variable [`CMAKE_GENERATOR=Ninja`](https://cmake.org/cmake/help/latest/envvar/CMAKE_GENERATOR.html) (either here or in the YAML file) for windows to override the default. Also, GEOS' default for `BUILD_SHARED_LIBS` for is `ON`, so there is no need for the following if/else/fi.

caspervdw: Thanks for the feedback @mwtoews I actually tried that in this commit: https://github.com/pygeos/pygeos/pull/387/commits/c471ade8dde611dd901d7ccf77425c8062b6566b but it didn’t seem to work. I didn’t want to spend too much time on that so I went for the if … else … fi. Can you discover what’s wrong in that commit?

docs/installation.rst
jorisvandenbossche: In my case it's a conda environment, so GEOS is located in the `/lib` directory of the environment (so I suppose the equivalent of `/usr/local/lib` for the environment).

It's just that I would find the current paragraph a bit confusing. Maybe more clearly say that one of those actions should only be used if not working out of the box (I now incorrectly got the impression that I would _need_ to do one of those 4 actions to get it working).

ci/install_geos.sh
mwtoews: The issue was setting the variable, which cannot have spaces, and needs to be exported. Give this a try:
```
    # Use Ninja on windows, otherwise, use the platform's default
    if [ "$RUNNER_OS" = "Windows" ]; then
        export CMAKE_GENERATOR=Ninja
    fi
```

docs/installation.rst
caspervdw: I worked a bit on the paragraph, I hope it is more clear now

ci/install_geos.sh
caspervdw: Thanks @mwtoews , I pushed the changes. Let's see what happens.

caspervdw: It worked, only now a new issue popped up. So I removed the “skip building test” option which apparently broke the GEOS 3.8 build.

jorisvandenbossche (approved):
Updated text looks good!

Issue #336, opened 2021-04-05
WKB representation of empty 3D Point is 2-dimensional


jorisvandenbossche:
I am a bit lost in the previous discussions we had about this topic of WKB and empty points (and GEOS version dependencies .. etc), but noticed the following difference between Shapely and PyGEOS while running the Shapely tests (in the shapely-2.0 branch) with the latest pygeos:

```
>>> pygeos.to_wkb(pygeos.from_wkt("POINT Z EMPTY"), hex=True)
'0101000000000000000000F87F000000000000F87F'
```

This in contrast to Shapely (using the same GEOS 3.9.1 version):

```
>>> from shapely import wkt

>>> wkt.loads("POINT Z EMPTY").wkb_hex
'0101000080000000000000F87F000000000000F87F000000000000F87F'
```

caspervdw:
The problem seems in the construction of the point:

```
In [1]: import pygeos
In [2]: pygeos.Geometry("POINT Z EMPTY")
Out[2]: <pygeos.Geometry POINT EMPTY>
In [3]: pygeos.from_wkt("POINT Z EMPTY")
Out[3]: <pygeos.Geometry POINT EMPTY>
In [4]: pygeos.from_wkb("0101000080000000000000F87F000000000000F87F000000000000F87F")
Out[4]: <pygeos.Geometry POINT EMPTY>
In [5]: from shapely import wkt
In [6]: pygeos.from_shapely(wkt.loads("POINT Z EMPTY"))
Out[6]: <pygeos.Geometry POINT EMPTY>
```

Looking at the tests, this issue appeared only since GEOS 3.8.0. I can't find any reason why shapely would be able to read 3D empty points and why pygeos is not...

jorisvandenbossche:
On the other hand, when checking the coordinate dimension, the _construction_ actually seems to have catched that it is 3D:

```
In [56]: pygeos.get_coordinate_dimension(pygeos.Geometry("POINT Z EMPTY"))
Out[56]: 3

In [57]: pygeos.get_coordinate_dimension(pygeos.Geometry("POINT EMPTY"))
Out[57]: 2
```

So it might also be an issue in the conversion to WKB/WKT. Although for WKT, I don't see anything obvious differently in the settings of the WKTWriter (3D is enabled in both, and for non-empty it's also working correctly).

jorisvandenbossche:
[...]

PR #392, opened 2021-09-15
[Done] Fix WKB/WKT of empty (multi)points on GEOS 3.9

caspervdw:
This fixes wkb serialisation of 3D empty points in GEOS>= 3.9.0

In #336 it appeared that GEOS 3.9.1 does this correctly and we don’t need our custom code (that doesn’t work for 3D empty points)

Two issues remain: wkt serialisation and has_z


pygeos/tests/test_io.py
```diff
@@ -375,17 +375,24 @@ def test_to_wkb_srid():
 
 
 @pytest.mark.skipif(
-    pygeos.geos_version >= (3, 8, 0), reason="Pre GEOS 3.8.0 has 3D empty points"
+    pygeos.geos_version < (3, 9, 0),
+    reason="Pre GEOS 3.9 empty point serialization is buggy",
```
jorisvandenbossche, line 379: Just to ensure I am correctly reading this: you removed testing for this for GEOS <= 3.8? (which I am fine with, given the version dependent behaviour, as long as we can guarantee correct and stable behaviour starting with GEOS 3.9)

pygeos/tests/test_io.py
```diff
@@ -375,17 +375,24 @@ def test_to_wkb_srid():
...
 )
 @pytest.mark.parametrize(
     "geom,dims,expected",
     [
         (empty_point, 2, POINT_NAN_WKB),
-        (empty_point, 3, POINTZ_NAN_WKB),
+        (empty_point, 3, POINT_NAN_WKB),
```
jorisvandenbossche, line 385: This behaviour changed? (which seems a correct change though)

pygeos/tests/test_io.py
```diff
@@ -394,42 +401,21 @@ def test_to_wkb_srid():
...
-            NESTED_COLLECTION_NAN_WKB,
+            NESTED_COLLECTIONZ_NAN_WKB,
         ),
     ],
 )
-def test_to_wkb_point_empty_post_geos38(geom, dims, expected):
+def test_to_wkb_point_empty(geom, dims, expected):
     # Post GEOS 3.8: empty point is 2D
     actual = pygeos.to_wkb(geom, output_dimension=dims, byte_order=1)
     # Use numpy.isnan; there are many byte representations for NaN
```
jorisvandenbossche, line 421: [about the line below] I find it a bit strange that the below works, while it is harcoded for 2 NaNs but some expected results have 3 NaNs (it seems the removed pre_geos38 test used `-dims` instead of a harcoded `-2`

(or maybe it would only fail on platforms we are not testing in CI?)

pygeos/tests/test_io.py
```diff
@@ -394,42 +401,21 @@ def test_to_wkb_srid():
...
+            pygeos.geometrycollections([pygeos.multipoints([empty_point_z])]),
             3,
-            NESTED_COLLECTION_NAN_WKB,
+            NESTED_COLLECTIONZ_NAN_WKB,
         ),
     ],
 )
-def test_to_wkb_point_empty_post_geos38(geom, dims, expected):
+def test_to_wkb_point_empty(geom, dims, expected):
     # Post GEOS 3.8: empty point is 2D
```
jorisvandenbossche, line 419: Maybe update this comment?

pygeos/tests/test_io.py
jorisvandenbossche: (although there is now a whole block of code in geos.c between `#if !GEOS_SINCE_3_9_0` / `#endif` that isn't really tested at all? Or is it still tested through other tests?)

jorisvandenbossche (commented):
Thanks a lot for looking into this! Added a few questions/comments

pygeos/tests/test_io.py
caspervdw: You’re right, there should be *some* testing in GEOS<3.9, now the code isn’t touched anymore in the unittests.

Not sure what exactly to test. I don’t like “confirming faulty behaviour” which the previous tests did. Maybe just run the function and make sure it doesn’t segfault. I’ll remove the skip to see what exactly we have to do on earlier versions.

pygeos/tests/test_io.py
caspervdw: On GEOS < 3.8, to_wkb(empty_point, output_dimension=3) indeed gave the (faulty) 3D WKB string
On GEOS ==3.8, this works correctly, but you always get a 2D WKB (also for empty_point_z
On GEOS >= 3.9, everything works fine.

I split the test into 3 to make this more clear and updated the docs.

caspervdw:
@jorisvandenbossche  This is now done, I also removed the protection for writing `MULTIPOINT (EMPTY, 2 3)` because that has been solved in GEOS 3.9 as well.

jorisvandenbossche (approved):
Looks good, thanks a lot!

Issue #323, opened 2021-03-25
BUG: if first z-coordinate is NaN, has_z incorrectly reports False


woest:
It seems that pygeos incorrectly reports that 'has_z' is false in case the first z-coordinate is NaN. This leads to the z-dimension getting culled in shapely/geopandas.

Traced this bug after getting funky results in geopandas.
https://github.com/geopandas/geopandas/issues/1888

```python
import numpy as np
import pygeos
import shapely.wkb

coords1 = [
    (1, 2, 3),
    (1, 1, np.NaN),
]

coords2 = [
    (1, 1, np.NaN),
    (1, 2, 3),
]

for coords in [coords1, coords2]:
    geom1 = pygeos.creation.linestrings(coords)
    print(pygeos.geometry.get_coordinate_dimension(geom1))
    print(pygeos.predicates.has_z(geom1))
    print(pygeos.coordinates.get_coordinates(geom1, include_z=True))
    geom2 = shapely.wkb.loads(pygeos.to_wkb(geom1))
    print(np.array(geom2.coords))
    print('\n')
```

#### Output

```python
3
True
[[ 1.  2.  3.]
 [ 1.  1. nan]]
[[ 1.  2.  3.]
 [ 1.  1. nan]]


3
False
[[ 1.  1. nan]
 [ 1.  2.  3.]]
[[1. 1.]
 [1. 2.]]
```

caspervdw:
Thanks @woest for your report. I looked into this and pygeos uses the GEOS function `GEOSHasZ_r` to determine the return value of `has_z`. So we can't really solve this here in pygeos.

There is a remarks by @pramsey in https://github.com/pygeos/pygeos/issues/306 
> Basically NaNs make everything bad, don't let them in.

So, you can't really trust any operation on a geometry that has nans. 

I am worried though about these funky results. Maybe it is better to prohibit nans in geometry constructors, by replacing the complete geometry with `None` or raising an exception. What do you think @brendan-ward @jorisvandenbossche ?

brendan-ward:
I think raising an exception would be best here; when these cases occur in intermediate steps, I want things to halt at the exception and then I know something is wrong and need to fix it.  If we instead return `None` that may not be obvious until much later steps, or not at all, depending on subsequent operations.

woest:
So NaN's are not allowed at all? I can imagine this for x and y, but z?

[...]

PR #396, opened 2021-09-26
[Done] Force 2D/3D

caspervdw:
This is a first step to addressing #323 .

### DIFF
--- a/src/geos.c
+++ b/src/geos.c
@@ -597,3 +597,71 @@ GEOSGeometry* create_point(GEOSContextHandle_t ctx, double x, double y) {
   return geom;
 #endif
 }
+
+
+/* Force 2D */
+
+
+
+GEOSGeometry* force_2d_simple(GEOSContextHandle_t ctx, GEOSGeometry* geom, int type) {
+  int dims, i;
+  unsigned int n, j;
+  int target_dims = 2;
+  double coord;
+  const GEOSCoordSequence* seq = GEOSGeom_getCoordSeq_r(ctx, geom);
+
+  if (GEOSCoordSeq_getDimensions_r(ctx, seq, &dims) == 0) {
+    return NULL;
+  }
+  if (dims == target_dims) {
+    return GEOSGeom_clone_r(ctx, geom);
+  }
+
+  if (GEOSCoordSeq_getSize_r(ctx, seq, &n) == 0) {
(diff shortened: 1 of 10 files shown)


brendan-ward (commented):
Thanks for working on this @caspervdw 

I didn't have a chance to fully review the C extension changes yet, but wanted to get you some initial feedback quickly.


Should our repr / WKT indicate the 3rd dimension for empty geometries?

```
>>> pygeos.Geometry('POINT Z EMPTY')
<pygeos.Geometry POINT EMPTY>
```

Since a dimension may be set but not visible here?  E.g.,
```
>>> pygeos.get_coordinate_dimension(pygeos.Geometry('POINT Z EMPTY'))
3
>>> pygeos.get_coordinate_dimension(pygeos.Geometry('POINT EMPTY'))
2
```

src/geos.c
```diff
@@ -597,3 +597,220 @@ GEOSGeometry* create_point(GEOSContextHandle_t ctx, double x, double y) {
...
+  }
+
+  /* Investigate the coordinate sequence, return when already of correct dimensionality */
+  if (GEOSCoordSeq_getDimensions_r(ctx, seq, &actual_dims) == 0) {
+    return NULL;
+  }
+  if (actual_dims == dims) {
+    return GEOSGeom_clone_r(ctx, geom);
+  }
+  if (GEOSCoordSeq_getSize_r(ctx, seq, &n) == 0) {
```
brendan-ward, line 663: I may be viewing this out of context, but it looks like this condition could be combined with the first:

```
  if ((GEOSCoordSeq_getSize_r(ctx, seq, &n) == 0) || (GEOSCoordSeq_getDimensions_r(ctx, seq, &actual_dims) == 0)) {
    return NULL;
  }

```

brendan-ward (commented):
Had a chance to look over the C extension changes; other than one minor suggestion, those look good as well.

src/geos.c
caspervdw: Yes you could combine these conditionals, but I don't like to do that because the conditional is just there to provide a route in case GEOS errors. It is a repeated pattern that, if you get used to it, improves readability of the code.

caspervdw:
@brendan-ward  I addressed your comments. Thanks for the review!

PR #399, opened 2021-09-28
[Done] Fix GEOS main caching

### LINKED ISSUE #385: CI caching does not work for 'main'  builds
In our CI we have a build against the `main` branch of GEOS. Currently, the cache does not work as intended.

The issue is as follows: for the `main` build, the `${{ github.workspace }}/geosinstall` directory is cached under the cache key `${{ runner.os }}-geos-main`. If there is a cache available, that whole directory is restored by GH actions, including `geosinstall/rev.txt` which contains the revision hash. If the current `HEAD` of `GEOS` does not agree, GEOS is rebuilt and the `rev.txt` updated. But then: the cache is not saved! So the cache remains at its original `rev.txt` which gets older until the cache is evicted (unsure what is the policy of that).

Glancing at the `actions/cache` docs, I don't think there is a way of forcing the adapted `geosinstall` to be cached. Maybe we could include the revision hash into the cache key.

caspervdw:
Closes #385

### DIFF
--- a/.github/workflows/test-pip.yml
+++ b/.github/workflows/test-pip.yml
@@ -49,10 +49,32 @@ jobs:
 
     env:
       GEOS_VERSION: ${{ matrix.geos }}
+      GEOS_VERSION_SPEC: ${{ matrix.geos }}
       GEOS_INSTALL: ${{ github.workspace }}/geosinstall/geos-${{ matrix.geos }}
+      GEOS_BUILD: ${{ github.workspace }}/geosbuild
 
     steps:
-      - uses: actions/checkout@v2
+      - name: Correct slashes in GEOS_INSTALL (Windows)
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test-pip.yml
```diff
@@ -118,9 +136,7 @@ jobs:
         if: ${{ matrix.os == 'windows-2019' }}
 
       - name: Build PyGEOS
-        run: |
-          python setup.py build_ext --inplace
-          pip install --no-deps -e .
+        run: python setup.py build_ext --inplace
```
jorisvandenbossche, line 139: Is there a reason to no longer install the package? (because we are running the tests from the source directory anyway?)

caspervdw: Yes there is no reason to “pip install” if you are in the source directory and if the extensions are built. Removing the line saves some seconds in the CI.

PR #401, opened 2021-10-01
[WIP] assert_geometries_equal

### LINKED ISSUE #253: API change: equals(None, None) should be True

caspervdw:
Closes #253

### DIFF
--- a/pygeos/testing.py
+++ b/pygeos/testing.py
@@ -1,0 +1,95 @@
+import numpy as np
+from numpy.testing import assert_almost_equal
+
+import pygeos
+
+__all__ = ["assert_geometries_equal"]
+
+
+def _assert_nan_coords_same(x, y, decimal):
+    x_ndim = pygeos.get_coordinate_dimension(x)
+    y_ndim = pygeos.get_coordinate_dimension(y)
+    if x_ndim != y_ndim:
+        return False
+
+    x_coords = pygeos.get_coordinates(x, include_z=x_ndim == 3)
+    y_coords = pygeos.get_coordinates(y, include_z=x_ndim == 3)
+
+    # Check NaN equality
+    try:
+        assert_almost_equal(x_coords, y_coords)
+    except AssertionError:
+        return False
+    else:
+        return True
+
+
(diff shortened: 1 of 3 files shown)


pygeos/testing.py
```diff
@@ -0,0 +1,95 @@
...
+    y_ndim = pygeos.get_coordinate_dimension(y)
+    if x_ndim != y_ndim:
+        return False
+
+    x_coords = pygeos.get_coordinates(x, include_z=x_ndim == 3)
+    y_coords = pygeos.get_coordinates(y, include_z=x_ndim == 3)
+
+    # Check NaN equality
+    try:
+        assert_almost_equal(x_coords, y_coords)
```
jorisvandenbossche, line 20: You probably want to pass `decimal` here?

I think `np.testing.assert_allclose` is considered as the newer, preferable method to do this instead of assert_almost_equals (which uses relative/absolute tolerance instead of number of decimals)

pygeos/testing.py
```diff
@@ -0,0 +1,95 @@
...
+        return bool(y_id)
+    else:
+        return y_id
+
+
+def assert_geometries_equal(x, y, decimal=7, equal_none=True, equal_nan=True):
+    """Raises an AssertionError if two geometry array_like objects are not equal.
+
+    Given two array_like objects, check that the shape is equal and all elements of
+    these objects are equal. An exception is raised at shape mismatch or conflicting
```
jorisvandenbossche, line 50: Currently this uses `equals_exact` for structural coordinate equality. Should we add an option to be able to get spatial equality (`equals`) as well?

jorisvandenbossche (commented):
Some initial comments

pygeos/testing.py
jorisvandenbossche: In addition, for equals_exact it might also be nice to add a `normalize=True` option (so you don't have to do that manually beforehand)

PR #403, opened 2021-10-03
[Done] Fix 3D empty WKT serialization

caspervdw:
This is a (hopefully) temporary fix for the 3D point serialization. See https://trac.osgeo.org/geos/ticket/1129

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,7 @@ Version 0.11 (unreleased)
 
 * Return True instead of False for LINEARRING geometries in ``is_closed`` (#379).
 * Fixed the WKB serialization of 3D empty points for GEOS >= 3.9.0 (#392).
+* Fixed the WKT serialization of 3D empty points for GEOS >= 3.9.0 (#402).
 * Fixed the WKT serialization of multipoints with empty points for GEOS >= 3.9.0 (#392).
 
 **Acknowledgments**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,7 +28,7 @@ Version 0.11 (unreleased)
 
 * Return True instead of False for LINEARRING geometries in ``is_closed`` (#379).
 * Fixed the WKB serialization of 3D empty points for GEOS >= 3.9.0 (#392).
-* Fixed the WKT serialization of 3D empty geometries for GEOS >= 3.9.0 (#402).
+* Fixed the WKT serialization of single part 3D empty geometries for GEOS >= 3.9.0 (#402).
 * Fixed the WKT serialization of multipoints with empty points for GEOS >= 3.9.0 (#392).
 
 **Acknowledgments**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,7 +26,7 @@ Version 0.11 (unreleased)
 
 * Return True instead of False for LINEARRING geometries in ``is_closed`` (#379).
 * Fixed the WKB serialization of 3D empty points for GEOS >= 3.9.0 (#392).
-* Fixed the WKT serialization of 3D empty points for GEOS >= 3.9.0 (#402).
+* Fixed the WKT serialization of 3D empty geometries for GEOS >= 3.9.0 (#402).
 * Fixed the WKT serialization of multipoints with empty points for GEOS >= 3.9.0 (#392).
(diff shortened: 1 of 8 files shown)


caspervdw:
@brendan-ward Hold your horses -- this issue is not only applying to the POINTS, but all other 3D empty geometries get serialized incorrectly. I'll try to fix that in this PR as well.

caspervdw:
@brendan-ward  It's done now :)

src/geos.c
```diff
@@ -291,6 +291,72 @@ char check_to_wkt_compatible(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
...
+  int geom_type;
+
+  is_empty = GEOSisEmpty_r(ctx, geom);
+  if (is_empty == 0) {
+    *wkt = NULL;
+    return PGERR_SUCCESS;
+  } else if (is_empty != 1) {
+    return PGERR_GEOS_EXCEPTION;
+  }
+  if (GEOSGeom_getCoordinateDimension_r(ctx, geom) == 2) {
```
brendan-ward, line 322: This appears to always return 2 for empty 3D geometry collections, which them prevents it from falling through to the condition below.
```
>>> pygeos.get_coordinate_dimension(pygeos.Geometry('GEOMETRYCOLLECTION Z EMPTY'))
2
```

brendan-ward (commented):
Thanks for working on this @caspervdw !

A few suggestions below, but it looks like the biggest issue is that `GEOMETRYCOLLECTION Z EMPTY` is not getting recognized as 3d, so it sidesteps your fix here.

src/geos.c
brendan-ward: Thanks for checking into that.  Perhaps the thing to do is remove the fix here for GeometryCollections, since it won't be reached, and update the docstring to indicate it won't work for empty GeometryCollections?

caspervdw: I updated the docstring of `to_wkt`.

Not sure what to do with the unused code. I know the parts currently cannot be reached, but I like to be forwards compatible if possible. What if empty collections turn out to be 3D at some point? If I would delete the lines, the `default:` would then lead to a `GEOSException`.

I think it is easier to leave the lines in and leave a comment about the situation. That is what I did now.

src/geos.c
```diff
@@ -291,6 +291,74 @@ char check_to_wkt_compatible(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
...
+      *wkt = "LINESTRING Z EMPTY";
+      break;
+    case GEOS_LINEARRING:
+      *wkt = "LINEARRING Z EMPTY";
+      break;
+    case GEOS_POLYGON:
+      *wkt = "POLYGON Z EMPTY";
+      break;
+    // Note: Empty collections cannot be 3D in GEOS.
+    // We do include the options incase of future support.
```
brendan-ward, line 341:
```suggestion
    // We do include the options in case of future support.
```

pygeos/io.py
```diff
@@ -99,6 +99,13 @@ def to_wkt(
     The Well-known Text format is defined in the `OGC Simple Features
     Specification for SQL <https://www.opengeospatial.org/standards/sfs>`__.
 
+    The following limitations apply to WKT serialization:
+
+    - for GEOS <= 3.8 a multipoint with an empty sub-geometry will raise an exception
+    - for GEOS <= 3.8 empty geometries are always serialized to 2D
+    - for GEOS == 3.9 only simple empty geometries can be 3D, collections are still
```
brendan-ward, line 106:
```suggestion
    - for GEOS >= 3.9 only simple empty geometries can be 3D, collections are still
```

(will need to revise if this ever changes in GEOS)

CHANGELOG.rst
```diff
@@ -28,6 +28,7 @@ Version 0.11 (unreleased)
 
 * Return True instead of False for LINEARRING geometries in ``is_closed`` (#379).
 * Fixed the WKB serialization of 3D empty points for GEOS >= 3.9.0 (#392).
+* Fixed the WKT serialization of 3D empty geometries for GEOS >= 3.9.0 (#402).
```
brendan-ward, line 31:
```suggestion
* Fixed the WKT serialization of single part 3D empty geometries for GEOS >= 3.9.0 (#402).
```

PR #413, opened 2021-10-26
[Done] GeoJSON IO for GEOS 3.10

caspervdw:
This implements to/from geojson ufuncs using the new GeoJSON reader and writer from GEOS 3.10.

There are a few open issues, we get crashes on:

```
>>> pygeos.from_geojson('{"no": "geojson"}')
python: /home/casper/code/geos/include/geos/vend/json.hpp:20363: const value_type& geos_nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::operator[](T*) const [with T = const char; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long int; NumberUnsignedType = long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = geos_nlohmann::adl_serializer; BinaryType = std::vector<unsigned char>; geos_nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::const_reference = const geos_nlohmann::basic_json<>&; geos_nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::value_type = geos_nlohmann::basic_json<>]: Assertion `m_value.object->find(key) != m_value.object->end()' failed.
Aborted (core dumped)
```

And on:

```
In [2]: pygeos.to_geojson(pygeos.Geometry("POINT EMPTY"))
```

The Z coordinates are dropped:

```
In [6]: pygeos.to_geojson(pygeos.Geometry("POINT Z (1 1 2)"))
Out[6]: '{"type":"Point","coordinates":[1.0,1.0]}
```

Also, nan coordinates are serialized to null, but null is not serialized to nan:
```
in [3] pygeos.to_geojson(pygeos.Geometry("LINESTRING (1 nan, 2 2)"))
Out[3]: '{"type":"LineString","coordinates":[[1.0,null],[2.0,2.0]]}'
In [4]: pygeos.from_geojson('{"type":"LineString","coordinates":[[1.0,null],[2.0,2.0]]}')
<ipython-input-4-22368789e223> in <module>
----> 1 pygeos.from_geojson('{"type":"LineString","coordinates":[[1.0,null],[2.0,2.0]]}')

~/code/pygeos/pygeos/io.py in from_geojson(geometry, on_invalid, unpack, **kwargs)
    453         )
    454 
--> 455     result = lib.from_geojson(geometry, invalid_handler, **kwargs)
    456     if unpack and get_type_id(result) == GeometryType.GEOMETRYCOLLECTION:
    457         result = get_parts(result)

GEOSException: ParseException: Error parsing JSON: '[json.exception.type_error.302] type must be number, but is null'

### DIFF
--- a/pygeos/decorators.py
+++ b/pygeos/decorators.py
@@ -1,9 +1,10 @@
 import os
 from functools import wraps
+from multiprocessing import Process, Queue
 
 import numpy as np
 
-from . import lib
+from . import GEOSException, lib
 
 
 class UnsupportedGEOSOperation(ImportError):
@@ -83,3 +84,38 @@ def multithreading_enabled(func):
                 arr.flags.writeable = old_flag
 
     return wrapped
+
+
+def may_segfault(func):
+    """The wrapped function will be called in another process.
+
+    If the execution crashes with a segfault or sigabort, a GEOSException
+    will be raised.
+    """
+
+    # wrap the function so that it also returns exceptions
+    def return_exc(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:
+            return e
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        queue = Queue()
+        process = Process(
+            target=lambda q: q.put(return_exc(*args, **kwargs)),
+            args=(queue,),
+        )
+        process.start()
+        process.join()
+        if process.exitcode == 0:
+            result = queue.get()
+            if isinstance(result, Exception):
+                raise result
+            else:
+                return result
+        else:
+            raise GEOSException(f"GEOS crashed with exit code {process.exitcode}.")
+
+    return wrapper
--- a/pygeos/decorators.py
+++ b/pygeos/decorators.py
@@ -1,10 +1,9 @@
-import multiprocessing
 import os
 from functools import wraps
 
 import numpy as np
 
-from . import GEOSException, lib
+from . import lib
 
 
 class UnsupportedGEOSOperation(ImportError):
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by caspervdw** [comment]
The GEOS crashes that I got locally are not reproduced by the CI! So that is good news.

**Comment by brendan-ward** [comment]
I just updated my local GEOS (on Mac) to tip of `main` and do not get a GEOS crash for invalid GeoJSON ; I get `ParseException` as expected.

**Comment by caspervdw** [comment]
That’s odd- I compiled 3.10.0 and got the SigAbort each time. Did you compile with “Debug” mode? I did that. I’ll recompile in “Release” mode and see if the issue goes away.

**Comment by brendan-ward** [comment]
Indeed - that's it!  I get the crash if I compile in Debug mode, and not in Release mode.

File: pygeos/decorators.py [hunk 1]
```diff
@@ -83,3 +84,47 @@ def wrapped(*args, **kwargs):
...
+            self._cconn.send(e)
+
+    @property
+    def result(self):
+        if self._result is None and self._pconn.poll():
+            self._result = self._pconn.recv()
+        return self._result
+
+
+def may_segfault(func):
```
brendan-ward, line 110: This is no longer used; now that the segfaults are either guarded against (in case of empty points) or are limited to Debug compiles of GEOS (hopefully rare for users).

It seems useful to archive this, but not sure that it should go into this PR.  If you keep it here, perhaps it would be helpful to add more comments to help guide devs in when to use it (i.e., this is a bit of a last resort for segfaults from GEOS that can't be otherwise guarded against)?

brendan-ward (approved):
Thanks for updating this!  A few minor comments.

It might be a good idea to add a few comments with crossreferences to the GEOS issues for LinearRing, empty points, etc so that we have those at hand to check in future releases to see if they've been resolved.

pygeos/decorators.py
caspervdw: I think it should not be in this repo if it isn't used. Instead, I created a gist: https://gist.github.com/caspervdw/10d9fd373a31642498cdea494d2f1bc4

Issue #414, opened 2021-10-26
Dropping python and GEOS 3.6


caspervdw:
Ref #125

caspervdw:
Some second thoughts: it is quite easy for us to keep the older versions, it just takes some more time in the test matrix. We don't do a lot of fancy new python stuff. And we have a good framework for working with older GEOS versions (macros on C-level, a decorator on the Python level).

So a policy on top of NEP 29 could be: also support the versions on the oldest LTS Ubuntu version. I personally work with Ubuntu 18.04 a lot and it is convenient to be able to use their system versions (which are Python 3.6, GEOS 3.6). Ubuntu 18 is in support until April '23.

PR #416, opened 2021-10-29
[Done] wheels: Add Python 3.10 and GEOS 3.10.0

caspervdw:
As discussed in #414 , this also drops support for GEOS<=3.6 and Python 3.6.


jorisvandenbossche:
Is this required for a 0.12 release? Things are working, so I would say that we could also release keeping those version support to have a last release for them, and only afterwards drop them?

I would also prefer to first check that the other Shapely devs are on the same page. I don't expect any problems, but I also don't want to create new blockers for starting to integrate the pygeos code into the shapely repo (now there is a maintenance branch for 1.8, I hope to do that soon)

caspervdw:
Right, it might be a bit rash to drop the support of the older versions. Ubuntu 18 for instance ships with GEOS 3.6 and Python 3.6.

I do like to have the wheels for py 3.10 in the matrix already.

jorisvandenbossche:
> I do like to have the wheels for py 3.10 in the matrix already.

Yes, that part would certainly be nice

jorisvandenbossche:
And I also +1 on using GEOS 3.10 for the wheels, to be clear

caspervdw:
I reverted the changes and kept only the GEOS 3.10.0 and Python 3.10 addition.
Should be good for a release after this is merged.

jorisvandenbossche (approved):
Thanks!

caspervdw:
The 64-bit build seems to end up in lib64 as of GEOS 3.10.0. Might be caused by https://github.com/libgeos/geos/pull/447.

Our script expects the libraries in lib. I am not used to working with cmake— @mwtoews could you give me a hint how to proceed?

mwtoews:
Use `-D CMAKE_INSTALL_LIBDIR=lib` as is done for conda [here](https://github.com/conda-forge/geos-feedstock/blob/master/recipe/build.sh).

caspervdw:
> Use `-D CMAKE_INSTALL_LIBDIR=lib` as is done for conda [here](https://github.com/conda-forge/geos-feedstock/blob/master/recipe/build.sh).

Thanks @mwtoews this saved me a lot of time going through cmake documentation. The build works, I'll wait for the aarch64 to complete and then we can merge & release.

PR #417, opened 2021-10-30
[Done] dwithin for GEOS 3.10.0

caspervdw:


brendan-ward (commented):
Thanks for working on this @caspervdw 

It may be a good idea to note that negative distances always returns `False`, but since negative values are allowed for things like `buffer` it doesn't seem necessary to guard against them here.

This works as expected for what I tested manually.

brendan-ward:
For this update to the benchmarks
```
class PointPolygonTimeSuite:
    """Benchmarks running on 100000 points and one polygon"""

    def setup(self):
        self.points = pygeos.points(np.random.random((100000, 2)))
        coords = np.random.random((3, 2))
        self.polygon = pygeos.polygons(coords)
        self.polygon_prepared = pygeos.polygons(coords)
        pygeos.prepare(self.polygon_prepared)

    def time_contains(self):
        pygeos.contains(self.points, self.polygon)

    def time_distance(self):
        pygeos.distance(self.points, self.polygon)

    def time_intersection(self):
        pygeos.intersection(self.points, self.polygon)

    def time_intersects(self):
        pygeos.intersects(self.polygon, self.points)

    def time_intersects_prepared(self):
        pygeos.intersects(self.polygon_prepared, self.points)

    def time_distance_within(self):
        pygeos.lib.distance_within(self.polygon, self.points, 10)

    def time_distance_within_prepared(self):
        pygeos.lib.distance_within(self.polygon_prepared, self.points, 10)

    def time_intersects_buffer(self):
        pygeos.intersects(self.polygon, pygeos.buffer(self.points, 10))
```

I'm seeing that `distance_within` is much faster than the standard trick of buffer then intersect:
```
time_contains                                                        1.79±0.2ms
time_distance                                                           150±7ms
time_intersection                                                       379±6ms
time_intersects                                                        151±20ms
time_intersects_prepared                                             4.07±0.3ms
time_intersects_buffer                                                4.16±0.2s
time_distance_within                                                    123±2ms
time_distance_within_prepared                                         123±0.5ms
```
This isn't a good test of how prepared compares to unprepared (need more complex polygons).  Though the `intersects` with prepared polygons is much faster than the unprepared version.

pygeos/tests/test_predicates.py
```diff
@@ -112,6 +118,19 @@ def test_equals_exact_tolerance():
     assert pygeos.equals_exact(p1, p2).item() is False
 
 
+@pytest.mark.skipif(pygeos.geos_version < (3, 10, 0), reason="GEOS < 3.10")
+def test_distance_within():
+    # specifying tolerance
```
jorisvandenbossche, line 123: Comment is probably from copy paste?

pygeos/tests/test_predicates.py
```diff
@@ -112,6 +118,19 @@ def test_equals_exact_tolerance():
...
+    p1 = pygeos.points(50, 4)
+    p2 = pygeos.points(50.1, 4.1)
+    actual = pygeos.distance_within([p1, p2, None], p1, distance=0.05)
+    np.testing.assert_allclose(actual, [True, False, False])
+    assert actual.dtype == np.bool_
+    actual = pygeos.distance_within([p1, p2, None], p1, distance=0.2)
+    np.testing.assert_allclose(actual, [True, True, False])
+    assert actual.dtype == np.bool_
```
jorisvandenbossche, line 131: Maybe add a test with a array for `distance`? (I see that this is possible from the docstring examples, and I suppose this might not be covered by the default predicate tests?)

jorisvandenbossche (commented):
Looks good, added two minor comments. 

Further I have one more question about the name: intuitively, I would find "within_distance" much clearer than "distance_within" (since we are not returning the "distance" within the geometry, or something like that). Thoughts?

I know that GEOS is using `DistanceWithin`, so that's of course a good reason to follow suite. PostGIS uses `DWithin` (https://postgis.net/docs/ST_DWithin.html), so also with that order of the names, but since we deviate from PostGIS in this case anyway, I find that less important here.

caspervdw:
> I know that GEOS is using `DistanceWithin`, so that's of course a good reason to follow suite. PostGIS uses `DWithin` (https://postgis.net/docs/ST_DWithin.html), so also with that order of the names, but since we deviate from PostGIS in this case anyway, I find that less important here.
 
I see your point here. But why deviating from postgis? “dwithin” would be acceptable too, and then we adhere to our own policy of taking the same names as PostGIS.

pygeos/tests/test_predicates.py
```diff
@@ -111,6 +117,26 @@ def test_equals_exact_tolerance():
...
+    actual = pygeos.equals_exact(p1, p2, tolerance=[0.05, 0.2, np.nan])
+    np.testing.assert_allclose(actual, [False, True, False])
+
+
+@pytest.mark.skipif(pygeos.geos_version < (3, 10, 0), reason="GEOS < 3.10")
+def test_dwithin():
+    p1 = pygeos.points(50, 4)
+    p2 = pygeos.points(50.1, 4.1)
+    actual = pygeos.dwithin([p1, p2, None], p1, distance=0.05)
+    np.testing.assert_allclose(actual, [True, False, False])
```
brendan-ward, line 130:
```suggestion
    np.testing.assert_equal(actual, [True, False, False])
```
(since these are not floating point numbers)

brendan-ward (approved):
Other than a couple minor tweaks to the tests, this looks good.  Thanks @caspervdw

PR #418, opened 2021-10-31
[Done] Reinstate may_segfault for from_geojson

caspervdw:
In #413  we concluded that from_geojson segfaults for some inputs only when GEOS is built in Debug mode.

Now, a CI failed (https://github.com/pygeos/pygeos/runs/4059627215?check_suite_focus=true) which builds GEOS in Release mode. So I think it is best to stay on the safe side and reinstate the segfault protection here.

Related GEOS Ticket has been solved in main already:
- https://trac.osgeo.org/geos/ticket/1138
- https://github.com/libgeos/geos/pull/498

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,9 @@ Version 0.12 (unreleased)
 
 **Bug fixes**
 
-* ...
+* Protect ``pygeos.from_geojson`` against segfaults by running the function in a
+  subprocess (GEOS 3.10.0 only).
+
 
 **Acknowledgments**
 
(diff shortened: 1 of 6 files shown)


caspervdw:
Note that the "main" test runners pass for these added tests; the issue has been fixed in GEOS main already.

brendan-ward (commented):
Since `may_segfault` is not used as a decorator, perhaps it and associated `ReturningProcess` should be moved to a different file?

Otherwise, this looks reasonable - the change is very isolated to a specific GEOS version.

PR #422, opened 2021-11-08
CI: upgrade GEOS versions, fix "Run doctests"

mwtoews:
GEOS version upgrades:

- 3.6.4 -> 3.6.5 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2020-December/009994.html)
- <s>3.8.1 -> 3.8.2 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2021-April/010201.html)</s>
- 3.9.1 -> 3.9.2 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2021-November/010526.html)
- 3.10.0 -> 3.10.1 [announcement](https://lists.osgeo.org/pipermail/geos-devel/2021-November/010556.html)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
     name: Build ${{ matrix.archs }} wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     env:
-      GEOS_VERSION: "3.10.0"
+      GEOS_VERSION: "3.10.1"
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test-pip.yml
+++ b/.github/workflows/test-pip.yml
@@ -14,11 +14,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-2019]
         architecture: [x64]
-        geos: [3.6.4, 3.7.3, 3.8.1, 3.9.1, 3.10.0, main]
+        geos: [3.6.5, 3.7.3, 3.8.2, 3.9.2, 3.10.1, main]
         include:
           # 2017
           - python: 3.6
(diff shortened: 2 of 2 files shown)


mwtoews:
Doctests now run ([and pass](https://github.com/pygeos/pygeos/runs/4137306442?check_suite_focus=true)).

One failure `windows-2019-x64 Py3.8 GEOS 3.8.2` does not build GEOS. The patch is https://github.com/libgeos/geos/commit/da1b246f28df4c2b7863920fbc913b316ca77062 too much trouble? If so, then I can reset back to GEOS 3.8.1.

caspervdw:
Thanks for fixing the doc builds, good catch! 

About the failing builds on windows with GEOS 3.8.2, I think it is not worth the hassle to apply the patch. If tests succeed with 3.8.1, that is sufficient for me. Could you revert that?

Good to see btw that the GeoJSON reader/writer fixes are already released in. I think I’ll revert #418

mwtoews:
GEOS 3.8.1 is kept, but I'll keep an eye out for GEOS 3.8.3, if released.

mwtoews:
So is this error with `Build aarch64 wheels on ubuntu-20.04` related to reverting #418? Should it be addressed in this PR?
```
  ______________________________ test_may_segfault _______________________________
  
      def test_may_segfault():
          if os.name == "nt":
              match = "access violation"
          else:
              match = "GEOS crashed"
          with pytest.raises(OSError, match=match):
  >           may_segfault(my_unstable_func)("segfault")
  E           Failed: DID NOT RAISE <class 'OSError'>
```

caspervdw:
The aarch64 issue is indeed related to #423 (revert #418) . Merging this

PR #425, opened 2021-11-10
ENH: add dwithin to STRtree

### LINKED ISSUE #402: API: STRtree handling of GEOSDistanceWithin
[GEOS 3.10 is bringing new functions to the CAPI](https://github.com/libgeos/geos/blob/main/NEWS): GEOSDistanceWithin / GEOSPreparedDistanceWithin

In addition to exposing these as predicates similar to `pygeos.intersects` (but with broadcastable distance value), we likely want to add this to STRtree.

There are a couple ways we could do this:
1. extend `STRtree.query_bulk` to take allow `within_distance` as a valid predicate that requres an additional parameter `distance`.  Distance could be broadcast to the same shape as the input geometries (meaning: unique distance value per input geometry).  This would first need to extend the envelopes of the input geometries +/- `distance` (plus a tiny epsilon value) before querying the tree (like we do in `nearest_all`), then apply `GEOSPreparedDistanceWithin` predicate after that.
2. add a dedicated `STRtree.within_distance` function that behaves much like `query_bulk` that only does the same operations as above.  My instinct is that this is the cleanest approach.

brendan-ward:
Resolves #402 

This adds support for the distance within predicate to STRtree.

API considerations:
- this currently handles distances as vectors instead of scalar values, so that each input geometry can have a different search distance.  I _think_ numpy handles this via `broadcast_to` so that the underlying memory is not greatly multiplied if a scalar value is provided as input.   (I have use cases where I need different distances per input)
- this handles scalar vs vector geometries at the Python tier, detecting if geometry is a scalar and simply returning the associated tree indices (similar to `query`).  Internally this is implemented as a bulk-type operation like `query_bulk`.

Early testing shows that this is slower than buffering input geometries by distance and then using `query_bulk` with `intersects` predicate.  Internally, both are using prepared geometries in the same way (always preparing if not provided that way).  There is only a little extra overhead from creating an expanded search window before querying the tree, but otherwise the steps are pretty similar between `dwithin` and `query_bulk`, so I'm not sure what is driving the time difference yet.


brendan-ward:
It looks like `get_bounds` that we use to create an expanded envelope for searching the tree is the main driver of why `dwithin` is slower than `buffer` + `query_bulk`.  Didn't expect that to be so slow.

brendan-ward:
I take that back, it doesn't appear to be `get_bounds`.  Given the same set of points returned by the expanded bbox filter (first step here) vs those returned by an envelope around a radius of that same distance (in query_bulk), using a prepared intersects predicate on those than a prepared distance within predicate.  It may well be that there is optimized pre-filtering used by the intersects predicate that hasn't yet been leveraged into the distance within predicate.  Will need to investigate the GEOS implementation further...

brendan-ward:
Indeed, the performance is only a little better for my test case than calculating the distance between geometries queried from the tree ourselves using the expanded bbox filter.  This suggests the majority of the time in GEOS is being spent calculating the distances.

I haven't come up with enough test cases to see how it varies; presumably at some point, buffering complex geometries will become slower than calculating the distance from those geometries to the target geometries (i.e., distance within will be more performant than buffer + query tree + intersects predicate).

caspervdw:
I looked through the code, it looks good and I am :+1: on moving forward. The benchmarks probably depend a lot on the size of your tree and the complexity of the geometries. I am not worried about that.

One thing I realized while looking at the C code: there is a lot of boilerplate code that is now duplicated between ` query_bulk` and `dwithin`. You *could* also add it as a predicate option, with an optional argument `distance`. Of course there will be some `if (predicate == DWITHIN)` clauses but my guess it it would not look as bad. But I am also fine with the current approach, I leave it up to you. On the Python side I think it makes sense to split them either way.

caspervdw:
@brendan-ward  I shifted the milestone to the attached issue; I am still :+1: on wrapping this up before 0.12

brendan-ward:
Definitely agreed on the amount of boilerplate redundancy.  Unfortunately, most is around validating inputs, which doesn't lend as well to consolidating into a shared function.

For now, I'd like to keep the C implementation separate from `query_bulk` for 2 reasons:
1. we have to version guard `dwithin` to GEOS 3.10
2. there may eventually be a function for this implemented within GEOS; there already is one for the [SimpleSTRtree](https://github.com/libgeos/geos/blob/3.10/src/index/strtree/SimpleSTRtree.cpp#L370-L379) (not exposed to C API).  Keeping this out of `query_bulk` would make adopting that easier.

brendan-ward:
Will need to come up with better benchmarks; `dwithin` is slower (sometimes by a lot) vs buffer and intersect for everything I've tried so far, including real-world data like country boundaries.  But I don't think any of that is on the `pygeos` side, so the performance shouldn't be a reason to hold this PR back.

pygeos/strtree.py
```diff
@@ -303,3 +303,67 @@ def nearest_all(self, geometry, max_distance=None, return_distance=False):
...
+        [[0, 0], [0, 1]]
+        >>> tree.dwithin(pygeos.points([[0,0]]), 0).tolist()  # doctest: +SKIP
+        [[0], [0]]
+        >>> tree.dwithin(pygeos.points([[0,0]]), np.nan).tolist()  # doctest: +SKIP
+        [[], []]
+        >>> tree.dwithin(pygeos.Geometry('POINT EMPTY'), 10).tolist()  # doctest: +SKIP
+        [[], []]
+        """
+
+        is_scalar = np.isscalar(geometry)
```
brendan-ward, line 349: This isn't working the way I expected.  I'm trying to test for singular geometries here (which may be `None`), but `np.isscalar` is false for these.

The goal was to try and make this behave like `query` for singular inputs, and `query_bulk` for array-like inputs.  Maybe we should always treat inputs as array-like?

caspervdw: I think I am against this API anyway. The type of the return value should not depend on the inputs given to a function.

We already have a pattern estabilished for this (`query` /  `query_bulk`), I think the logical option would be  `dwithin` / `dwithin_bulk`

caspervdw (commented):
I had the time to review now. All looks good, thanks a lot for doing this so thoroughly.

The only open issue is the external (Python) API. My preferred option would be `STRtree().query(geoms, predicate=“dwithin”, distance=1.0)` instead of the separate method. What do you think?

pygeos/strtree.py
caspervdw: One thought: I understand that the C implementation is separate, but we could integrate dwithin into query/query_bulk on the python side only?

brendan-ward:
Thanks for the review!

I had interpreted 
> On the Python side I think it makes sense to split them either way.
as indicating that we should keep `dwithin` and `query` / `query_bulk` separate?  Apologies if I had misinterpreted that.

> `STRtree().query[_bulk](geoms, predicate=“dwithin”, distance=1.0)`

This seems reasonable and simplifies the singular vs multiple inputs issue I was struggling a bit with.  I'll start working on a refactor.

However, it presents a potential issue for us to consider:  we currently use the `BinaryPredicate` enum to list valid predicates for use with `query` / `query_bulk`.  Downstream, this is used by geopandas to determine valid predicates for spatial join operations.

Since `dwithin` is a special case - both because of inputs and GEOS version required - I was thinking that we should omit it from this enum.  Which I think makes sense, because in geopandas, we wouldn't want to expose that blindly as an sjoin operation when it requires a distance parameter; if we ever want to allow a spatial join on this particular predicate, we'd need to handle that as a special case on the geopandas side as well.

caspervdw:
I might have given a different indication earlier - sometimes I take a different point of view after having seen the implementation. Sorry if it caused confusion.

About the enum; unless there is something really breaking in geopandas, I would just extend it. Extending an enum is mostly considered backwards compatible. Worst case the user tries  “dwithin” in a spatial join and gets an error. “dwithin” would remain undocumented on the geopandas side right?

brendan-ward:
`dwithin` is now a predicate in `query` / `query_bulk`.

I did not add it to the `BinaryPredicates` enum; it wasn't needed.

pygeos/strtree.py
```diff
@@ -97,20 +102,35 @@ def query(self, geometry, predicate=None):
         >>> # Query geometries that are contained by input geometry
         >>> tree.query(pygeos.box(2, 2, 4, 4), predicate='contains').tolist()
         [3]
+        >>> # Query geometries within 1 unit distance of input geometry
+        >>> tree.query(pygeos.points(0.5,0.5), predicate='dwithin', distance=1.0).tolist()  # doctest: +SKIP
```
caspervdw, line 106:
```suggestion
        >>> tree.query(pygeos.points(0.5, 0.5), predicate='dwithin', distance=1.0).tolist()  # doctest: +SKIP
```

pygeos/strtree.py
```diff
@@ -97,20 +102,35 @@ def query(self, geometry, predicate=None):
...
-            predicate = 0
-
-        else:
-            predicate = BinaryPredicate.get_value(predicate)
-
+            return self._tree.query(geometry, 0)
+
+        elif predicate == "dwithin":
+            if lib.geos_version < (3, 10, 0):
+                raise ImportError("dwithin predicate requires GEOS >= 3.10")
```
caspervdw, line 118:
```suggestion
                raise UnsupportedGEOSOperation("dwithin predicate requires GEOS >= 3.10")
```

pygeos/strtree.py
```diff
@@ -163,18 +189,38 @@ def query_bulk(self, geometry, predicate=None):
...
-
-        else:
-            predicate = BinaryPredicate.get_value(predicate)
-
+            return self._tree.query_bulk(geometry, 0)
+
+        # Requires GEOS >= 3.10
+        elif predicate == "dwithin":
+            if lib.geos_version < (3, 10, 0):
+                raise ImportError("dwithin predicate requires GEOS >= 3.10")
```
caspervdw, line 207:
```suggestion
                raise UnsupportedGEOSOperation("dwithin predicate requires GEOS >= 3.10")
```

caspervdw (approved):
Thanks @brendan-ward , this looks good to merge after a few minor comments.

PR #430, opened 2021-11-14
Documentation fixes

### LINKED ISSUE #289: DOC: pygeos.get_dimensions on empty LINESTRING
Based on [pygeos.get_dimensions](https://pygeos.readthedocs.io/en/latest/geometry.html#pygeos.geometry.get_dimensions)

> Returns the inherent dimensionality of a geometry.
> 
> The inherent dimension is 0 for points, 1 for linestrings and linearrings, and 2 for polygons. For geometrycollections it is the max of the containing elements. **Empty and None geometries return -1**.

Strange, I was expecting -1 here:
```
pygeos.get_dimensions(pygeos.Geometry("LINESTRING EMPTY"))
> 1
```

I am trying to filter out LINESTRING EMPTY, maybe I should use another method to check "emptiness"
```
mask = array([<pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry MULTIPOINT (-7.23e+06 4.88e+06, -6.7e+06 4.96e+06)>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry POINT (-6.7e+06 4.96e+06)>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>,
       <pygeos.Geometry LINESTRING EMPTY>], dtype=object)

pygeos.get_dimensions(mask)
> array([[1, 1, 1, 1, 1, 0, 1, 1, 1],
       [1, 1, 1, 1, 1, 0, 1, 1, 1]], dtype=int32)
```

---
Edit:
Looks like this would do it, but it is quite verbose, to filter out all empty LINESTRINGS
```
mask[(pygeos.get_num_coordinates(crosses_mask) == 2) & (pygeos.get_type_id(crosses_mask) == pygeos.GeometryType.LINESTRING)]
```

### LINKED ISSUE #397: points conveniance function does not evaluate lazily
If point geometries are created with the the `points` conveniance function and dask arrays the geometries are not returned lazily because in https://github.com/pygeos/pygeos/blob/ea76ac9b287f85bf120eeb3d326d1251eff1f995/pygeos/creation.py#L67
`np.broadcast` is called on the arrays.

I worked around this by directly using `pygeos.lib.points` which works fine unless a `dask.distributed` cluster is used in which case a critical error `Failed to deserialize` is raised by `distributed`.

As a sidenote: using dask arrays and distributed to create box polygons works fine.

I use pygeos 0.10.2

### LINKED ISSUE #424: Spatial join returns incorrect results with Shapely 1.8, due to shapely_compatible optimistically set to True
#### Code Sample, a copy-pastable example

```python
import geopandas as gpd
from shapely import wkt

# Define 2 geometries
point = wkt.loads('POINT (-3.61104 54.12069)')
polygon = wkt.loads('''
    Polygon ((-3.63156666666667 54.1267166666667,-3.53435 54.0763833333333,-3.5634 54.0383333333333,
    -3.58396666666667 54.0467,-3.60175 54.0521333333333,-3.62465 54.05735,-3.64296666666667
    54.0595333333333,-3.65568333333333 54.0740166666667,-3.6695 54.0932833333333,-3.65568333333333
    54.1054666666667,-3.63156666666667 54.1267166666667))
''')
print('\npoint within polygon?')
print(point.within(polygon))

# Geodataframe with 1 polygon
df = gpd.GeoDataFrame.from_dict({
    'id': [1],
    'geometry': [polygon],
})
# Geodataframe with 1 Point
poi = gpd.GeoDataFrame.from_dict({
    'id': [2],
    'geometry': [point],
})

print('\nSpatial join:')
print(gpd.sjoin(poi, df, predicate="within"))

```

#### Problem description

The spatial join behaviour (with GeoPandas PyGeos) has changed between Shapely 1.7.1 (correct behaviour) and Shapely 1.8 (incorrect behaviour).
Note that the point does not fall into the polygon. These are WGS84 coordinates, and the distance from the point to the polygon is ~400 meters.
image

**With Shapely 1.7.1 (correct behaviour)**
image

**With Shapely 1.8 (incorrect behaviour)**
Note that the point is still considered out of the polygon, but the spatial joins incorrectly considers the point to be in the polygon.
image

#### Expected Output

The expected output is the one with Shapely 1.7.1. 
I think this comes from the fact that PyGeos & Shapely 1.8 are built on the same GEOS version, hence no conversions apply (the warnings does not appear). Maybe the formats are not exactly compatible. 

Indeed, the problem disappears if I force Pygeos to perform the WKB conversion between Shapely & Pygeos geometries.
More precisely, if I set the boolean shapely_compatible to False directly in Pygeos source code, I get the expected behaviour.
https://github.com/pygeos/pygeos/blob/41a48cfde8a0823c7ac44e4dc9b2cf6ae63f0f07/pygeos/io.py#L69

More details on the relative issue on GeoPandas: https://github.com/geopandas/geopandas/issues/2198

Note: the problem does not happen with pygeos 0.11.1, that is built on newer version of GEOS, but may happen again in the future if Shapely GEOS version is upgraded.

#### Output of ``geopandas.show_versions()``

This is the output with shapely 1.8.0. Simply run `pip install shapely==1.7.1` to test the working configuration.

<details>

SYSTEM INFO
-----------
python     : 3.9.5 (default, May  4 2021, 03:33:11)  [Clang 12.0.0 (clang-1200.0.32.29)]
executable : /Users/nicolaslegrand/.virtualenvs/py39/bin/python
machine    : macOS-10.16-x86_64-i386-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.9.1
GEOS lib   : /usr/local/Cellar/geos/3.9.1/lib/libgeos_c.dylib
GDAL       : 3.2.2
GDAL data dir: None
PROJ       : 7.2.1
PROJ data dir: /Users/nicolaslegrand/.virtualenvs/py39/lib/python3.9/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.2
pandas     : 1.2.5
fiona      : 1.8.19
numpy      : 1.21.3
shapely    : 1.8.0
rtree      : None
pyproj     : 3.0.1
matplotlib : 3.4.3
mapclassify: None
geopy      : 2.2.0
psycopg2   : 2.9.1 (dt dec pq3 ext lo64)
geoalchemy2: None
pyarrow    : None
pygeos     : 0.10.2

</details>

caspervdw:
Closes #289, Closes #397, Closes #424

### DIFF
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -66,7 +66,7 @@ def points(coords, y=None, z=None, indices=None, out=None, **kwargs):
 
     - GEOS >=3.10 automatically converts POINT (nan nan) to POINT EMPTY.
     - Usage of the ``y`` and ``z`` arguments will prevents lazy evaluation in ``dask``.
-      Instead provide the coordinates as a ``(..., 2)`` or ``(..., 3)`` array using only ``coords``.
+      Instead provide the coordinates as an array with shape ``(..., 2)`` or ``(..., 3)`` using only the ``coords`` argument.
     """
     coords = _xyz_to_coords(coords, y, z)
     if indices is None:
--- a/pygeos/creation.py
+++ b/pygeos/creation.py
@@ -34,9 +34,6 @@ def _xyz_to_coords(x, y, z):
 def points(coords, y=None, z=None, indices=None, out=None, **kwargs):
     """Create an array of points.
 
-    Note that GEOS >=3.10 automatically converts POINT (nan nan) to
-    POINT EMPTY.
-
     Parameters
     ----------
     coords : array_like
@@ -63,6 +60,13 @@ def points(coords, y=None, z=None, indices=None, out=None, **kwargs):
     [<pygeos.Geometry POINT (0 1)>, <pygeos.Geometry POINT (4 5)>]
     >>> points([0, 1, 2])
     <pygeos.Geometry POINT Z (0 1 2)>
+
+    Notes
+    -----
+
+    - GEOS >=3.10 automatically converts POINT (nan nan) to POINT EMPTY.
+    - Usage of the ``y`` and ``z`` arguments will prevents lazy evaluation in ``dask``.
+      Instead provide the coordinates as a ``(..., 2)`` or ``(..., 3)`` array using only ``coords``.
     """
     coords = _xyz_to_coords(coords, y, z)
     if indices is None:
@@ -104,6 +108,11 @@ def linestrings(coords, y=None, z=None, indices=None, out=None, **kwargs):
     [<pygeos.Geometry LINESTRING (0 1, 4 5)>, <pygeos.Geometry LINESTRING (2 3, 5 6)>]
     >>> linestrings([[0, 1], [4, 5], [2, 3], [5, 6], [7, 8]], indices=[0, 0, 1, 1, 1]).tolist()
     [<pygeos.Geometry LINESTRING (0 1, 4 5)>, <pygeos.Geometry LINESTRING (2 3, 5 6, 7 8)>]
+
+    Notes
+    -----
+    - Usage of the ``y`` and ``z`` arguments will prevents lazy evaluation in ``dask``.
+      Instead provide the coordinates as a ``(..., 2)`` or ``(..., 3)`` array using only ``coords``.
     """
     coords = _xyz_to_coords(coords, y, z)
     if indices is None:
@@ -151,6 +160,11 @@ def linearrings(coords, y=None, z=None, indices=None, out=None, **kwargs):
     <pygeos.Geometry LINEARRING (0 0, 0 1, 1 1, 0 0)>
     >>> linearrings([[0, 0], [0, 1], [1, 1]])
     <pygeos.Geometry LINEARRING (0 0, 0 1, 1 1, 0 0)>
+
+    Notes
+    -----
+    - Usage of the ``y`` and ``z`` arguments will prevents lazy evaluation in ``dask``.
+      Instead provide the coordinates as a ``(..., 2)`` or ``(..., 3)`` array using only ``coords``.
     """
     coords = _xyz_to_coords(coords, y, z)
     if indices is None:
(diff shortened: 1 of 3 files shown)


pygeos/creation.py
```diff
@@ -63,6 +60,13 @@ def points(coords, y=None, z=None, indices=None, out=None, **kwargs):
     [<pygeos.Geometry POINT (0 1)>, <pygeos.Geometry POINT (4 5)>]
     >>> points([0, 1, 2])
     <pygeos.Geometry POINT Z (0 1 2)>
+
+    Notes
+    -----
+
+    - GEOS >=3.10 automatically converts POINT (nan nan) to POINT EMPTY.
+    - Usage of the ``y`` and ``z`` arguments will prevents lazy evaluation in ``dask``.
+      Instead provide the coordinates as a ``(..., 2)`` or ``(..., 3)`` array using only ``coords``.
```
jorisvandenbossche, line 69:
```suggestion
      Instead provide the coordinates as an array with shape ``(..., 2)`` or ``(..., 3)`` using only the ``coords`` argument.
```

jorisvandenbossche (commented):
Looks good!

(I might still try to look into detecting the usage of wheels vs conda, and then set it automatically, but at least documenting it is a good first step)

PR #432, opened 2021-11-14
Fix error handling in STRtree

### LINKED ISSUE #139: ENH: More user friendly error for STRtree failure
I found that for certain invalid geometries, STRtree fails with no information on the cause. This is certainly an edge case, but it would be nice to have a more helpful error message.

```python3
from geopandas import read_file, datasets, _compat
import pygeos

_compat.set_use_pygeos(True)

world = read_file(datasets.get_path('naturalearth_lowres'))
countries = world.to_crs('epsg:3395')
countries = countries[['geometry', 'name']]
geometry = countries.geometry.values.data


tree = pygeos.strtree.STRtree(geometry[countries.name == 'Antarctica'])
tree.query(geometry[0], "touches")
```
Failure:
```
<built-in method query of pygeos.lib.STRtree object> returned a result with an error set
```

The other way around has the same result:
```python3
tree = pygeos.strtree.STRtree([geometry[0]])
tree.query(geometry[countries.name == 'Antarctica'][0], "touches")
```

The failure seems to depend on the combination of input/tree geometries as well as the predicate (ex: `"intersects"` does not fail).

caspervdw:
Closes #139

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,7 +29,7 @@ Version 0.12 (unreleased)
 
 * Raise ``GEOSException`` in the rare case when predicate evalution in ``STRtree.query``
   errors. Previously, the exceptions were ignored silently and the geometry was added 
-  to the result (as if the predicated returned ``True``) (#432).
+  to the result (as if the predicate returned ``True``) (#432).
 
 
 **Acknowledgments**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,9 @@ Version 0.12 (unreleased)
 
 **Bug fixes**
 
-* ...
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -27,7 +27,9 @@ Version 0.12 (unreleased)
 
 **Bug fixes**
 
-* ...
+* Raise ``GEOSException`` in the rare case when predicate evalution in ``STRtree.query``
+  errors. Previously, the exceptions were ignored silently and the geometry was added 
+  to the result (as if the predicated returned ``True``) (#432).
```
jorisvandenbossche, line 32:
```suggestion
  to the result (as if the predicate returned ``True``) (#432).
```

jorisvandenbossche (approved):
Nice!

PR #436, opened 2021-11-18
PERF: Use GEOSCoordSeq_copyFromBuffer for creating simple geometries

jorisvandenbossche:
Experiment in using the new `GEOSCoordSeq_copyFromBuffer` for creating CoordinateSequences instead of setting each ordinate individually.

This requires GEOS 3.10, so will require https://github.com/pygeos/pygeos/issues/255 to use this in cython.

I was benchmarking it with the US census polygons ([tl_2019_us_zcta510.zip](https://catalog.data.gov/dataset/tiger-line-shapefile-2019-2010-nation-u-s-2010-census-5-digit-zip-code-tabulation-area-zcta5-na), 33144 (multi)polygons, with on average ~1200 coordinates per ring).

Focusing on LinearRing creation (the actual creation of (Multi)Polygons from those rings will of course see less of a difference):

I saved the coordinates and offsets in a numpy file, to load them again to benchmark (see https://nbviewer.org/gist/jorisvandenbossche/4fab9c94525dcb5ddda8a89b4a5ad9f3 for the full notebook):

```python
import numpy as np
import pygeos

with np.load("pygeos-benchmark-linearrings.npz") as data:
    offsets = data["offsets"]
    coords = data["coords"]

ring_lengths = np.diff((offsets / 2).astype(int))
ring_indices = np.repeat(np.arange(len(ring_lengths)), ring_lengths)
rings = pygeos.linearrings(coords, indices=ring_indices)
```

and this gives:

```
In [2]: %timeit pygeos.linearrings(coords, indices=ring_indices)
1.13 s ± 6.86 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)   # <-- master
558 ms ± 2.35 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <-- PR
```

I had maybe hoped on a bit more :), but it's certainly a nice speed-up.

### DIFF
--- a/pygeos/_geometry.pyx
+++ b/pygeos/_geometry.pyx
@@ -38,22 +38,11 @@ from pygeos._pygeos_api cimport (
     import_pygeos_c_api,
     PyGEOS_CreateGeometry,
     PyGEOS_GetGEOSGeometry,
+    PyGEOSCoordSeq_FromBuffer,
 )
 
 # initialize PyGEOS C API
 import_pygeos_c_api()
-
-
-cdef char _set_xyz(GEOSContextHandle_t geos_handle, GEOSCoordSequence *seq, unsigned int coord_idx,
-                   unsigned int dims, double[:, :] coord_view, Py_ssize_t idx):
-    if GEOSCoordSeq_setX_r(geos_handle, seq, coord_idx, coord_view[idx, 0]) == 0:
-        return 0
-    if GEOSCoordSeq_setY_r(geos_handle, seq, coord_idx, coord_view[idx, 1]) == 0:
-        return 0
-    if dims == 3:
-        if GEOSCoordSeq_setZ_r(geos_handle, seq, coord_idx, coord_view[idx, 2]) == 0:
-            return 0
-    return 1
 
 
 def _check_out_array(object out, Py_ssize_t size):
@@ -75,17 +64,16 @@ def _check_out_array(object out, Py_ssize_t size):
 @cython.boundscheck(False)
 @cython.wraparound(False)
 def simple_geometries_1d(object coordinates, object indices, int geometry_type, object out = None):
-    cdef Py_ssize_t idx = 0
+    cdef Py_ssize_t idx = 0, flat_idx = 0
     cdef unsigned int coord_idx = 0
     cdef Py_ssize_t geom_idx = 0
     cdef unsigned int geom_size = 0
     cdef unsigned int ring_closure = 0
-    cdef Py_ssize_t coll_geom_idx = 0
     cdef GEOSGeometry *geom = NULL
     cdef GEOSCoordSequence *seq = NULL
 
     # Cast input arrays and define memoryviews for later usage
-    coordinates = np.asarray(coordinates, dtype=np.float64)
+    coordinates = np.asarray(coordinates, dtype=np.float64, order="C")
     if coordinates.ndim != 2:
         raise TypeError("coordinates must be a two-dimensional array.")
 
@@ -99,7 +87,6 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type, 
     cdef unsigned int dims = coordinates.shape[1]
     if dims not in {2, 3}:
         raise ValueError("coordinates should N by 2 or N by 3.")
-    cdef int has_z = int(dims == 3)
 
     if geometry_type not in {0, 1, 2}:
         raise ValueError(f"Invalid geometry_type: {geometry_type}.")
@@ -111,7 +98,7 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type, 
     if np.any(indices[1:] < indices[:indices.shape[0] - 1]):
         raise ValueError("The indices must be sorted.")  
 
-    cdef double[:] coord_view = coordinates.ravel()
+    cdef double[:, :] coord_view = coordinates
 
     # get the geometry count per collection (this raises on negative indices)
     cdef unsigned int[:] coord_counts = np.bincount(indices).astype(np.uint32)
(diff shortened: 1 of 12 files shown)


pygeos/_geometry.pyx
```diff
@@ -109,7 +111,7 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type,
     if np.any(indices[1:] < indices[:indices.shape[0] - 1]):
         raise ValueError("The indices must be sorted.")  
 
-    cdef double[:, :] coord_view = coordinates
+    cdef double[:] coord_view = coordinates.ravel()
```
jorisvandenbossche, line 114: We would need to ensure the correct layout here (row major (XYXY), i.e. C contiguous). I suppose this is simply `.ravel(order="C")`, although that might actually already be the default.

pygeos/_geometry.pyx
```diff
@@ -134,37 +136,43 @@ def simple_geometries_1d(object coordinates, object indices, int geometry_type,
...
                 geom = GEOSGeom_createLineString_r(geos_handle, seq)
             elif geometry_type == 2:
-                if ring_closure == 1:
-                    if _set_xyz(geos_handle, seq, geom_size, dims, coord_view, idx - geom_size) == 0:
-                        GEOSCoordSeq_destroy_r(geos_handle, seq)
-                        return  # GEOSException is raised by get_geos_handle
+                # if ring_closure == 1:
+                #     if _set_xyz(geos_handle, seq, geom_size, dims, coord_view, idx - geom_size) == 0:
+                #         GEOSCoordSeq_destroy_r(geos_handle, seq)
+                #         return  # GEOSException is raised by get_geos_handle
```
jorisvandenbossche, line 175: When using the `GEOSCoordSeq_copyFromBuffer_r`, I am note sure we should automatically add a coordinate for ring closure.

(although I suppose it is still possible; we can first copy in one extra element (the first of the next geometry?) and then overwrite that later?)

Personally I think that when you already are providing `indices`, it should also be a rare use case (eg if you got those indices from one of our functions that returns indices, this will never be a problem). 
We could of course still check for it (now I also commented out that code). 

This could maybe also be an option of the function?

jorisvandenbossche:
Doing a profile of the above benchmark with py-spy gives the following:

pygeos-linearrings

So this confirms that inside `pygeos.linearrings` now most time is spent in `GEOSCoordSeq_copyFromBuffer` (66%). Another 25% is from `np.bincount`, and another 8% is from the comparison `indices[1:] < indices[:indices.shape[0] - 1]`(validating the indices are sorted)

brendan-ward:
Thanks for working on this!  The results are promising (though a bummer that we spend so much time just validating inputs).

I wonder if it would be easier to introduce this first in the ufuncs (e.g., `lib.linestrings`) where we can use the C macros to guard the GEOS version?  That would unblock the dependency on adding GEOS version control mechanisms to Cython in order to first start using `GEOSCoordSeq_copyFromBuffer` in pygeos?  That said, we'll certainly want this in Cython too, just was trying to think through smaller steps to bringing this in.

jorisvandenbossche:
> I wonder if it would be easier to introduce this first in the ufuncs (e.g., `lib.linestrings`) where we can use the C macros to guard the GEOS version?

That would indeed certainly be _easier_, but in practice those ufuncs are not super useful for general IO (except for points), since you almost never have all polygons of the same size (eg I couldn't have done my benchmark with real world data as I did above). So that's the reason I wanted to start looking at the non-ufunc versions. 
(but we certainly want to add it to the ufuncs as well)

jorisvandenbossche:
> though a bummer that we spend so much time just validating inputs

It's not actually only validation, but also the calculation of the size of each CoordSeq that takes quite a bit of time. That's a consequence of how we expect the indices to be laid out (as `[0, 0, 0, 0, 1, 1, 2, 2, 2]` instead of `[0, 4, 6, 9]`). 
Long term, I think it would be nice to also have a routine that takes that kind of offsets (since this is what Arrow will natively store). 

The validation indeed also takes some time. We could maybe add a keyword to disable this, if you already know for sure that this is the case (but with a default to always validate).

brendan-ward:
Likely crazy idea, not yet thought through: what if we exposed a function from pygeos C API that we consume in Cython so that it takes the same inputs as are now being passed into `GEOSCoordSeq_copyFromBuffer`.  Then, instead of using this function direction from GEOS, we call back into pygeos C API, which then uses C Macros to version-guard the underlying GEOS functions used.  I'm thinking that what we add would be lightweight wrappers around some of these important low-level GEOS functions where they are only available in certain recent versions of GEOS.

I.e., instead of adding shims to Cython build system to conditionally make certain GEOS functions available to Cython, we provide Cython C function shims that are implemented back in C on our end.

I don't know yet how the complexity of that would differ vs just making the Cython build system GEOS version aware.

jorisvandenbossche:
That's not so crazy I think :)

We could indeed have a C function that either is a simple call of `GEOSCoordSeq_copyFromBuffer` or either creates a CoordSeq manually and fills this in iteratively, depending on the GEOS version (which is easy to do in C). And then we can call that in the cython code. This is basically the `_set_xyz` + the loop in which it is called that we now do custom in cython. 

That might indeed end up simpler.

caspervdw:
Thanks for investigating this and doing the benchmarks! Interesting results. For sure the new GEOS function gives significant improvements. And we have some handles to optimize even further.

We could indeed make an optional “offsets” keyword argument that is preferred above the “indices”. It depends on your use case which representation is fastest, apparently here “offsets” is a better match. This is material for another issue I guess.

For the coordinates -> coordseq conversion I think the proposed solution of @brendan-ward is not crazy at all. I think that is the way to go.

jorisvandenbossche:
I did a first attempt to write a generic "create and fill in coord seq" function that can be used in in the ufuncs (and then later in the cython code as well). Some complications: the ufuncs can also handle differently strided arrays (so I have to check if the steps are correct to use the copyFromBuffer), and needed some casts back and forth between `*double` and `*char` (the steps are expressed in bytes, not for `* double`). This can probably be cleaned up a bit more.

jorisvandenbossche:
Quick performance check for the ufunc as well (not compared with master, but comparing with itself but triggering the copyFromBuffer path by providing unclosed vs closed ring coordinates):

```python
In [15]: coords1 = np.random.randn(10000, 100, 2)

In [16]: %timeit pygeos.linearrings(coords1)
17.2 ms ± 603 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [17]: coords1[:, [99], :] = coords1[: , [0], :]

In [18]: %timeit pygeos.linearrings(coords1)   # <-- closed rings, uses the copyFromBuffer fast path
8.93 ms ± 598 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

and with fewer but larger rings:

```python
In [19]: coords1 = np.random.randn(100, 10000, 2)

In [20]: %timeit pygeos.linearrings(coords1)
10.8 ms ± 111 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [21]: coords1[:, [9999], :] = coords1[: , [0], :]

In [22]: %timeit pygeos.linearrings(coords1)
3.54 ms ± 44.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

src/geos.c
```diff
@@ -880,3 +881,48 @@ GEOSGeometry* PyGEOSForce2D(GEOSContextHandle_t ctx, GEOSGeometry* geom) {
...
+  unsigned int i, j;
+  double first_coord;
+
+#if GEOS_SINCE_3_10_0
+
+  if ((!ring_closure) && (cs1 == dims * 8) && (cs2 == 8)) {
+    int hasZ = dims == 3;
+    coord_seq = GEOSCoordSeq_copyFromBuffer_r(ctx, buf, size, hasZ, 0);
+    return coord_seq;
+  }
```
caspervdw, line 899: I think we could use `GEOSCoordSeq_copyFromArrays ` here if `(cs1 == 8) && (cs2 == size * 8)`

in that case `x = buf, y = buf + cs2, z = (dims == 3) ? buf + 2 * cs2 : NULL

jorisvandenbossche: Looking into this, I am not sure it would actually work. At least not for "standard" fortran ordered arrays, because the strides are not exactly how we would need it:

```python
>>> coords = np.asarray(np.random.randn(10, 4, 2), order="F")
>>> coords.shape
(10, 4, 2)
>>> coords.strides
(8, 80, 320)
```

So the cs1 and cs2 is actually `total_size * 8` and `total_size * size * 8`, with `total_size` being 10 in this case (the total number of geometries that are going to be created), and that's something you don't know in the ufunc inner function + then it won't have the contiguous layout for CopyFromArrays anyway

So I could easily add the check you suggested, but I am not sure how useful it would be in practice. I need to do quite some axis wrangling in numpy to get an array with the correct strides to make use of this:

```
>>> coords = np.asarray(np.random.randn(4, 2, 10), order="F")
>>> coords = np.swapaxes(np.swapaxes(coords, 0, 2), 1, 2)
>>> coords.shape
(10, 4, 2)
>>> coords.strides
(64, 8, 32)
```

caspervdw: I see. I guess it would 'work' for two-dimensional input arrays (scalar outputs), but in those cases you probably can't tell which GEOS function was used as the Python interpreter overhead would take up most of the time

caspervdw (approved):

caspervdw:
Looks good to merge! Can you add a line to the changelog?

src/geos.c
jorisvandenbossche: That's actually a good point. When creating a relatively large LineString or LinearRing (tested 1000 vertices), the difference is actually notable also in Python. So added that, and a test case using a scalar.

PR #444, opened 2021-11-29
Arm64 wheel builds on Travis

### LINKED ISSUE #442: CI: Run the aarch64 build in a different workflow
The aarch64 build may take hours while our other builds complete within 10-30 minutes. This is because we use emulation. Also, the build sporadically hangs after which it is killed at 6 hours. There is no way of restarting the build in that case.

I don’t like this as it delays the actual release (which is done only after all envs are completed.

Maybe we can move this build to Travis (which has native support for aarch64) or else move it into an independent workflow so that it does not block the others.

caspervdw:
Closes #442


jorisvandenbossche:
Seems to be working already (partly) quite nicely! The manylinux wheels seem to went fine, it's now failing for musllinux (popd / pushd issue, I seem to remember we ran into that recently in other CI)

jorisvandenbossche:
Ah, it was in https://github.com/pygeos/pygeos/pull/416 where you reverted and re-reverted the popd/pushd changes

caspervdw:
This should be OK now @jorisvandenbossche . I must say the Travis builds feel very unstable, mostly I have to restart at least one of them. But still, cibuildwheel completes in ~10 minutes (if it works) and the others in 1-2 minutes. So I guess we can afford pressing "Rerun job" a few times. Good thing that Travis has the option of rerunning builds.

I'll leave it at running on master & travis branches only to save build minutes.

Also nice to know: the s390x failures disappeared somewhere between GEOS 3.8 and 3.10.1 (see #297)

jorisvandenbossche:
Nice, that's certainly a win compared to the emulated ones on github actions. Thanks a lot for doing this!

PR #445, opened 2021-12-01
Move to manylinux2014 wheels

caspervdw:
manylinux2010 wheels are built for CentOS 6, which reached end of life in November 2020 (see https://www.python.org/dev/peps/pep-0599/)

So it is good to move on to manylinux2014 which is for CentOS 7.

Numpy also started doing this for their cp310 wheels; because of this our cp310 builds are compiling numpy. This PR solves that issue too.


caspervdw:
@jorisvandenbossche I’ll port this to shapely as well - this is mainly to be able to release 0.12

jorisvandenbossche:
Sounds good!

PR #450, opened 2022-04-08
MAINT: upgrade release and tests to use GEOS 3.10.2

mwtoews:
The motivation for this is from GeoPandas and pygeos installed via pip:
```
../../../../tmp/py39/lib/python3.9/site-packages/geopandas/_compat.py:111
  /tmp/py39/lib/python3.9/site-packages/geopandas/_compat.py:111: UserWarning: The Shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the GEOS version PyGEOS was compiled with (3.10.1-CAPI-1.16.0). Conversions between both will be slow.
    warnings.warn(
```
While Shapely-2.0 is not out yet, it would be nice to have binary wheels on PyPI that have he same GEOS version as Shapely.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
     name: Build ${{ matrix.arch }} wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     env:
-      GEOS_VERSION: "3.10.1"
+      GEOS_VERSION: "3.10.2"
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test-pip.yml
+++ b/.github/workflows/test-pip.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-2019]
         architecture: [x64]
-        geos: [3.6.5, 3.7.3, 3.8.1, 3.9.2, 3.10.1, main]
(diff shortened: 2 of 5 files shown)


bnaul:
@jorisvandenbossche @brendan-ward any chance the wheels can be rebuilt on 3.10.2? would be nice to get rid of this warning, especially since per https://github.com/geopandas/geopandas/pull/2275 it sounds like Shapely 2.0 support is a ways off.

jorisvandenbossche:
I am fine with doing a new release, just to update the wheels (assuming our release infrastructure is still working) 
@caspervdw OK with that?

jorisvandenbossche:
In addition, we will have to do a release anyway to fix compatibility with shapely 2.0 -> https://github.com/pygeos/pygeos/pull/452

jorisvandenbossche:
The update from GEOS 3.10.1 to 3.10.2 should be trivial, and this is also fixing the CI, so going to merge this.

Thanks @mwtoews !

jorisvandenbossche:
@mwtoews in the meantime, we updated the shapely release workflow to use GEOS 3.10.3 (instead of 3.10.2). If we are going to release a shapely 1.8.3 with that the coming days, we should probably also here update to GEOS 3.10.3 to ensure wheel compatibility with the latest shapely.

mwtoews:
Yes, I agree to the paired releases of pygeos and shapely-1.8 with the same GEOS version, to avoid the same error shown above. I'll prepare another PR here...

PR #452, opened 2022-08-08
Fix to/from_shapely for compat with shapely 2

jorisvandenbossche:
To allow to have pygeos and shapely 2 installed together and interoperable (useful for pygeos users that have to gradually migrate to shapely 2)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 files: 'pygeos'
 repos:
 -   repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
     - id: black
 -   repo: https://gitlab.com/pycqa/flake8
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
cc @caspervdw I would like to do a release with this fix included relatively soon, does that sound OK?

caspervdw (approved):
Sounds good to include this in a maintenance release!
I noted that there are no unittests done with shapely actually present. Did you test this with both 1.8 and 2.0 versions?

Also the change log should be updated, but that can also be done as part of the release.

jorisvandenbossche:
> I noted that there are no unittests done with shapely actually present. 

Yeah, in the past we added those mock objects to avoid adding shapely as a test dependency, but we could also easily add an actual shapely test (although I don't know if that's still worth it at this point)

> Did you test this with both 1.8 and 2.0 versions?

Yes, I ran locally the geopandas test suite (which has this interaction of converting between shapely and pygeos) with this branch and with both shapely 1.8 and 2.0 (with shapely 2.0 it is also tested on geopandas' CI at the moment in https://github.com/geopandas/geopandas/pull/2275)

PR #453, opened 2022-08-08
CI: update black version

jorisvandenbossche:
Currently failing, see https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click


jorisvandenbossche:
The pytest failures are already being fixed in https://github.com/pygeos/pygeos/pull/450/, so going to merge this as such

PR #454, opened 2022-08-09
CI: upgrade GEOS patch versions for tests, use GEOS 3.10.3 for release

mwtoews:
See https://lists.osgeo.org/pipermail/geos-devel/2022-June/010685.html for announcement of several patch releases.

Builds for releases have changed to GEOS 3.10.3 (see also #450).

Also, build sdist using PyPA's [build](https://github.com/pypa/build) tool.


jorisvandenbossche (approved):
Thanks!

Issue #447, opened 2021-12-03
GEOS version compatibility with shapely, MacOS and set_operations.intersection/union_all


nialov:
# Introduction

When `pygeos` and `shapely` have matching `GEOS` version enabling faster geometry transforms between the packages there is an odd bug in the MacOS platform that does not exist for the Ubuntu platform. I've managed to partly pin this down in this repository:

https://github.com/nialov/pygeos-macos-shapely-incompatibility

The issue is caused by conducting `geopandas.clip` which itself calls `pygeos.set_operations.intersection` (or `pygeos.set_operations.union_all`) which either call `lib.intersection` (or `lib.unary_union`) which return the same error when the `mask` inputted into `geopandas.clip` is a `MultiPolygon` (or when `GEOMETRYCOLLECTION` is passed into `union_all` I'm not certain if in that case the mask was actually a `MultiPolygon`):

~~~
pygeos.GEOSException: IllegalArgumentException: Argument must be Polygonal or LinearRing
~~~

# System

Platform: MacOS (from github actions)

pygeos==0.10.2
geopandas==0.10.2
shapely==1.8.0

~~~
In [17]: from shapely.geos import geos_version_string; geos_version_string
Out[17]: '3.9.1-CAPI-1.14.2'

In [18]: pygeos.geos_capi_version_string
Out[18]: '3.9.1-CAPI-1.14.2'
~~~

# Logs

I've pinned the intersection error here:
https://github.com/nialov/pygeos-macos-shapely-incompatibility/runs/4411204893?check_suite_focus=true

~~~
a = array([<pygeos.Geometry MULTILINESTRING ((4.66e+05 6.69e+06, 4.66e+05 6.69e+06, 4.6...>],
      dtype=object)
b = <pygeos.Geometry MULTIPOLYGON (((4.66e+05 6.69e+06, 4.66e+05 6.69e+06, 4.66e...>
grid_size = None, kwargs = {}

    @multithreading_enabled
    def intersection(a, b, grid_size=None, **kwargs):
        """Returns the geometry that is shared between input geometries.

...

         if grid_size is not None:
            if lib.geos_version < (3, 9, 0):
                raise UnsupportedGEOSOperation("grid_size parameter requires GEOS >= 3.9.0")
    
            if not np.isscalar(grid_size):
                raise ValueError("grid_size parameter only accepts scalar values")
    
            return lib.intersection_prec(a, b, grid_size, **kwargs)
    
[...]

Issue #377, opened 2021-08-19
Test new binary wheels


caspervdw:
The new binary wheels are distributed under version 0.10.2a2!
I'd like to spot-test some of them before tagging the release.

Please try it out:
```
$ pip install --pre pygeos[test]
$ pytest --pyargs pygeos.tests
```

And report back the exact wheel version you used + if tests passed.

For me: ```pygeos-0.10.2a2-cp38-cp38-manylinux_2_12_x86_64.manylinux2010_x86_64.whl ``` works.

@jorisvandenbossche @mwtoews @brendan-ward

jorisvandenbossche:
Nice!

Worked for me as wel (but also on linux, using `pygeos-0.10.2a2-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.whl`)

Which wheels are still missing at the moment? Mainly the Mac Arm64? (plus linux ppc64le and s390x)

caspervdw:
Some tests on my windows laptop:
- pygeos-0.10.2a2-cp36-cp36m-win_amd64.whl works
- pygeos-0.10.2a2-cp39-cp39m-win_amd64.whl works

> Which wheels are still missing at the moment? Mainly the Mac Arm64? (plus linux ppc64le and s390x)

There are open issues for Mac Arm64 (#372) and Pypy (#371). The two other architectures you mention are also not yet available for numpy, and we need that to build pygeos.

martinfleis:
`pygeos-0.10.2a2-cp39-cp39-macosx_10_9_x86_64.whl` on Intel macOS is fine.

martinfleis:
pygeos tests are fine it breaks GeoPandas on the MRE below.

```py
import geopandas
df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
df.dissolve('continent')
```

```py
GEOSException: IllegalArgumentException: Argument must be Polygonal or LinearRing
---------------------------------------------------------------------------
GEOSException                             Traceback (most recent call last)
/var/folders/kp/fxnnw89x5qbb9gryn507p03c0000gn/T/ipykernel_40038/2272171905.py in <module>
      1 import geopandas
      2 df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
----> 3 df.dissolve('continent')

~/Git/geopandas/geopandas/geodataframe.py in dissolve(self, by, aggfunc, as_index, level, sort, observed, dropna)
[...]

Issue #122, opened 2020-03-23
Document that from_shapely requires exactly the same GEOS library for both shapely and pygeos to function without potential errors


brendan-ward:
Given a watershed - should be simple polygon
(shapefile: [bad_poly.zip](https://github.com/pygeos/pygeos/files/4370849/bad_poly.zip))

Read in via geopandas 0.7 (`df`) and converted to pygeos:

```
>>> geometry = pygeos.from_shapely(df.geometry)
```

Geopandas / shapely claim it is valid:
```
>>> df.geometry.is_valid
0    True
```

Pygeos raises exception:
```
>>> pygeos.is_valid(geometry)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/bcward/projects/pygeos/pygeos/predicates.py", line 258, in is_valid
    result = lib.is_valid(geometry, **kwargs)
pygeos.GEOSException: IllegalArgumentException: Argument must be Polygonal or LinearRing
```

But otherwise pygeos thinks it is OK:
```
>>> pygeos.is_valid_input(geometry)
array([ True])
>>> pygeos.is_geometry(geometry)
array([ True])
```

If we round-trip the geometry through WKT or WKB, then it is valid and raises no exception:
```
>>> pygeos.is_valid(pygeos.from_wkt(pygeos.to_wkt(geometry)))
array([ True])
>>> pygeos.is_valid(pygeos.from_wkb(pygeos.to_wkb(geometry)))
array([ True])
```

Likewise, if I use WKT or WKB to go between the shapely geometry and pygeos, no errors:
```
>>> pg.is_valid(pg.from_wkt(df.geometry.apply(lambda g: g.to_wkt())))
0    True
Name: geometry, dtype: bool
>>> pg.is_valid(pg.from_wkb(df.geometry.apply(lambda g: g.to_wkb())))
array([ True])
```

I think there is a couple things going on here:
1) `pygeos::is_valid` should probably return `False` in this case rather than raise an exception.  It should handle exceptions from GEOS as indicating `False`.

2) there is a difference in behavior between `from_shapely` versus `from_wkt` / `from_wkb`.  It could be that that latter correct some underlying issue in the raw geometry that is passed through as-is via `from_shapely`.

jorisvandenbossche:
This might be a GEOS version issue / build issue. I get 

```
In [48]: df = geopandas.read_file("zip://./../Downloads/bad_poly.zip")                                                                                                                                             

[...]

PR #459, opened 2022-12-11
Only use fastpath in to/from_shapely conversion when using conda

jorisvandenbossche:
Since we have had several reports of subtle crashes in certain cases when using wheels that seem to be related to this sharing of GEOS pointers (for example https://github.com/pygeos/pygeos/issues/447, https://github.com/pygeos/pygeos/issues/377, https://github.com/pygeos/pygeos/issues/122, https://github.com/geopandas/geopandas/issues/2551, https://github.com/geopandas/geopandas/issues/2613#issuecomment-1290387118), it seems safer to restrict this fastpath to conda-only. 

This shouldn't impact results, only a slowdown in shapely / pygeos conversion. 
This also doesn't actually check that you are using a shapely and pygeos conda package, only if you are in a python managed by conda (but you can still install wheels into a conda environment).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
     rev: 22.3.0
     hooks:
     - id: black
--   repo: https://gitlab.com/pycqa/flake8
+-   repo: https://github.com/pycqa/flake8
     rev: 3.9.0
     hooks:
     - id: flake8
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@caspervdw if we would still do one more pygeos release for 3.11 support, this is something to consider including as well.

caspervdw (approved):

